Issue #190, opened 2013-08-03
What is the proper way to set the timezone of datetime columns in finagle-mysql?


eric:
finagle-mysql [writes DATETIME and TIMESTAMP columns in the current timezone](https://github.com/twitter/finagle/blob/master/finagle-mysql/src/main/scala/com/twitter/finagle/mysql/Value.scala#L168) and does not provide a way to specify a timezone (or force them to be in UTC).

Is this intentional? Are there plans to provide a way to force them to be sent in UTC?

mosesn:
I think this is a solid starter issue.  We might be able to rehydrate #191, add a test and merge it in.

vtatai:
@mosesn I will take a stab at it, hopefully will be able to get some progress.

mkantor:
Unless I'm misunderstanding it looks like this was solved way back in 2014 with 4258d70f60c956e6c6f361f84d84ef1a6e2b38dc.

eric:
Looks like this would work great to solve the problem. I’m not using finagle anymore but I’m glad this has a solution. 🎉

kevinoliver:
As noted [above](https://github.com/twitter/finagle/issues/190#issuecomment-345471631), this was fixed [back in 2014](https://github.com/twitter/finagle/commit/4258d70f60c956e6c6f361f84d84ef1a6e2b38dc).

PR #191, opened 2013-08-05
Provide a timezone for serializing timestamps

eric:
Adds a way to specify the timezone that dates and times are serialized in.

This is a proposed fix for #190.


eric:
Sorry, I ended up not having time to circle back on this and have moved to just using JDBC.

mosesn:
Bummer, thanks for the update.

PR #280, opened 2014-06-12
finagle-memcached: KetamaClientBuilder - expose the client type

fwbrasil:
In order to access the client nodes for operations like individual node stats.


mosesn:
@fwbrasil we've stopped following semver so religiously, so I think we'd be happy to merge this in.  you think it's still worth doing?

fwbrasil:
@mosesn I'd prefer to close it, I'm not sure if it'd be useful anymore.

mosesn:
Works for me.

PR #283, opened 2014-06-19
finagle-redis tests need refactoring to from specs2 to scalatest

penland365:
### WIP Do not merge!

The refactoring required for replacing all specs tests with scalatests requires a few intermediate code reviews.
This commit removes none of the original tests, but adds refacorted tests from the Naggati codec and the Client
Integration tests.  Code review is needed before continuing.

Tested w/ 
1. Scala 2.9.2
2. Redis 2.8.9
3. FreeBSD 10.0

Now that the formal stuff is out of the way, I  want to ensure that the path I'm driving down for the test refactoring doesn't end up with me going off a cliff Thelma and Louise style.  My goal on this path is:

1. DRY up some of the testing code
2. Separate out setup / tear down code from actual tests
3. Separate tests out by both functionality ( Codec / Client / Server ), and group them by Command.  For example, there is a current file named ```SimpleClientSuite.scala``` that simply removed the "simple" tests from ```ClientSpec.scala```.  However, I find this difficult to grasp in practice ( how does one qualify a "simple" test? ) and would like to change that hierarchy to command groupings as Redis does.  I show an example of this in the file ```KeyCommandsSuite.scala``` , which contains all codec tests for a command.

I'm sure I'm missing something or made something overly complicated.  Please yell at me in an orderly fashion.


travisbrown:
Now in [`develop`](https://github.com/twitter/finagle/commit/377e7ad38bc41a0d14a3d21ee3c9dd7cf6d53237) (see also #331).

PR #287, opened 2014-06-25
GSoC: finagle-smtp - initial

suncelesta:
This is initial version of SMTP client for Finagle, which I am working on during GSoC'14.
For now it supports sending independent commands to SMTP server and sending a simple text email to one ore more addresses (without cc, bcc and reply-to). Usage is described in `README.md`.


mosesn:
This finally graduated to the finagle organization, and can be found here: https://github.com/finagle/finagle-smtp

Issue #272, opened 2014-05-28
Support for Redis scripting commands


travisbrown:
[finagle-redis](https://github.com/twitter/finagle/tree/master/finagle-redis) currently only supports a subset of Redis commands. Adding [`EVAL` and the related Lua scripting commands](http://redis.io/commands/eval) would make some operations more convenient and efficient.

See [this mailing list thread](https://groups.google.com/d/msg/finaglers/CR17XdpnOWw/JInWl3hBqTEJ) for more context and discussion. You can find the currently supported commands [here](https://github.com/twitter/finagle/blob/master/finagle-redis/src/main/scala/com/twitter/finagle/redis/protocol/Command.scala).

timxzl:
I have a fully functional extension to finagle-redis that supports EVAL, EVALSHA, and other scripting commands. would be happy to share if it's interesting to folks.

mosesn:
Yeah, would definitely love to merge it in!

timxzl:
excellent! I have just created a pull request.

timxzl:
@mosesn I have managed to make the [pull request 381] (https://github.com/twitter/finagle/pull/381) pass all the tests on travis CI. feel free to contact me if you find any problem with it.

mosesn:
Thanks a lot for tackling this in https://github.com/twitter/finagle/pull/383 @timxzl!

PR #296, opened 2014-07-13
Support for Redis scripting commands #272

kashif:
initial attempt at #272  not working


mosesn:
@kashif still happy to help if you have any further questions :+1:

kashif:
thanks @mosesn  I have a question: how to I escape the quotes when wraping and unwraping strings?

The `EVAL` command needs a string as its script:

```
eval "return {KEYS[1],KEYS[2],ARGV[1],ARGV[2]}" 2 key1 key2 first second
```

but when parsing it, I get an error since it thinks that `"return`  is the first argument...

mosesn:
I think there's a bug in the parser where it always delimits on ' ' instead of treating stuff between quotes specially.  I'll take a closer look.

mosesn:
@kashif I threw together something that should be an OK first approximation.  I haven't tested it at all, so it may take some tweaking, but this patch should generally explain how to go about fixing the bug.

https://github.com/mosesn/finagle/compare/support-string-split.patch

mosesn:
ugh ok now it should work (although it's still totally untested)

NB: it doesn't handle quotes in the middle of a quoted area, you can't escape quotes, you can't have more than one space.

kashif:
@mosesn ah great I will try it out after dinner thank you!

mosesn:
@kashif did you have a chance to try it?

kashif:
@mosesn I seem to now get an error:

```
...
[error]         x EVAL
[error]           Unsupported command:  (Command.scala:270)
[error]           com.twitter.finagle.redis.protocol.Commands$$anonfun$doMatch$2.apply(Command.scala:270)
...
```

which might suggest that the parsing is now setting the `command` to just empty?

mosesn:
Can you push your stuff to this branch so I can try it out myself?

kashif:
all the stuff i have done till now is pushed here... I just added in your patch and tried to run the test:

```
$ ./sbt finagle-redis/test
...
```

mosesn:
OK, I made a pull request against your branch.  I had confused the semantics of filter / filterNot.  It should be fine now.

https://github.com/kashif/finagle/pull/1

kashif:
great that works! I will now add the rest of the `SCRIPT` commands

mosesn:
@kashif howdy, do you need any help from me?

luciferous:
@kashif How's this coming along? Anything we can do to assist?

mosesn:
@kashif hey, I just wanted to check in.  Is this something you're still interested in working on?

mosesn:
@kashif I think https://github.com/twitter/finagle/pull/383 solves the same problem, so I'm going to close this ticket once that merges in.  Thanks for your work on this PR–I'd be happy to work with you again in the future if you want to take another stab at a PR for finagle!

roanta:
This landed in https://github.com/twitter/finagle/commit/18a8387e6c1157595c77447f9942e46fed39c8a3

PR #320, opened 2014-11-03
Add documentation about tries stats scope

travisbrown:
Problem

The fact that the /success and /failure stats don't take retries into account
can lead to confusion, and the /tries scope isn't documented very well.

Solution

Add a paragraph about /tries and measuring success rate to the metrics section
of the users' guide.

Result

Less confusion about how to measure success accurately for clients that are
configured to retry.


luciferous:
What happened with this, did it get merged?

mosesn:
No, it exists in a mostly-complete state internally as a google doc, but I got pulled off on other things before I can finish.  lmk if you want to see it.

luciferous:
No rush, but would be nice to get it merged. How much more work on it you think?

mosesn:
Maybe just a day or two?  If you want to tackle it that would be rad.

vkostyukov:
Let me try to cherry-pick this for the internal review.

vkostyukov:
Oh. Just re-read @mosesn's comment. Okay, how about to close this one and file an internal ticket to keep an eye on this?

luciferous:
sgtm

vkostyukov:
Okay. I've created an internal issue to track the progress on tihs. Closing it now.

PR #329, opened 2014-12-20
Update statistics for org.jboss.netty.channel.FileRegion writes

whiter4bbit:
Handle FileRegion writes in ChannelStatsHandler

### Problem 

Currently, when writing instance of `FileRegion` (in custom Codec for example), handler logs many annoying messages:
```
WAR [20141220-05:51:59.151] channel: ChannelStatsHandler received non-channelbuffer write
```
And sent bytes count from FileRegion  not included in `sent_bytes` counter.

### Solution

Check if message is instance of `FileRegion` and increment stats counters by `FileRegion.getCount` value

### Result

Sent bytes from FileRegion are included in `sent_bytes` and `connection_sent_bytes` counters

### DIFF
--- a/finagle-core/src/main/scala/com/twitter/finagle/channel/ChannelStatsHandler.scala
+++ b/finagle-core/src/main/scala/com/twitter/finagle/channel/ChannelStatsHandler.scala
@@ -9,7 +9,7 @@ import java.util.concurrent.atomic.{AtomicInteger, AtomicLong}
 import java.util.logging.{Level, Logger}
 import org.jboss.netty.buffer.ChannelBuffer
 import org.jboss.netty.channel.{ChannelHandlerContext, ChannelStateEvent,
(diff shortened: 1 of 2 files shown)


trustin:
This pull request itself looks good to me.

However, I think it's not a good idea to increase the counters at `writeRequested()`, because it only tells us when a write operation has been requested, not when the write operation has been actually fulfilled.

I would recommend overriding `SimpleChannelUpstreamHandler.writeComplete()` and get the number of written bytes via `WriteCompletionEvent.getWrittenAmount()`.

mosesn:
@whiter4bbit I just wanted to make sure this PR didn't get lost between the cracks.

whiter4bbit:
was out of time, I will try to 're-implement things in way that @trustin proposed

mosesn:
Thanks!

whiter4bbit:
Created new PR: https://github.com/twitter/finagle/pull/355

PR #331, opened 2014-12-31
Complete migration of finagle-redis from specs to ScalaTest

travisbrown:
Problem

Jeff Davis [has migrated](https://github.com/twitter/finagle/pull/283) the majority of finagle-redis's tests from specs to ScalaTest, but there are a few remaining gaps and some formatting issues.

Solution

This commit completes the migration, changes multi-line asserts (with separate definitions for actual and expected values) to inlined asserts, and removes remaining specs tests and dependencies. It removes the outdated (and unused) `BtreeClientSpec` without a ScalaTest replacement. It also adds ignore annotations to all integration tests and removes the filter from the SBT configuration.

Result

Finagle does not depend on specs.


travisbrown:
Now in [`develop`](https://github.com/twitter/finagle/commit/377e7ad38bc41a0d14a3d21ee3c9dd7cf6d53237).

PR #337, opened 2015-01-30
ugh how is anyone supposed to even understand finagle?!!!

olix0r:
# PROBLEM:
    
I was trying to use Finagle but I had to keep looking at a dictionary because Finagle kept using this word "Request" and I don't know what it means.  This makes it virtually impossible for new programmers to get involved in Finagle.  I feel like Twitter did this specifically to discourage me from understanding Finagle.
    
# SOLUTION:
    
Replace all instances of 'Request' with 'Ask'.
    
# RESULT:
    
Finagle is now *much* easier to understand and it MUCH easier to type things.


mosesn:
LGTM, but this is a pretty big change, and I don't think we have the bandwidth to cutover everyone to use the new API.  Maybe make another module that people can depend on?  Obviously we'll need to do this for finagle-httpx also.  Maybe finagle-http-ask and finagle-httpx-ask?  That way we can slowly phase out finagle-http{,x} without causing too much trouble.

olix0r:
@mosesn i understand your concerns but do you really think people are going to use finagle without this change?  i'd consider this a serious blocker to widespread FOSS adoption.

bmdhacks:
This is an excellent pull-ask

coops:
burn 1 2 dis

nshkrob:
I'm going to wait for a green build, but shouldn't you also replace Response with Answer?

mariusae:
This is a good start, but why stop here? I really think that the result of an Ask is a Learning, so, the type variables should be: Service[Ask, Learning]. 

This also teaches an important life lesson: an ask today equals a Future[Learning]

olix0r:
@nshkrob perhaps ```Services[Ask, Learning]``` best captures this?

travisbrown:
I thought we decided no punctuation on the "Problem" / "Solution" / "Result" headers in the commit message? Please consult CONTRIBUTING.md in the future @olix0r.

vkostyukov:
@olix0r since you're here, can also replace `c.t.u.Future` with `scala.concurrent.Future`?

sonjake:
We can take holistic approach here and synergize until we reach consensus. I'll touch base with our internal stakeholders and cycle back.

mariusae:
@travisbrown Those headings are admittedly outdated. They probably should be named Customer ask/Business Plan/Business Case/

softprops:
Without additional context, perhaps you can define a package level type alias

```scala
type Argument = Service[Ask, Learning]
```

For which you could derive concrete instances of Strawmen.

teodimoff:
Guys, whats wrong with Request, Response types?

dschobel:
@teodimoff the core issue is that "Request" / "Response" are technical jargon which are practically meaningless in the real world. I like this change since finagle shouldn't require a formal CS background to use and enjoy.

travisbrown:
The entire Finagle team set aside today to discuss this change, and after a lot of debate we've decided to close this PR as `wont-fix`. Sorry @olix0r. If you feel the need to create a more pragmatically-focused Finagle fork, we can talk about whether the [Finagle organization](https://github.com/finagle) would be an appropriate place to host it.

PR #339, opened 2015-02-02
Add get function for ThriftClientBufferedCodec.

tonyd3:
Add get function for ThriftClientBufferedCodec.

### DIFF
--- a/finagle-thrift/src/main/scala/com/twitter/finagle/thrift/ThriftClientBufferedCodec.scala
+++ b/finagle-thrift/src/main/scala/com/twitter/finagle/thrift/ThriftClientBufferedCodec.scala
@@ -28,6 +28,8 @@ object ThriftClientBufferedCodec {
    */
   def apply() = new ThriftClientBufferedCodecFactory
 
+  def get(): ThriftClientBufferedCodecFactory = apply()
+  
   def apply(protocolFactory: TProtocolFactory) =
     new ThriftClientBufferedCodecFactory(protocolFactory)
 
--- a/finagle-thrift/src/test/java/com/twitter/finagle/thrift/ThriftClientCodecTest.java
+++ b/finagle-thrift/src/test/java/com/twitter/finagle/thrift/ThriftClientCodecTest.java
@@ -1,0 +1,31 @@
+/**
+ * @copyright - Copyright (C) 2015 TellApart, Inc. All Rights Reserved.
+ */
+package com.twitter.finagle.thrift;
+
+import com.twitter.finagle.stats.DefaultStatsReceiver;
+
+import org.junit.Assert;
+import org.junit.Test;
+
+import scala.Option;
+
+public class ThriftClientCodecTest {
+
+  @Test
+  public void testCodecFactoryCreation() {
+    // ThriftClientBufferedCodec
+    Assert.assertTrue(ThriftClientBufferedCodec.get() instanceof ThriftClientBufferedCodecFactory);
+    Assert.assertTrue(
+        ThriftClientBufferedCodec.apply() instanceof ThriftClientBufferedCodecFactory);
+    Assert.assertTrue(
+        ThriftClientBufferedCodec.apply(Protocols.factory(DefaultStatsReceiver.self()))
+            instanceof ThriftClientBufferedCodecFactory);
+
+    //ThriftClientFramedCodec
+    Assert.assertTrue(ThriftClientFramedCodec.get() instanceof ThriftClientFramedCodecFactory);
+    Assert.assertTrue(
+        ThriftClientFramedCodec
+            .apply(Option.<ClientId>empty()) instanceof ThriftClientFramedCodecFactory);
+  }
+}
--- a/finagle-thrift/src/test/java/com/twitter/finagle/thrift/ThriftClientCodecTest.java
+++ b/finagle-thrift/src/test/java/com/twitter/finagle/thrift/ThriftClientCodecTest.java
@@ -1,6 +1,3 @@
-/**
- * @copyright - Copyright (C) 2015 TellApart, Inc. All Rights Reserved.
- */
 package com.twitter.finagle.thrift;
 
 import com.twitter.finagle.stats.DefaultStatsReceiver;


finagle-thrift/src/main/scala/com/twitter/finagle/thrift/ThriftClientBufferedCodec.scala
```diff
@@ -28,6 +28,8 @@ object ThriftClientBufferedCodec {
    */
   def apply() = new ThriftClientBufferedCodecFactory
 
+  def get() = apply()
```
mosesn: can you add a return type here?  public methods should be annotated (it's OK not to fix up the rest of the file).

mosesn:
Sorry, could you follow the problem / solution / result guidelines in [CONTRIBUTED.md][0]?  I don't think I quite understand where you're coming from with this review.

[0]: https://github.com/twitter/finagle/blob/master/CONTRIBUTING.md#style

tonyd3:
Problem
ThriftClientBufferedCodec class extends ThriftClientBufferedCodec, so when we call ThriftClientBufferedCodec.get() it actually returns ThriftClientFramedCodec

Solution
Implement get function in the ThriftClientBufferedCodec object so that it returns ThriftClientBufferedCodec.

Result
You should be able to call ThriftClientBufferedCodec.get() and return a correct value, which is very useful for Java development.

mosesn:
I don't understand "so when we call ThriftClientBufferedCodec.get() it actually returns ThriftClientFramedCodec".

ThriftClientBufferedCodec doesn't have a get method yet, so there's no way you could call it today.

What does the get() method get you that you can't use apply() for?

tonyd3:
I actually haven't tried it in scala, but in java here is the problem I encountered. 
ClientBuilder.get().codec(ThriftClientFramedCodec.get())
This give me the correct codec.
but when I tried 
ClientBuilder.get().codec(ThriftClientBufferedCodec.get()) 
it actually uses the get function in the ThriftClientFramedCodec object because the get function doesn't exist in ThriftClientFramedCodec. The apply() function works and returns the correct codec, and that's what I have now. 

I haven't tested if it's the same issue in scala, but I guess even if it is, it doesn't come up because people never call the get function.

mosesn:
So I checked out the bytecode, in case I might have missed something, but it seems pretty unambiguous.

```
 mnakamura@tw-mbp13-mnakamura ⮀ ~ ⮀ javap -classpath ~/.m2/repository/com/twitter/finagle-thrift_2.10/6.21.0/finagle-thrift_2.10-6.21.0.jar com.twitter.finagle.thrift.ThriftClientFramedCodec
Picked up JAVA_TOOL_OPTIONS: -Dfile.encoding=utf8
Compiled from "ThriftClientFramedCodec.scala"
public class com.twitter.finagle.thrift.ThriftClientFramedCodec implements com.twitter.finagle.Codec<com.twitter.finagle.thrift.ThriftClientRequest, byte[]> {
  public static scala.Option<com.twitter.finagle.thrift.ClientId> apply$default$1();
  public static boolean $lessinit$greater$default$4();
  public static scala.Option<com.twitter.finagle.thrift.ClientId> $lessinit$greater$default$3();
  public static com.twitter.finagle.thrift.ThriftClientFramedCodecFactory get();
  public static com.twitter.finagle.thrift.ThriftClientFramedCodecFactory apply(scala.Option<com.twitter.finagle.thrift.ClientId>);
  public com.twitter.finagle.ServiceFactory<com.twitter.finagle.thrift.ThriftClientRequest, byte[]> prepareServiceFactory(com.twitter.finagle.ServiceFactory<com.twitter.finagle.thrift.ThriftClientRequest, byte[]>);
  public com.twitter.finagle.transport.Transport<java.lang.Object, java.lang.Object> newClientTransport(org.jboss.netty.channel.Channel, com.twitter.finagle.stats.StatsReceiver);
  public com.twitter.finagle.Service<com.twitter.finagle.thrift.ThriftClientRequest, byte[]> newClientDispatcher(com.twitter.finagle.transport.Transport<java.lang.Object, java.lang.Object>);
  public com.twitter.util.Closable newServerDispatcher(com.twitter.finagle.transport.Transport<java.lang.Object, java.lang.Object>, com.twitter.finagle.Service<com.twitter.finagle.thrift.ThriftClientRequest, byte[]>);
  public boolean failFastOk();
  public org.jboss.netty.channel.ChannelPipelineFactory pipelineFactory();
  public com.twitter.finagle.ServiceFactory<com.twitter.finagle.thrift.ThriftClientRequest, byte[]> prepareConnFactory(com.twitter.finagle.ServiceFactory<com.twitter.finagle.thrift.ThriftClientRequest, byte[]>);
  public com.twitter.finagle.thrift.ThriftClientFramedCodec(org.apache.thrift.protocol.TProtocolFactory, com.twitter.finagle.ClientCodecConfig, scala.Option<com.twitter.finagle.thrift.ClientId>, boolean);
}
 mnakamura@tw-mbp13-mnakamura ⮀ ~ ⮀ javap -classpath ~/.m2/repository/com/twitter/finagle-thrift_2.10/6.21.0/finagle-thrift_2.10-6.21.0.jar com.twitter.finagle.thrift.ThriftClientBufferedCodec
Picked up JAVA_TOOL_OPTIONS: -Dfile.encoding=utf8
Compiled from "ThriftClientBufferedCodec.scala"
public class com.twitter.finagle.thrift.ThriftClientBufferedCodec extends com.twitter.finagle.thrift.ThriftClientFramedCodec {
  public final org.apache.thrift.protocol.TProtocolFactory com$twitter$finagle$thrift$ThriftClientBufferedCodec$$protocolFactory;
  public static com.twitter.finagle.thrift.ThriftClientBufferedCodecFactory apply(org.apache.thrift.protocol.TProtocolFactory);
  public static com.twitter.finagle.thrift.ThriftClientBufferedCodecFactory apply();
  public org.jboss.netty.channel.ChannelPipelineFactory pipelineFactory();
  public com.twitter.finagle.thrift.ThriftClientBufferedCodec(org.apache.thrift.protocol.TProtocolFactory, com.twitter.finagle.ClientCodecConfig);
}
```

There is no get method on `ThriftClientBufferedCodec`, so I think the reason why it returned the `ThriftClientFramedCodec` was because your code wasn't able to compile.

Using apply seems fine to me.  What was the problem you ran into with apply?

tonyd3:
ThriftClientBufferedCodec extends ThriftClientFramedCodec. So since ThriftClientBufferedCodec doesn't have it's own get function, ThriftClientBufferedCodec.get() calls its parent's get(), which returns an instance of ThriftClientFramedCodec.

mosesn:
Oh god, that's horrible.  I had totally forgotten you inherit static methods in java.  OK, LGTM.  Could you add a comment that the problem is that because an object's methods get added as static forwarders on the class that it's a companion of, and static methods are inherited, this must be done so that you don't accidentally inherit get from ThriftClientFramedCodec?

It might make sense to add a test too.

tonyd3:
ping

kevinoliver:
lgtm. I'll take this over and should get it applied internally.

tonyd3:
Thanks Kevin!

kevinoliver:
Ok this has been merged locally. Should land on the develop branch soon. Thanks for the patch.

dschobel:
merged in https://github.com/twitter/finagle/commit/392e3ed01d0c7e78cd5de8a999861e68e0b045c3. thanks @tonyd3

PR #341, opened 2015-02-06
[WIP] Typeclass-based value extraction for finagle-mysql datatypes.

folone:
Here's one piece of code we found useful at SoundCloud when converting between scala types and finagle-mysql types. /cc @fwbrasil @williamboxhall

```scala
> finagle-mysql/console
[info] Starting scala interpreter...
[info]
Welcome to Scala version 2.10.4 (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_20).
Type in expressions to have them evaluated.
Type :help for more information.

scala> import com.twitter.finagle.exp.mysql._
import com.twitter.finagle.exp.mysql._

scala> NullValue.mapToOpt[String]
res0: Option[String] = None

scala> StringValue("Hej").mapTo[String]
res1: String = Hej

scala> StringValue("Hej").mapTo[Option[Int]]
res2: Option[Int] = None

scala> StringValue("Hej").mapToOpt[String]
res3: Option[String] = Some(Hej)

scala> EmptyValue.mapTo[Int]
java.lang.RuntimeException: Cannot extract a requested value from EmptyValue
```

### DIFF
--- a/finagle-mysql/src/main/scala/com/twitter/finagle/mysql/TypeExtractor.scala
+++ b/finagle-mysql/src/main/scala/com/twitter/finagle/mysql/TypeExtractor.scala
@@ -1,0 +1,71 @@
+package com.twitter.finagle.exp.mysql
+
+import annotation.implicitNotFound
+import java.sql.Timestamp
+
+@implicitNotFound(msg = "Don't know how to extract ${T} from a Value.")
+trait TypeExtractor[T] {
+  def extractFrom(v: Value): Option[T]
+}
+
+object TypeExtractor {
+  implicit object StringExtractor extends TypeExtractor[String] {
+    def extractFrom(v: Value): Option[String] = v match {
+      case StringValue(s) => Some(s)
+      case _              => None
+    }
+  }
+  implicit object BooleanExtractor extends TypeExtractor[Boolean] {
+    def extractFrom(v: Value): Option[Boolean] = v match {
+      case ByteValue(b) => Some(b == 1)
+      case _            => None
+    }
+  }
+  implicit object ByteExtractor extends TypeExtractor[Byte] {
+    def extractFrom(v: Value): Option[Byte] = v match {
+      case ByteValue(b) => Some(b) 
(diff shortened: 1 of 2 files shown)


finagle-mysql/src/main/scala/com/twitter/finagle/mysql/TypeExtractor.scala
```diff
@@ -0,0 +1,71 @@
...
+      case _            => None
+    }
+  }
+  implicit object ShortExtractor extends TypeExtractor[Short] {
+    def extractFrom(v: Value): Option[Short] = v match {
+      case ShortValue(s) => Some(s)
+      case _             => None
+    }
+  }
+  implicit object IntMapeer extends TypeExtractor[Int] {
```
mosesn: these should be `Mapper`, not `Mapeer`

folone: that's right, sorry for that. fix is coming in a sec.

fwbrasil:
Nice addition! Can we have tests for it?

mosesn:
Hmm, neat!  I have to say though, I wonder if maybe this would make more sense as just a set of [Bijections][0]?  Then we'll also get things like 

```
implicit val dateToTime: Bijection[Date, Time] = ...
implicit val timeStampToTime: Bijection[Timestamp, Time] = ...

Timestamp((Time.now - 5.minutes) - Time.zero).as[Date]
```

We would have to add bijection-core as a dep, but that seems not so bad.

[0]: https://github.com/twitter/bijection

folone:
@mosesn That does look very interesting, will look into that. I'll mark this PR as WIP in the meantime.

williamboxhall:
Nice one @folone ! Also agree Bijections could be nice too

folone:
@mosesn I've opened a PR in the bijection project: https://github.com/twitter/bijection/pull/197.
Closing this one.

PR #342, opened 2015-02-08
Target move database should be an integer string

mrvisser:
Problem

Redis integration tests are failing with an exception. Usage of the "move"
command expects an integer string as the second parameter, but we give it
alphabet characters. Redis' response to this is undefined.

Solution

Changed the second parameter to the `toDb` index, which was probably the
original intention.

Result

Integration tests now pass.


mosesn:
LGTM

mrvisser:
I've just found Travis' ScalaTest conversion seems to address this as well https://github.com/twitter/finagle/pull/331 but it's not clear the timeline on that. Would recommend just closing this PR without merge in favor of not delaying his w/ more merge conflicts.

mosesn:
Oh, his has actually been merged already internally, but we've been having trouble with pushing the internal version to github. I'll just close this ticket though.  Good looking out!

PR #344, opened 2015-02-17
Add get Function to Aperture.scala for Java access

tonyd3:
Problem
We want to use the Aperture LoadBalancerFactory in a Java client. 

Solution
Implement a get function in the object, so it can have Java access.

Result
You should be able to call ApertureBalancerFactory.get() and return the object, which is very useful for Java development.

### DIFF
--- a/finagle-core/src/main/scala/com/twitter/finagle/loadbalancer/Aperture.scala
+++ b/finagle-core/src/main/scala/com/twitter/finagle/loadbalancer/Aperture.scala
@@ -28,6 +28,11 @@ object ApertureBalancerFactory extends WeightedLoadBalancerFactory {
     new ApertureLoadBandBalancer(activity,
       statsReceiver = statsReceiver,
       emptyException = emptyException)
+  
+  /**
+   * Used for Java access.
+   */
+  def get() = this
 }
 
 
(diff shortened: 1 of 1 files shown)


kevinoliver:
lgtm. i'll get this applied locally.

tonyd3:
Thanks!

finagle-core/src/main/scala/com/twitter/finagle/loadbalancer/Aperture.scala
```diff
@@ -28,6 +28,11 @@ object ApertureBalancerFactory extends WeightedLoadBalancerFactory {
     new ApertureLoadBandBalancer(activity,
       statsReceiver = statsReceiver,
       emptyException = emptyException)
+  
+  /**
+   * Used for Java access.
+   */
+  def get() = this
```
mosesn: can we add an explicit annotation here?

tonyd3: Done

kevinoliver:
Thanks.

This has been merged locally. Once we get our github push issue sorted out, it'll show up on the develop branch.

dschobel:
merged in https://github.com/twitter/finagle/commit/27aa087dc93dc635265905a09debbdc39ff2076f. thanks for the contribution @tonyd3.

PR #346, opened 2015-02-26
finagle-core: Set finite SSL handshake timeout. Fixes #345

### LINKED ISSUE #345: No handshake timout on SSLHandler
`Netty3Transporter` does not set explicit handshake timeout on `SSLHandler`.

If a channel configured this way establishes TCP connection with a remote that does not respond to a handshake, the channel would get stuck in "connecting" state indefinitely.

This may cause various problems. One that I'm aware of is with the `FailFastFactory`.
When `FailFastFactory` tries to establish a connection it sets no timeouts on its own [1], thus it could be driven into a permanent fail-fast state, even if the connectivity with the remote service is long ago reestablished.

FYI, Netty 4 has a default handshake timeout of 10 seconds that can be overridden [2].
Unfortunately, it should be done explicitly in Netty 3, and the only way to do it is via a multi-arg `SSLHandler` constructor [3].

---

[1] – https://github.com/twitter/finagle/blob/6.22.0/finagle-core/src/main/scala/com/twitter/finagle/service/FailFastFactory.scala#L105
[2] – https://github.com/netty/netty/commit/817309c#diff-2fe7b22a8d650f1ea0bf56a809c061f9R237
[3] – https://github.com/netty/netty/blob/netty-3.6.4.Final/src/main/java/org/jboss/netty/handler/ssl/SslHandler.java#L370

sirmax:
Problem

Netty3Transporter does not set explicit handshake timeout on SslHandler.

If a channel configured this way establishes TCP connection with a
remote that does not respond to a handshake, the channel would get stuck
in "connecting" state indefinitely.

Solution

Add Transporter.SSLHandshakeTimeout stack param. Set its value to
10 seconds by default. Use this param to configure SslHandler.

Result

Connections do not get stuck in SSL handshake.
Fixes #345

### DIFF
--- a/finagle-core/src/main/scala/com/twitter/finagle/netty3/Netty3Transporter.scala
+++ b/finagle-core/src/main/scala/com/twitter/finagle/netty3/Netty3Transporter.scala
@@ -1,5 +1,6 @@
 package com.twitter.finagle.netty3
 
+import com.twitter.conversions.time._
 import com.twitter.finagle.channel.{ChannelRequestStatsHandler, ChannelStatsHandler, IdleChannelHandler}
 import com.twitter.finagle.client.Transporter
 import com.twitter.finagle.httpproxy.HttpConnectHandler
@@ -123,6 +124,7 @@ object Netty3Transporter {
     val ChannelFactory(cf) = params[ChannelFactory]
     val TransportFactory(newTransport) = params[TransportFactory]
     val Transporter.ConnectTimeout(connectTimeout) = params[Transporter.ConnectTimeout]
+    val Transporter.SSLHandshakeTimeout(sslHandshakeTimeout) = params[Transporter.SSLHandshakeTimeout]
     val Transporter.TLSHostname(tlsHostname) = params[Transporter.TLSHostname]
     val Transporter.HttpProxy(httpProxy) = params[Transporter.HttpProxy]
     val Transporter.SocksProxy(socksProxy, credentials) = params[Transporter.SocksProxy]
@@ -139,7 +141,9 @@ object Netty3Transporter {
       pipelineFactory,
       newChannel = cf.newChannel(_),
       newTransport = (ch: Channel) => newTransport(ch).cast[In, Out],
-      tlsConfig = tls map { case engine => Netty3TransporterTLSConfig(engine, tlsHostname) },
+      tlsConfig = tls map { engine =>
+        Netty3TransporterTLSConfig(engine, tlsHostname, sslHandshakeTimeout)
+      },
       httpProxy = httpProxy,
       socksProxy = socksProxy,
       socksUsernameAndPassword = credentials,
@@ -191,7 +195,7 @@ object Netty3Transporter {
  * against the given value.
  */
 case class Netty3TransporterTLSConfig(
-  newEngine: SocketAddress => Engine, verifyHost: Option[String])
+  newEngine: SocketAddress => Engine, verifyHost: Option[String], handshakeTimeout: Duration = 10.seconds)
 
 /**
  * A transporter for netty3 which, given an endpoint name (socket
@@ -281,7 +285,7 @@ case class Netty3Transporter[In, Out](
         new IdleStateHandler(DefaultTimer, rms, wms, 0, TimeUnit.MILLISECONDS))
     }
 
-    for (Netty3TransporterTLSConfig(newEngine, verifyHost) <- tlsConfig) {
+    for (Netty3TransporterTLSConfig(newEngine, verifyHost, handshakeTimeout) <- tlsConfig) {
       import org.jboss.netty.handler.ssl._
 
       val engine = newEngine(addr)
@@ -292,7 +296,15 @@ case class Netty3Transporter[In, Out](
         Function.const(None) _
       }
 
-      val sslHandler = new SslHandler(engine.self)
+      val sslHandler = {
+        val startTls = false
+        new SslHandler(
+          engine.self,
+          SslHandler.getDefaultBufferPool,
+          startTls,
+          DefaultTimer,
+          handshakeTimeout.inMillis)
+      }
       val sslConnectHandler = new SslConnectHandler(sslHandler, verifier)
 
       pipeline.addFirst("sslConnect", sslConnectHandler)
--- a/finagle-core/src/test/scala/com/twitter/finagle/netty3/Netty3TransporterTest.scala
+++ b/finagle-core/src/test/scala/com/twitter/finagle/netty3/Netty3TransporterTest.scala
@@ -236,6 +236,29 @@ class Netty3TransporterTest extends FunSpec with MockitoSugar {
 
         assert(!channel.isOpen)
       }
+
+      describe ("handshake timeout") {
+        val pipelineFactory = Channels.pipelineFactory(Channels.pipeline())
+        val tlsParams = Stack.Params.empty + Transport.TLSClientEngine(Some(_ => Engine(mock[SSLEngine])))
+
+        it ("should be finite by default") {
+          val transporter = Netty3Transporter.make[Int, Int](pipelineFactory, tlsParams)
+          val pipeline = transporter.newPipeline(null, NullStatsReceiver)
+          val sslHandler = pipeline.get(classOf[SslHandler])
+
+          assert(sslHandler.getHandshakeTimeout > 0)
+        }
+
+        it ("should be configurable with stack params") {
+          val timeout = 1.second
+          val customParams = tlsParams + Transporter.SSLHandshakeTimeout(timeout)
+          val transporter = Netty3Transporter.make[Int, Int](pipelineFactory, customParams)
+          val pipeline = transporter.newPipeline(null, NullStatsReceiver)
+          val sslHandler = pipeline.get(classOf[SslHandler])
+
+          assert(sslHandler.getHandshakeTimeout == timeout.inMillis)
+        }
+      }
     }
   }
 }
--- a/CHANGES
+++ b/CHANGES
@@ -71,6 +71,10 @@ Runtime Behavior Changes
     includes time spent in name resolution, which is now covered by
     "namer/bind_latency_us" as discussed above.
 
+  * finagle-core: `com.twitter.finagle.netty3.Netty3Transporter` now sets a 10 second
+    SSL handshake timeout to avoid getting stuck in "connecting" state. This timeout
+    may be reconfigured using `c.t.f.client.Transporter.SSLHandshakeTimeout` stack param.
+
 Deprecations
 ~~~~~~~~~~~~
 
--- a/finagle-core/src/main/scala/com/twitter/finagle/client/Transporter.scala
+++ b/finagle-core/src/main/scala/com/twitter/finagle/client/Transporter.scala
@@ -54,6 +54,17 @@ object Transporter {
   }
 
   /**
+   * $param SSL handshake timeout, if TLS is enabled.
+   */
+  case class SSLHandshakeTimeout(howlong: Duration) {
+    def mk(): (SSLHandshakeTimeout, Stack.Param[SSLHandshakeTimeout]) =
+      (this, SSLHandshakeTimeout.param)
+  }
+  object SSLHandshakeTimeout {
+    implicit val param = Stack.Param(SSLHandshakeTimeout(10.seconds))
+  }
+
+  /**
    * $param hostname verification, if TLS is enabled.
    * @see [[com.twitter.finagle.transport.Transport#TLSEngine]]
    */


sirmax:
Two things to note:

- I have a ["neighbour" test](https://github.com/twitter/finagle/blob/8616595/finagle-core/src/test/scala/com/twitter/finagle/netty3/Netty3TransporterTest.scala#L192) crashing in SBT on assertion in Netty (and had it crashing prior to the fix):

```
    - should close the channel if the remote peer closed TLS session *** FAILED ***
      java.lang.AssertionError:
      at org.jboss.netty.handler.ssl.SslHandler.decode(SslHandler.java:853)
      at org.jboss.netty.handler.codec.frame.FrameDecoder.callDecode(FrameDecoder.java:425)
      at org.jboss.netty.handler.codec.frame.FrameDecoder.messageReceived(FrameDecoder.java:303)
      at org.jboss.netty.channel.SimpleChannelUpstreamHandler.handleUpstream(SimpleChannelUpstreamHandler.java:70)
      at com.twitter.finagle.netty3.Netty3TransporterTest$$anonfun$1$$anonfun$apply$mcV$sp$12$$anonfun$apply$mcV$sp$13.apply$mcV$sp(Netty3TransporterTest.scala:234)
```

In IDEA's runner it is green. I didn't investigate it further.

- `SslHandler` requires a timer. I used `DefaultTimer` as the rest of the `Netty3Transporter` does, but it probably would be better to use the one from `com.twitter.finagle.param.Timer`.

vkostyukov:
This is odd that the neighbor test is failing. Let's hear from TravisCI.

finagle-core/src/test/scala/com/twitter/finagle/netty3/Netty3TransporterTest.scala
```diff
@@ -189,7 +189,7 @@ class Netty3TransporterTest extends FunSpec with MockitoSugar {
     }
 
     describe("SslHandler") {
-      it ("should close the channel if the remote peer closed TLS session") {
+      def mockSslEngine = {
```
vkostyukov: I looked at the failed assert:

```java
  assert !inNetBuf.hasRemaining() || engine.isInboundDone();
```

I'm not sure we run the tests with assertion enabled. So, could you please add the following configuration to the mock engine:

```java
when(engine.isInboundDone) thenReturn true
```

and check the `sbt +test` locally?

sirmax:
Yup, `when(engine.isInboundDone) thenReturn true` fixes it. However I'm not entirely happy with such mocking, because it might break some yet-to-be-written tests in an unexpected way.

What's out options here:
- Add this mock anyway with a comment regarding the possible errors for those who might want to reuse it.
- Write a conditional mock that would return `true` only when some (but what exactly?) condition is met. Downside – this would get ugly immediately.
- Do nothing and make a mental notice that the tests should be run with assertions disabled.

So, WDYT?

vkostyukov:
I wrote that neighbor and totally forgot about assertions. In that case this "fix" with `when(engine.isInboundDone) thenReturn true` is correct since the test is assuming that SSL engine is closed (which I was trying to configure in mock engine).

Regarding the new test (the one that tests timeout). I think we shouldn't reuse that engine here since you doesn't actually sends anything to it. I would suggest to return the origin SSL engine inside the neighbor tests (+ new mock configuration on `isInboundDone`) and create a new and super simple one (like `val engine = mock[SSLEngine]`) in the new test. 

Does it sounds reasonable?

mosesn:
:+1: to don't worry about it and just don't let tests share state.

sirmax:
@vkostyukov:
>  I would suggest to move the origin SSL engine inside the neighbor tests (+ new mock configuration on `isInboundDone`) and create a new and super simple one (like `val engine = mock[SSLEngine]`) in the new test.
>
> Does it sounds reasonable?

It does. I've updated the PR accordingly.

vkostyukov:
:+1: Looks good to me.

stevegury:
LGTM

finagle-core/src/main/scala/com/twitter/finagle/netty3/Netty3Transporter.scala
```diff
@@ -139,7 +141,7 @@ object Netty3Transporter {
       pipelineFactory,
       newChannel = cf.newChannel(_),
       newTransport = (ch: Channel) => newTransport(ch).cast[In, Out],
-      tlsConfig = tls map { case engine => Netty3TransporterTLSConfig(engine, tlsHostname) },
+      tlsConfig = tls map { case engine => Netty3TransporterTLSConfig(engine, tlsHostname, sslHandshakeTimeout) },
```
dschobel: I know you didn't add it but the 'case' keyword isn't needed so do you mind cleaning it up while you're here?

finagle-core/src/main/scala/com/twitter/finagle/netty3/Netty3Transporter.scala
```diff
@@ -292,7 +294,7 @@ case class Netty3Transporter[In, Out](
         Function.const(None) _
       }
 
-      val sslHandler = new SslHandler(engine.self)
+      val sslHandler = new SslHandler(engine.self, SslHandler.getDefaultBufferPool, false, DefaultTimer, handshakeTimeout.inMillis)
```
dschobel: can you please reformat per http://docs.scala-lang.org/style/indentation.html so that the line doesn't exceed ~100 chars?

also, can you use named params to give that boolean literal some context?

dschobel:
this is looking good, can you add a note to the runtime behavior changes section of CHANGES covering the new default timeout?

mosesn:
LGTM

sirmax:
@dschobel, reformatted, added a CHANGES entry. Couldn't use named args with `SslHandler` since it's Java, added `val startTls = false` instead.

sirmax:
You know, guys, there is one thing that bothers me. 

A handshake may be considered as a part of the connection process – both Netty's `SslHandler` and Finagle's own `SslConnectHandler` actually delay "connected" event propagation until the handshake is done. Thus it may have sense to cover the handshake with `Transporter.ConnectTimeout` instead of adding an implementation specific param.

Maybe instead of setting up timeouts in `SslHandler` and other places, might the need appear, Finagle should handle the connection timeout itself? It would bring a small overhead of course, and is a bit trickier, but maybe this it the right way?

dschobel:
I think if we had no existing clients it would be the right thing to do but expanding the scope of Transporter.ConnectTimeout to cover those additional twelve packet transmissions of ssl handshake would likely break clients applying tight timeouts tuned against the existing narrower scope. I think it’s better to keep this as an expert knob that maps directly to the netty param, as written.

finagle-core/src/test/scala/com/twitter/finagle/netty3/Netty3TransporterTest.scala
```diff
@@ -236,6 +236,29 @@ class Netty3TransporterTest extends FunSpec with MockitoSugar {
 
         assert(!channel.isOpen)
       }
+
+      describe ("handshake timeout") {
```
dschobel: not necessary but if you're feeling ambitious... you could write a test case covering the symptom you describe in https://github.com/twitter/finagle/issues/345, specifically, a hung ssl handshake causing the channel to get stuck in the connecting state and showing how this new timeout closes the channel.

mosesn:
I think the ssl handshake section is already part the service acquisition timeout, but the hook to fail isn't setup.  I think you're right that regular service acquisition timeout should fail the ssl handshake.  Do you think you might have a use case for setting an explicitly fast ssl handshake timeout?  If not, then we should do what you suggested.

@dschobel there's precedence for having connectTimeout (cf tcpConnectTimeout) so I think that making it simpler and more consistent is useful enough to risk breaking someone who was relying on existing behavior.

vkostyukov:
Hi @sirmax and all!

I think that @sirmax is right and we might want to reuse [`Transporter.ConnectTimeout`](https://github.com/twitter/finagle/blob/master/finagle-core/src/main/scala/com/twitter/finagle/netty3/Netty3Transporter.scala#L127) param instead of introducing a new (implementation-specific) one. 

For now, this param is used in `Netty3Transported` and propagates to `Channel` options via hash-map:
```scala
o += "connectTimeoutMillis" -> ((connectTimeout + compensation).inMilliseconds: java.lang.Long)
```

That's being said, it's totally safe to reuse it in `SslHandler` since at the moment of SSL handshake channel is already connected.

What do you think?

sirmax:
Hi guys!

First of all I'm sorry for going silent for so long. Been pretty busy :(

@vkostyukov just to clarify, do you suggest reusing `Transporter.ConnectTimeout` for SSL handshake timeout?

While this would be, probably, the easiest thing to do, it would effectively double the timeout. For me personally this is not a problem at all, but I can't speak for all the Finagle users out there, so it is your choice to make.

mosesn:
Yeah, I'd rather not pass it as a param to netty that way.  Could we use the finagle interruption mechanism to interrupt it instead?  We know the interrupt will come already, we just don't have anything hooked up actually cancel it after the interrupt is sent.

sirmax:
@mosesn I don't quite get it.

`Netty3Transporter` already does the interrupt handling [1]. But it neither can nor should expect the clients to handle the connection timeouts themselves.

I see two possible ways to deal with timeouts:

1. Somehow configure netty's handlers to close the channel when the timeout happens.
2. Do not use netty timeouts, instead add a timeout to one of the methods that return `Future[Transport[In, Out]]`.

We've started with the first one, and it does not seem right :) Now I think I'd prefer the second one. While it adds a small overhead, it results in a predictable uniform timeout behaviour, agnostic to the pipeline configuration. In fact since `ConnectTimeout` parameter is defined on the `Transporter` abstraction, the timeout handler may be written as a helper function that can be used in any `Transporter` implementation, be it a netty-based or not.

[1] https://github.com/twitter/finagle/blob/develop/finagle-core/src/main/scala/com/twitter/finagle/netty3/Netty3Transporter.scala#L50

mosesn:
Although the Netty3Transporter does cancel the ChannelFuture, it doesn't get propagated to the [handshake][0] properly.  My impression is that if we add cancelling the handshake to the onCancellation, it will be correct.

Could you elaborate on what you mean by "But it neither can nor should expect the clients to handle the connection timeouts themselves".  I'm not sure I understand.

[0]: https://github.com/twitter/finagle/blob/develop/finagle-core/src/main/scala/com/twitter/finagle/ssl/SslConnectHandler.scala#L86

sirmax:
@mosesn in your previous comment you've said _"We know the interrupt will come already"_. I interpreted this as an assumption, that there is always some other code outside of the `Netty3Transporter` that eventually interrupts the connection, thus we just need to handle interrupts properly.

But the problem is this is not always true. In #345 I mentioned `FailFastFactory` getting stuck indefinitely, waiting for the connection to succeed or fail, and not sending any interrupts.

And regarding "nor should expect the clients to handle the connection timeouts themselves", I mean it is the `Transporter`'s responsibility to handle connection timeouts one way or another.

Hope this makes some sense :)

mosesn:
I think we should say that's a bug in `FailFastFactory` then.  It seems reasonable to me that we interrupt the previous failfast connection attempt when we start a new one.  Otherwise, if we really end up hanging somewhere in here, we can leak memory.

It shouldn't have to be the `Transporter`'s job to handle connection timeouts.  I think trying as hard as we can until we receive an explicit interruption is a completely reasonable way to go about doing it.  If folks never set a connectTimeout, no need to kill it, and we can just try forever.

sirmax:
> interrupt the previous failfast connection attempt when we start a new one

But the thing is `FailFastFactory` does not start a new connection until the previous attempt succeeds or fails. So no memleaks here, just a completely broken service :)

And I'm not sure it is good idea to make something else responsible for _connection_ timeouts. …Perhaps we can make a `ServiceFactory` proxy that interrupts connection attempts just like the `TimeoutFilter` interrupts requests to the underlying `Service`, but would it be more robust and easier to use?

mosesn:
Hmm, yeah, you're right.  I thought that the retry was out of band, but clearly it's not.  Maybe we should move the retries out of band and also interrupt the old ones?

My point is that we already have what you described, in [TimeoutFactory][0].  It's just too high up for FailFastFactory to take advantage of.

[0]: https://github.com/twitter/finagle/blob/develop/finagle-core/src/main/scala/com/twitter/finagle/factory/TimeoutFactory.scala

sirmax:
Oh wow! I've completely overlooked the `TimeoutFactory`. Now everything you said about `FailFastFactory` being broken makes much more sense. I feel really dumb right now.

Could you, please, elaborate on _"move the retries out of band and also interrupt the old ones"_?

mosesn:
As far as I can tell, the current behavior of FailFastFactory (and I need to sit down and spend some time with it, I might be totally off) is that it shoots off a connection attempt, and once it fails, we start waiting a specified backoff time before shooting off another.

If instead of waiting for it to fail, we just start waiting the backoff time before sending another request, we can use the backoff time as both backoff and timeout.

luciferous:
@sirmax How's this coming along? Can we help with anything?

sirmax:
Hi @luciferous! Frankly speaking this is not coming along at all – I'm out of spare time :(

I think we'd better close this PR and just keep #345 open until someone fixes it. Here's why:

- In this PR I tried to fix #345 by adding an SSL-specific configuration parameter. Later we concluded it was not a good idea.
- @mosesn was against re-using `Transporter.ConnectTimeout` for SSL, and he's probably right. It would be more like a workaround hack than a proper fix.
- The proper fix should be somewhere around `FailFastFactory`. I can't yet pinpoint where exactly it should be.

luciferous:
Closing this PR and updating #345 with @sirmax's last comment.

PR #347, opened 2015-02-28
Redis newRichClient should take be a raw service.

tonyd3:
Problem
We want to create a customized Redis client using the client builder, but still use it with the RedisClient object.

Solution
Define a new constructor that takes an already created service.

Result
RedisClient.newRichClient(someService)

### DIFF
--- a/finagle-redis/src/main/scala/com/twitter/finagle/Redis.scala
+++ b/finagle-redis/src/main/scala/com/twitter/finagle/Redis.scala
@@ -17,6 +17,9 @@ trait RedisRichClient { self: Client[Command, Reply] =>
 
   def newRichClient(dest: Name, label: String): redis.Client =
     redis.Client(newService(dest, label))
+    
+  def newRichClient(raw: Service[Command, Reply]): redis.Client =
+    redis.Client(raw)
 }
 
 object RedisTransporter extends Netty3Transporter[Command, Reply]("redis", redis.RedisClientPipelineFactory)


mosesn:
@tonyd3 why not use the Client constructor directly?

https://github.com/twitter/finagle/blob/master/finagle-redis/src/main/scala/com/twitter/finagle/redis/Client.scala#L31

tonyd3:
The reason why was I wanted to use RedisClient was because of it extends DefaultClient, which has a custom endpointer. I can of course re-write that wrapper, but I thought It would be nice to have it in the finagle code.

mosesn:
Sorry, I don't understand what you're saying.  I don't think there's a difference between using the Client constructor directly and the method you're proposing.  Isn't the method you're proposing invoking the redis.Client constructor?

tonyd3:
For context, check this thread. https://groups.google.com/forum/#!topic/finaglers/7CZkQOemhp4

The last comment, Sanjay's is what I'm trying to solve. We use the old client builder api because we can customize it better. The problem is that the built client won't have the pipeline, as mentioned in Marius' comment. We want to integrate the built client with pipelining provided in RedisClient. Right now, internally we basically duplicate the code provided in RedisClient and RedisTransporter. Ideally we'd like to be able to just call RedisClient.newRichClient(someService), and get the same result.

mosesn:
@tonyd3 I don't think you really addressed my problem.

Let me be more precise.  The API you've added is exactly the same as using the constructor directly.  It doesn't seem to be easier to use, and it doesn't solve the problem you think it does.  Using the RedisClient.newService gives you the PipeliningDispatcher because it uses a different StackClient than the ClientBuilder.  What your method does is wrap a Service in a redis.Client, which we already have an API for.

tonyd3:
I think we're probably missing something really obvious. I was wondering if you can advise on how to convert the following code snippet to use the new finagle api. 

Java Code:
```java
  public static ClientBuilder<
      Command, Reply, ClientConfig.Yes, ClientConfig.Yes, ClientConfig.Yes> redisStubBuilder(
      String name, ServerSpec serverSpec) {

    final ClientBuilder<Command, Reply, ClientConfig.Yes, ClientConfig.Yes, ClientConfig.Yes>
        builder = ClientBuilder.get()
        .codec(Redis.get())
        .hostConnectionLimit(redisConnectionsPerHost.value())
        .dest(serverSpec.toFinagleString())
        .name(sanitizeName(name))
        .failFast(useFailFastByDefault.value())
        .retries(redisDefaultRetries.value())
        .tcpConnectTimeout(Duration.apply(5, TimeUnit.SECONDS))
        .timeout(finagleDurationFromAirliftDuration(redisTimeout.value()))
        .reportHostStats(defaultFinagleStatsReceiver)
        .reportTo(varzAndFinagleStatsReceiver)
        .logger(logger);

    return builder;
  }

  public static Client redisClient(String name, ServerSpec serverSpec){
    Service<Command, Reply> client = ClientBuilder.safeBuild(redisStubBuilder(name, serverSpec));
    return RedisClientWrapper.newFinagleRedisClient(client);
  }
```

our Scala Wrapper:
```scala
trait FinagleRedisClient { self: Client[Command, Reply] =>
  def newFinagleRedisClient(raw: Service[Command, Reply]): com.twitter.finagle.redis.Client =
    com.twitter.finagle.redis.Client(raw)
}

object RedisTransporter extends Netty3Transporter[Command, Reply](
  "redis", com.twitter.finagle.redis.RedisClientPipelineFactory)

object RedisClientWrapper extends DefaultClient[Command, Reply](
  name = "redis",
  endpointer = Bridge[Command, Reply, Command, Reply](RedisTransporter, new PipeliningDispatcher(_)),
  pool = (sr: StatsReceiver) => new SingletonPool(_, sr)
) with FinagleRedisClient
```

mosesn:
You can't use the ClientBuilder if you want the PipelinedDispatcher right now.  The part that you're missing is that the RedisClientWrapper's "DefaultClient" section is never used.  redis.Client is a global object, and you're just directly passing the ClientBuilder created instance to that global instance.  I think if you go and look at the piece of [code you're invoking][0] it will be more clear why it's not using any of the DefaultClient methods.

As an aside, DefaultClient is not really used anymore, although we still have a few protocols that are still on it.  StackClient is the way forward (for one thing it is much more easily configurable).

[0]: https://github.com/twitter/finagle/blob/master/finagle-redis/src/main/scala/com/twitter/finagle/redis/Client.scala#L31

Issue #309, opened 2014-09-19
LoadService gets stuck recursively descending into /sys


groestl:
When spinning up our server on a rackspace instance (uname "Linux 3.2.0-60-generic #91-Ubuntu SMP Wed Feb 19 03:54:44 UTC 2014 x86_64 x86_64") the startup procedure gets stuck as soon as the finage client tries to set up LoadedStatsReceiver.

In the stack trace we see LoadService descending into a directory tree:
```
"main" prio=10 tid=0x00007f30b4010000 nid=0x7ff runnable [0x00007f30bdc00000]
   java.lang.Thread.State: RUNNABLE
	at java.io.UnixFileSystem.getBooleanAttributes0(Native Method)
	at java.io.UnixFileSystem.getBooleanAttributes(UnixFileSystem.java:242)
	at java.io.File.isDirectory(File.java:843)
	at com.twitter.finagle.util.ClassPath$$anonfun$com$twitter$finagle$util$ClassPath$$browseDir$1.apply(LoadService.scala:87)
	at com.twitter.finagle.util.ClassPath$$anonfun$com$twitter$finagle$util$ClassPath$$browseDir$1.apply(LoadService.scala:86)
	at scala.collection.IndexedSeqOptimized$class.foreach(IndexedSeqOptimized.scala:33)
--- repeated 69 times
	at scala.collection.mutable.ArrayOps$ofRef.foreach(ArrayOps.scala:108)
	at com.twitter.finagle.util.ClassPath$.com$twitter$finagle$util$ClassPath$$browseDir(LoadService.scala:86)
	at com.twitter.finagle.util.ClassPath$$anonfun$com$twitter$finagle$util$ClassPath$$browseDir$1.apply(LoadService.scala:88)
	at com.twitter.finagle.util.ClassPath$$anonfun$com$twitter$finagle$util$ClassPath$$browseDir$1.apply(LoadService.scala:86)
	at scala.collection.IndexedSeqOptimized$class.foreach(IndexedSeqOptimized.scala:33)
--- end repeat
	at scala.collection.mutable.ArrayOps$ofRef.foreach(ArrayOps.scala:108)
	at com.twitter.finagle.util.ClassPath$.com$twitter$finagle$util$ClassPath$$browseDir(LoadService.scala:86)
	at com.twitter.finagle.util.ClassPath$.browseUri(LoadService.scala:75)
[...]

PR #351, opened 2015-03-16
Added ability to add to the ignored packages sequence for LoadService

LithiumTD:
Problem

The problem being fixed here is something we ran into while working with finagle in our application similar to the issue posted in the following issue https://github.com/twitter/finagle/issues/309. In our case it was not the /sys directory however but an extremely huge FS mounted on the project root and LoadService was getting stuck descending into that directory structure everytime.

Solution

This pull request adds the ability to add additional packages/paths to the 'ignoredPackages' sequence in LoadService via a method called 'addIgnoredPaths' in LoadService. This method can take a single or multiple paths as String arguments and add them to the private sequence (which is now mutable).

Result

After this change, users would be able to customize the ignoredPackages for the recursive directory scan that LoadService performs (in several flows) and avoid getting stuck in directory structures that they do not want scanned.

### DIFF
--- a/finagle-core/src/main/scala/com/twitter/finagle/util/LoadService.scala
+++ b/finagle-core/src/main/scala/com/twitter/finagle/util/LoadService.scala
@@ -29,16 +29,13 @@ import com.twitter.app.GlobalFlag
  */
 private object ClassPath {
 
-  private[util] var ignoredPackages = Seq(
+  private val defaultIgnoredPackages = Seq(
     "apple/", "ch/epfl/", "com/apple/", "com/oracle/",
     "com/sun/", "java/", "javax/", "scala/", "sun/", "sunw/")
 
-  /**
-   * Adding additional packages to be ignored.
-   */
-  def addIgnoredPackages(packages: String*) = {
-    ignoredPackages = ignoredPackages ++ packages
-  }
+  private[util] def ignoredPackages =
+    if (ignoredPaths.isDefined) defaultIgnoredPackages ++ ignoredPaths()
+    else defaultIgnoredPackages
 
   /**
    * Information about a classpath entry.
@@ -188,7 +185,8 @@ private object ClassPath {
   }
 }
 
-object ignoredPaths extends GlobalFlag("", "Additional packages to be excluded from recursive directory scan")
+object ignoredPaths extends GlobalFlag(Seq.empty[String],
+    "Additional packages to be excluded from recursive directory scan")
 
 /**
  * Load a singleton class in the manner of [[java.util.ServiceLoader]]. It is
--- a/finagle-core/src/main/scala/com/twitter/finagle/util/LoadService.scala
+++ b/finagle-core/src/main/scala/com/twitter/finagle/util/LoadService.scala
@@ -22,9 +22,16 @@ import scala.reflect.ClassTag
  */
 private object ClassPath {
 
-  private val ignoredPackages = Seq(
+  private var ignoredPackages = Seq(
     "apple/", "ch/epfl/", "com/apple/", "com/oracle/",
     "com/sun/", "java/", "javax/", "scala/", "sun/", "sunw/")
+
+  /**
+   * Adding additional packages to be ignored.
+   */
+  def addIgnoredPackages(packages: String*) = {
+    ignoredPackages = ignoredPackages ++ packages
+  }
 
   /**
    * Information about a classpath entry.
@@ -184,6 +191,11 @@ object LoadService {
 
   def apply[T: ClassTag](): Seq[T] = synchronized {
     val iface = implicitly[ClassTag[T]].runtimeClass.asInstanceOf[Class[T]]
+
+  def addIgnoredPaths(paths: String*) = {
+    ClassPath.addIgnoredPackages(paths: _*)
+  }
+
     val ifaceName = iface.getName
     val loader = iface.getClassLoader
 
--- a/finagle-core/src/main/scala/com/twitter/finagle/util/LoadService.scala
+++ b/finagle-core/src/main/scala/com/twitter/finagle/util/LoadService.scala
@@ -27,9 +27,7 @@ private object ClassPath {
     "apple/", "ch/epfl/", "com/apple/", "com/oracle/",
     "com/sun/", "java/", "javax/", "scala/", "sun/", "sunw/")
 
-  private[util] def ignoredPackages =
-    if (ignoredPaths.isDefined) defaultIgnoredPackages ++ ignoredPaths()
-    else defaultIgnoredPackages
+  private[util] def ignoredPackages = defaultIgnoredPackages ++ ignoredPaths()
 
   /**
    * Information about a classpath entry.
--- a/finagle-core/src/main/scala/com/twitter/finagle/util/LoadService.scala
+++ b/finagle-core/src/main/scala/com/twitter/finagle/util/LoadService.scala
@@ -1,5 +1,6 @@
 package com.twitter.finagle.util
 
+import com.twitter.app.GlobalFlag
 import com.twitter.logging.Level
 import com.twitter.util.NonFatal
 import com.twitter.util.registry.GlobalRegistry
@@ -7,18 +8,11 @@ import java.io.{File, IOException}
 import java.net.{URI, URISyntaxException, URLClassLoader}
 import java.nio.charset.MalformedInputException
 import java.util.ServiceConfigurationError
-<<<<<<< HEAD
 import java.util.jar.JarFile
 import scala.collection.JavaConverters._
 import scala.collection.mutable
 import scala.io.Source
 import scala.reflect.ClassTag
-=======
-import scala.collection.JavaConverters._
-import scala.collection.mutable
-import scala.io.Source
-import com.twitter.app.GlobalFlag
->>>>>>> added ignoredPaths GlobalFlag and unit test
 
 /**
  * Inspect and load the classpath. Inspired by Guava's ClassPath
@@ -198,17 +192,6 @@ object LoadService {
 
   def apply[T: ClassTag](): Seq[T] = synchronized {
     val iface = implicitly[ClassTag[T]].runtimeClass.asInstanceOf[Class[T]]
-
-  def addIgnoredPaths(paths: String*) = {
-    ClassPath.addIgnoredPackages(paths: _*)
-
-  private[util] def updateIgnoredPackages = {
-    Option(ignoredPaths()) filter (_.trim.nonEmpty) foreach { ips =>
-      val packages = ips.split(',') map(_.trim)
-      ClassPath.addIgnoredPackages(packages: _*)
-    }
-  }
-  updateIgnoredPackages
 
     val ifaceName = iface.getName
     val loader = iface.getClassLoader
--- a/finagle-core/src/main/scala/com/twitter/finagle/util/LoadService.scala
+++ b/finagle-core/src/main/scala/com/twitter/finagle/util/LoadService.scala
@@ -27,7 +27,7 @@ private object ClassPath {
     "apple/", "ch/epfl/", "com/apple/", "com/oracle/",
     "com/sun/", "java/", "javax/", "scala/", "sun/", "sunw/")
 
-  private[util] def ignoredPackages = defaultIgnoredPackages ++ ignoredPaths()
+  private[util] def ignoredPackages = defaultIgnoredPackages ++ loadServiceIgnoredPaths()
 
   /**
    * Information about a classpath entry.
@@ -177,7 +177,7 @@ private object ClassPath {
   }
 }
 
-object ignoredPaths extends GlobalFlag(Seq.empty[String],
+object loadServiceIgnoredPaths extends GlobalFlag(Seq.empty[String],
     "Additional packages to be excluded from recursive directory scan")
 
 /**
--- a/finagle-core/src/main/scala/com/twitter/finagle/util/LoadService.scala
+++ b/finagle-core/src/main/scala/com/twitter/finagle/util/LoadService.scala
@@ -7,11 +7,18 @@ import java.io.{File, IOException}
 import java.net.{URI, URISyntaxException, URLClassLoader}
 import java.nio.charset.MalformedInputException
 import java.util.ServiceConfigurationError
+<<<<<<< HEAD
 import java.util.jar.JarFile
 import scala.collection.JavaConverters._
 import scala.collection.mutable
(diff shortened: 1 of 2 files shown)


mosesn:
This seems reasonable, but maybe the right way to do this is with a `GlobalFlag`?

LithiumTD:
Alright, i will modify the code to use the "GlobalFlag" instead of the extra method in the LoadService companion object. Will update the pull request asap :)

stevegury:
Sounds reasonable as well to me, please add a simple test so that we won't regress this feature (that we won't use at Twitter).

LithiumTD:
Test added and GlobalFlag used for updating the ignoredPackages. Hope this works for you guys :)

finagle-core/src/main/scala/com/twitter/finagle/util/LoadService.scala
```diff
@@ -156,6 +163,8 @@ private object ClassPath {
   }
 }
 
+object ignoredPaths extends GlobalFlag("", "Additional packages to be excluded from recursive directory scan")
```
mosesn: Instead of a String-flavored Flag, this should be a Seq[String] flavored flag.  That way you don't need to read the code to guess how to parse it.

finagle-core/src/main/scala/com/twitter/finagle/util/LoadService.scala
```diff
@@ -21,11 +21,18 @@ import scala.io.Source
  */
 private object ClassPath {
 
-  private val ignoredPackages = Seq(
+  private[util] var ignoredPackages = Seq(
```
mosesn: instead of making this mutable and adding a method in LoadService to update it, let's just add ignoredPaths directly when we declare ignoredPackages.

finagle-core/src/main/scala/com/twitter/finagle/util/LoadService.scala
```diff
@@ -21,11 +21,18 @@ import scala.io.Source
...
     "apple/", "ch/epfl/", "com/apple/", "com/oracle/",
     "com/sun/", "java/", "javax/", "scala/", "sun/", "sunw/")
 
   /**
+   * Adding additional packages to be ignored.
+   */
+  def addIgnoredPackages(packages: String*) = {
```
mosesn: seems like you wanted to expose this for testing, but we could instead make ignoredPackages a method, or else add a private[util] method version of ignoredPackages.

finagle-core/src/main/scala/com/twitter/finagle/util/LoadService.scala
```diff
@@ -21,10 +21,14 @@ import scala.io.Source
  */
 private object ClassPath {
 
-  private val ignoredPackages = Seq(
+  private val defaultIgnoredPackages = Seq(
     "apple/", "ch/epfl/", "com/apple/", "com/oracle/",
     "com/sun/", "java/", "javax/", "scala/", "sun/", "sunw/")
 
+  private[util] def ignoredPackages =
```
LithiumTD: This method was done as suggested along with adding of the val for the default ignored packages. We are not sure how much of a performance hit it would be to calculate this as a def on each call but this would definitely ensure no regressions as well as avoid any explicit method calls etc to update them.

finagle-core/src/main/scala/com/twitter/finagle/util/LoadService.scala
```diff
@@ -7,10 +7,10 @@ import java.net.{URI, URISyntaxException, URLClassLoader}
 import java.nio.charset.MalformedInputException
 import java.util.jar.JarFile
 import java.util.ServiceConfigurationError
-
 import scala.collection.JavaConverters._
 import scala.collection.mutable
 import scala.io.Source
+import com.twitter.app.GlobalFlag
```
mosesn: please keep imports sorted

finagle-core/src/main/scala/com/twitter/finagle/util/LoadService.scala
```diff
@@ -21,10 +21,14 @@ import scala.io.Source
  */
 private object ClassPath {
 
-  private val ignoredPackages = Seq(
+  private val defaultIgnoredPackages = Seq(
     "apple/", "ch/epfl/", "com/apple/", "com/oracle/",
     "com/sun/", "java/", "javax/", "scala/", "sun/", "sunw/")
 
+  private[util] def ignoredPackages =
+    if (ignoredPaths.isDefined) defaultIgnoredPackages ++ ignoredPaths()
```
mosesn: the if statement isn't necessary.  we can just add ignoredPaths without checking.

finagle-core/src/test/scala/com/twitter/finagle/util/LoadServiceTest.scala
```diff
@@ -13,6 +13,7 @@ import org.scalatest.junit.JUnitRunner
 import org.scalatest.mock.MockitoSugar
 import scala.collection.mutable
 import scala.util.Random
+import com.twitter.app.GlobalFlag
```
mosesn: keep imports sorted

finagle-core/src/main/scala/com/twitter/finagle/util/LoadService.scala
mosesn: Yeah, my guess is that it's very inexpensive.  If it starts being a bottleneck, we can work on it.

mosesn:
LGTM except for a few final nits.

finagle-core/src/test/scala/com/twitter/finagle/util/LoadServiceTest.scala
```diff
@@ -94,6 +95,12 @@ class LoadServiceTest extends FunSuite with MockitoSugar {
     executor.shutdown()
   }
 
+  test("LoadService should ignore packages according to ignoredPaths GlobalFlag") {
```
stevegury: I would like to see a more advanced test.
e.g. ignore "com.twitter.finagle.util" and try to load all the instances of `LoadServiceRandomInterface`, the result Seq should be empty

LithiumTD: I tried to run a test with the requirements you specified. For some reason, even though the addition of the package "com.twitter.finagle.util" to the list of ignoredPackages used in the browseUri it still seemed to pick up the LoadServiceRandomInterfaceImpl with the LoadService. Perhaps i am missing something here :)

mosesn: Could you paste the test you wrote?

LithiumTD: test("LoadService should ignore packages according to ignoredPaths GlobalFlag") {
  ignoredPaths.let(Seq("com.twitter.finagle.util")){
      assert(LoadService[LoadServiceRandomInterface]().isEmpty)
   }
 }

I used a similar LoadService call as done for the Announcer above so i assumed this should ignore things within the package. However it did return an object. I even tried it with directory notation with "com/twitter/finagle/util/" with the same result.

stevegury: I think I see why it doesn't work.
We are caching the classes meta-data per classloader, and LoadService has already scanned the classpath in a previous test.
Maybe adding a method to flush the cache (private to the package) or setting a new Classloader would solve the problem.

LithiumTD: I suspected it could have been something of that sort so i had commented out all previous tests and ran only the one i wrote but with the same results. Regardless, i would need to look into this and see how i can work around it (probably trying to use the methods you mentioned).

LithiumTD: Alright, so after further investigation as to why it was not picking up classes as expected based on the ignoredPackages, i tried the new Classloader as Steve suggested but that also yielded the same result.

I, however, did stumble across some interesting information. Several classes like Resolver, Announcer, LoadServiceMaybeInterface and LoadServiceRandomInterface were actually being picked up from "META-INF/services/" and then the getName for those files contained their fully qualified names such as "com.twitter.finagle.util.LoadServiceRandomInterface". Therefore, the only way i was able to exclude it from the LoadService scan was to add "META-INF/services/" to the ignoredPaths via the GlobalFlag.

I am not sure if this would be very readable test unless i add some sort of explanation there but if you have any idea as to how i could have a more straightforward test, it would be most appreciated.

On a not so related note, i was wondering if you could tell me how i could use the GlobalFlag programmatically in my application. Setting it via command-line args isn't really possible for us since we do not really use this as a complete Twitter App. We directly start up the Thrift/ThriftMux clients and this LoadService is simply a by-product for the scan for Resolvers when we try to do a bind. It would be nice to know if there is some way to set GlobalFlags within the code (besides the let() method).

georgepucea:
Hello, 
Will this pull request get merged? We are facing the same issue. 
Thanks!

LithiumTD:
Hi,
Thanks for bumping this. There is already a new finagle release and having to custom patch each release is becoming an overhead. We still require this feature and we are simply working around at the moment.
Thanks :)

luciferous:
@LithiumTD Thanks for all the work so far. Is this ready to merge yet, @mosesn @stevegury? It looks like there are still some issues outstanding, but it also seems like we're nearly at the finish line.

Anything else other than the following?
1. sort imports
2. don't need to check for `ignoredPaths.isDefined`
3. @LithiumTD's questions about usage of `GlobalFlag`s and the specifications for a test

finagle-core/src/test/scala/com/twitter/finagle/util/LoadServiceTest.scala
mosesn: You should be able to set it as a java flag (ie -Dcom.myproject.flag=3).

mosesn: I think it's fine to just add an explanation.

mosesn:
I think it's not quite ready, but very close!

LithiumTD:
Hi,
Thanks a lot for your response. Besides the questions that i had on how i would actually use the GlobalFlag (which is quite important since it would be pretty lame if we can't use something that we set out to do from the start) and the advanced test that Steve wanted to see for this feature i do not see any problems.

I will organize the imports and submit the request again but i would be grateful if someone could help me out on how i can programmatically set the GlobalFlag in my application.

Cheers !

luciferous:
@LithiumTD you can set the GlobalFlag with `ignoredPackages.let`. I might be misunderstanding your question because you seem to have used it correctly a few comments above.

LithiumTD:
Yeah i know i used let :) But the problem is that let only stays for statements within the let block. I tested called to some outside code for example placing my application's bind call inside a let. But the let only holds true as long as i have the check inside the class where i used my let. But as soon as it went inside one of the twitter resolvers, the ignoredpackages wasn't what i set it to be anymore. Maybe you could shed some light on this or maybe i am just using it in the wrong way.

luciferous:
Yes, that's the correct way to use it. The flag is bound to the value only within the let block. Can you describe how this doesn't work for your use case?

LithiumTD:
Alright, so a bit of detail about the problem that we face.
- We use the thriftmux clients from finagle and use the zkResolver to do our binds.
- As you know the zkResolver (and others as well) internally call LoadService which then scans our directory structure for the application for possible custom resolvers.
- We have a really huge FS mounted on one of the folders in our application root where the LoadService gets stuck every time that we tried to do the bind (which didn't cause any problems on local system tests).
- So now we would like to add that directory as an ignoredPackage (which we currently have by using a custom finagle library with that folder hard-coded in the class)

Problem is that when i try to use this GlobalFlag solution and place the bind call within a let block, as soon as the control moves out of the class which contains the let block, the ignoredPackages does not reflect what i set in the let block. So i need some means to actually set the GlobalFlag 'globally' for my application. Since we don't use a full fledged TwitterApp, the CL solution would not work out for us. It would be great if you could suggest some alternative to this or correct me if i am using the let in some incorrect manner :)

Cheers !

luciferous:
I think I'm missing why you can't use regular JVM flags, you don't need a "full fledged TwitterApp" to use them.

LithiumTD:
The reason i wanted to avoid using Java system properties or command line flags was because we use this component as a part of a bigger application where this is just included as a jar file. If i did want to add some argument, i would have to get it added on the java start scripts for that application (which involves convincing sysadmins to change something which is never easy :) ). But, i will try it anyway if there isnt another way to get this set programmatically application wide.

Thanks for your help. Will get back to you with an updated patch soon :)

LithiumTD:
Alright guys, I am done with the organizing of imports and rebasing of the branch. Please let me know if there are any further changes required. Also, i managed to place some java system properties in our server code somewhere and it worked like a charm so the fix works out fine for us :)

Cheers !

finagle-core/src/main/scala/com/twitter/finagle/util/LoadService.scala
mosesn: this is still the case, but I won't block this if you feel really strongly about it.

mosesn:
The `if (ignoredPaths.isDefined)` thing is still unnecessary, but if you feel strongly about it, I don't want to block this PR.  :+1: otherwise.

LithiumTD:
Hmm, i should have seen that comment of yours :)
Let me just change it real quick then. You did mention it should not have a huge impact on performance.

finagle-core/src/main/scala/com/twitter/finagle/util/LoadService.scala
```diff
@@ -174,6 +177,9 @@ private object ClassPath {
   }
 }
 
+object ignoredPaths extends GlobalFlag(Seq.empty[String],
```
kevinoliver: do you mind making this flag's name more specific? 

right now its "com.twitter.finagle.util.ignoredPaths" which is a bit opaque for my tastes. perhaps "com.twitter.finagle.util.loadServiceIgnoredPaths" ?

kevinoliver:
@LithiumTD just one minor naming fix and then i'll merge this in.

LithiumTD:
Yeah makes sense :)
Will get this done right away.

kevinoliver:
Great, thanks. I'll try to get this merged in today or tomorrow.

kevinoliver:
Ok this is merged in but just missed this week's push to the develop branch. Thanks again for the contribution.

PR #352, opened 2015-03-22
Made RetryingFilter extendable.

spockz:
Problem

The main logic of the RetryingFilter is good enough for me, however it does not
pass the request to the policy. In my case I always want to retry GET requests.
Therefore, I want to extend the RetryingFilter such that the new apply() method
can use the request to create a new policy. In the current situation that is not
possible because the dispatch method is private. Consequently, the
RetryingFilter is not extendable.

Solution

Changed the access modifier from private to protected.

Result

Classes that extend the RetryingFilter can now call dispatch.

### DIFF
--- a/finagle-core/src/main/scala/com/twitter/finagle/service/RetryingFilter.scala
+++ b/finagle-core/src/main/scala/com/twitter/finagle/service/RetryingFilter.scala
@@ -55,7 +55,7 @@ class RetryingFilter[Req, Rep](
     } else f
   }
 
-  private[this] def dispatch(
+  protected def dispatch(
     req: Req,
     service: Service[Req, Rep],
     policy: RetryPolicy[Try[Nothing]],


travisbrown:
For reference, this change is being [discussed on Finaglers](https://groups.google.com/d/msg/finaglers/Z7wgOhNmF2Q/ec4CBVSL5ikJ). Thanks @spockz!

travisbrown:
(Note that in general it's best to submit pull requests against the `develop` branch, but in this case it's not a big deal.)

spockz:
@travisbrown: The contributing document states to create a pr against master. (https://github.com/twitter/finagle/blob/master/CONTRIBUTING.md)

travisbrown:
@spockz Ah, sorry!—thought that had been updated, but now that you mention it it may have just been discussed in [this announcement](https://finagle.github.io/blog/2014/10/07/finagle-releases/). We'll get it fixed asap.

stevegury:
Not a big deal about master/develop branch, nobody touched the RetryingFilter class!
@spockz Your change LGTM, I'm pulling it internally.

dschobel:
merged in https://github.com/twitter/finagle/commit/24d6e3e307c814410278ceca20c6ff5f7caeb161 thanks @spockz

PR #353, opened 2015-03-22
Changed localhost to 127.0.0.1 to prevent ipv6 from being added to the pool

spockz:
Problem

When running the tests on my machine, which has both an ipv4 and 6 interface,
the tests fail because the endpoints are added to the load-balancer twice and
therefore the tests fail with the message that 2 != 1.

Solution

I've changed the hostname from localhost to 127.0.0.1 in the afflicted test. If
this is really the solution than there are probably more tests where tests where
this should be fixed. Additionally I added a better error message to the
assertion.

Result

Tests will run on my machine.

### DIFF
--- a/finagle-core/src/test/scala/com/twitter/finagle/client/StackClientTest.scala
+++ b/finagle-core/src/test/scala/com/twitter/finagle/client/StackClientTest.scala
@@ -27,19 +27,20 @@ class StackClientTest extends FunSuite
 
   test("client stats are scoped to label")(new Ctx {
     // use dest when no label is set
-    client.newService("inet!localhost:8080")
+    client.newService("inet!127.0.0.1:8080")
     eventually {
-      assert(sr.counters(Seq("inet!localhost:8080", "loadbalancer", "adds")) === 1)
+      val counter = sr.counters(Seq("inet!127.0.0.1:8080", "loadbalancer", "adds"))
+      assert(counter == 1, s"The instance should be to the loadbalancer once instead of $counter times.")
     }
 
     // use param.Label when set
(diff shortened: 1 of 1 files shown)


mosesn:
I'm not sure what we should do here.  The problem is that finagle has a few tests where it depends on the JVM setting java.net.preferIPv4Stack=true.  It might be that it's better to make sure that that's set than to change localhost to 127.0.0.1, at least until we have time to upgrade all of the tests to support ipv6 (in particular, memcached has trouble with ipv6).

stevegury:
@mosesn I think it's better to be explicit in the code, rather than relying on the JVM settings.
(I think we still have time before switching to ipv6.)

dschobel:
@spockz were you using finagle's [sbt wrapper](https://github.com/twitter/finagle/blob/master/sbt) to run your tests?

I think hard-coding 127.0.0.1 is fine because it is easily greppable but I'm curious how users are running into cases where the property isn't defined.

spockz:
@dschobel, no I was just running it with my local sbt. I that the wrapper uses the prefer IPv4 flag which also should fix the issue.

dschobel:
LGTM, I think this should still come in if it makes people's lives easier since the cost to maintain is a grep run.

luciferous:
@dschobel @mosesn Where are with this?

dschobel:
@luciferous still lgtm

mosesn:
:ship: :it:

vkostyukov:
This has been merged internally. Thanks @spockz! You contribution will show up here on the next sync cycle.

PR #354, opened 2015-03-24
Bugfix memcached Zipkin integration.

kristofa:
Problem

The memcached Zipkin integration was broken. When getting a value from memcached
the cache hit/miss annotation was submitted after the cr annotation was submitted.
This had as a result that a new span was created with same id which was eventually
submitted by the DeadlineSpanMap once the deadline was exceeded. The span submitted
by the DeadlineSpanMap overrides the duration of the earlier submitted span and messes
it up. It also makes the zipkin-web trace overview almost unusable.

Solution

Instead of submitting the cache hit/miss annotations using the TraceInitializerFilter.role,
submit them using ClientTracingFilter.role before the cr annotation is submitted.
This means the memcached specific tracing filter submits following annotations:
cs -> memcache specific annotations -> cr. To be able to do this I updated the Codec trait
to be able to override filters which have a custom role and not the TraceInitializerFilter.role.

Result

The memcached span does not end up in the DeadlineSpanMap and the spans are correctly
submitted now including cache hit/miss annotations.  The duration is correct and the
trace overview in zipkin-web looks good again.

### DIFF
--- a/finagle-core/src/main/scala/com/twitter/finagle/Codec.scala
+++ b/finagle-core/src/main/scala/com/twitter/finagle/Codec.scala
@@ -73,7 +73,15 @@ trait Codec[Req, Rep] {
    * A hack to allow for overriding the TraceInitializerFilter when using
    * Client/Server Builders rather than stacks.
    */
+  @deprecated(message="Replaced by stackRolesToReplace which is more generic.", since = "6.26.0")
   def newTraceInitializer: Stackable[ServiceFactory[Req, Rep]] = TraceInitializerFilter.clientModule[Req, Rep]
+
+  /**
+   * Supports overriding Stack roles when using Client/Server Builders
+   * rather than stacks.
+   */
+  def stackRolesToReplace: Map[Stack.Role, Stackable[ServiceFactory[Req, Rep]]] = Map()
+
 }
 
 /**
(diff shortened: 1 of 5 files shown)


kristofa:
Because the memcached client still is implemented using ClientBuilder and is not a StdStackClient this change involved supporting overriding `ClientTracingFilter.role` instead of `TraceInitializerFilter.role` which was already supported.  Also overriding `TraceInitializerFilter.role` was supported both in `Codec` and in `DefaultClient` constructor.  I decided to update Codec to allow overriding any 'role' and deprecating the old method for overriding `TraceInitializerFilter.role`.

If you guys are fine with this change I can add an additional commit to also fix the memcachedx client.

kristofa:
I don't know why Github keeps indicating the build is still in progress. When looking at Travis you see the builds are completed however only 1 of the 3 jobs finished successful.  The 2 failed jobs seem to have failed for reasons not related to my change.  The Oracle jdk 7 one succeeded. Any input?

luciferous:
Hi @kristofa we're working on something right now to fix the Travis builds.

kristofa:
@luciferous Thanks for letting me know!

kristofa:
Since `finagle-memcached` is deprecated in favour of `finagle-memcachedx` as of release 6.26 I'll close this and have a look at `finagle-memcachedx`.

PR #356, opened 2015-03-31
Update documentation about contributing

travisbrown:
Problem

The contributing documentation currently recommends submitting pull requests
against master.

Solution

Update the contributing documentation to reflect the current workflow, and add
a note about branches to the main README.

PR #358, opened 2015-04-03
Add support for Redis HyperLogLog commands

charithe:
Problem:

Finagle-redis currently lacks support for HyperLogLog commands
(http://redis.io/commands#hyperloglog).

Solution:

Add code to enable calling PFADD, PFCOUNT and PFMERGE from RedisClient

Result:

RedisClient API will have three new methods:
- pfAdd
- pfCount
- pfMerge

### DIFF
--- a/finagle-redis/src/main/scala/com/twitter/finagle/redis/Client.scala
+++ b/finagle-redis/src/main/scala/com/twitter/finagle/redis/Client.scala
@@ -37,6 +37,7 @@ class Client(service: Service[Command, Reply])
   with Lists
   with Sets
   with BtreeSortedSetCommands
+  with HyperLogLogs
 
 /**
  * Connects to a single Redis host
--- a/finagle-redis/src/main/scala/com/twitter/finagle/redis/HyperLogLogCommands.scala
+++ b/finagle-redis/src/main/scala/com/twitter/finagle/redis/HyperLogLogCommands.scala
@@ -1,0 +1,48 @@
+package com.twitter.finagle.redis
+
+import _root_.java.lang.{Long => JLong,Boolean => JBoolean}
+import com.twitter.finagle.redis.protocol.{StatusReply, IntegerReply}
+import com.twitter.finagle.redis.protocol.commands.{PFMerge, PFCount, PFAdd}
+import com.twitter.util.Future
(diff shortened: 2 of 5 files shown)


charithe:
Travis build fails on `finagle-common-stats` and `finagle-stream`. `finagle-redis` builds successfully.

```
The command "./sbt ++$TRAVIS_SCALA_VERSION finagle-redis/test" exited with 0
```

travisbrown:
@charithe This is really great. There have been some changes to our workflow (see my updates to the documentation in #356) and to finagle-redis specifically (#331) that mean that this will need to be rebased against the `develop` branch and have its tests migrated from Specs to ScalaTest. I've taken the liberty of doing both of those things in my fork [here](https://github.com/travisbrown/finagle/commits/hyperloglog_support), and if you want to grab those two commits and submit a new PR against `develop`, we should be able to move this forward more quickly.

Two other small points: I've also moved the commands out of the `c.t.f.r.protocol.commands` package directly into `c.t.f.r.protocol`. This is kind of confusing since the package structure doesn't match the directory structure, but Scala doesn't require that, and this is the convention the other commands follow, so it makes sense to be consistent.

I've also only added stubs for the integration tests with basic tests for `PFADD`. Ideally these would be more fleshed out.

charithe:
Thanks. I wasn't aware of the change to the contribution guide. I will resubmit the PR with the suggested changes.

charithe:
@travisbrown I pulled your branch and added ScalaTest tests for all commands. However, now I can't compile it locally due to a missing dependency: `com.twitter#util-core_2.10;6.23.0-SNAPSHOT`. Any suggestions?

New diff: https://github.com/twitter/finagle/compare/develop...charithe:redis_hyperloglog

travisbrown:
There's no reason you should have been aware of the PR with the doc changes—it's our fault that's it's not up already. About the new error: Finagle depends on Twitter Util, and at the moment you'll need to publish it locally by cloning [the repo](https://github.com/twitter/util), checking out the `develop` branch, and running `./sbt +publishLocal`. Over the past few months the way that we publish and consume Finagle and Util internally has changed, and we're still working out the best way to handle publishing here, but this inconvenience should be unnecessary very soon.

PR #359, opened 2015-04-05
Redis hyperloglog support

charithe:
Problem:

- Finagle-redis currently lacks support for HyperLogLog commands
(http://redis.io/commands#hyperloglog).
- First pull request #358 did not adhere to contribution guidelines

Solution:

- Add code to enable calling PFADD, PFCOUNT and PFMERGE from RedisClient
- Pull suggested refactoring from https://github.com/travisbrown/finagle/commit/556761c94d0d37bc6b8c3369ccc66ee5d7a540a8 and add ScalaTest integration tests

Result:

RedisClient API will have three new methods:

- pfAdd
- pfCount
- pfMerge

### DIFF
--- a/finagle-redis/src/main/scala/com/twitter/finagle/redis/HyperLogLogCommands.scala
+++ b/finagle-redis/src/main/scala/com/twitter/finagle/redis/HyperLogLogCommands.scala
@@ -1,8 +1,7 @@
 package com.twitter.finagle.redis
 
 import _root_.java.lang.{Long => JLong,Boolean => JBoolean}
-import com.twitter.finagle.redis.protocol.{StatusReply, IntegerReply}
-import com.twitter.finagle.redis.protocol.commands.{PFMerge, PFCount, PFAdd}
+import com.twitter.finagle.redis.protocol.{IntegerReply, PFMerge, PFCount, PFAdd, StatusReply}
 import com.twitter.util.Future
 import org.jboss.netty.buffer.ChannelBuffer
 
--- a/finagle-redis/src/main/scala/com/twitter/finagle/redis/HyperLogLogCommands.scala
+++ b/finagle-redis/src/main/scala/com/twitter/finagle/redis/HyperLogLogCommands.scala
@@ -1,0 +1,48 @@
+package com.twitter.finagle.redis
+
+import _root_.java.lang.{Long => JLong,Boolean => JBoolean}
+import com.twitter.finagle.redis.protocol.{StatusReply, IntegerReply}
+import com.twitter.finagle.redis.protocol.commands.{PFMerge, PFCount, PFAdd}
+import com.twitter.util.Future
+import org.jboss.netty.buffer.ChannelBuffer
+
+trait HyperLogLogs { self: BaseClient =>
+
+  /**
+   * Adds elements to a HyperLogLog data structure.
+   *
+   * @param key
+   * @param elements
+   * @return True if a bit was set in the HyperLogLog data structure
+   * @see http://redis.io/commands/pfadd
+   */
+  def pfAdd(key: ChannelBuffer, elements: List[ChannelBuffer]): Future[JBoolean] =
+    doRequest(PFAdd(key, elements)) {
+      case IntegerReply(n) => Future.value(n == 1)
+    }
+
+
+  /**
+   * Get the approximated cardinality of sets observed by the HyperLogLog at key(s)
+   * @param keys
+   * @return Approximated number of unique elements
+   * @see http://redis.io/commands/pfcount
+   */
+  def pfCount(keys: Seq[ChannelBuffer]): Future[JLong] =
+    doRequest(PFCount(keys)) {
+      case IntegerReply(n) => Future.value(n)
+    }
+
+
+  /**
+   * Merge HyperLogLogs at srcKeys to create a new HyperLogLog at destKey
+   * @param destKey
+   * @param srcKeys
+   * @see http://redis.io/commands/pfmerge
+   */
+  def pfMerge(destKey: ChannelBuffer, srcKeys: Seq[ChannelBuffer]): Future[Unit] =
+    doRequest(PFMerge(destKey, srcKeys)) {
+      case StatusReply(_) => Future.Unit
+    }
+
+}
--- a/finagle-redis/src/main/scala/com/twitter/finagle/redis/protocol/commands/HyperLogLog.scala
+++ b/finagle-redis/src/main/scala/com/twitter/finagle/redis/protocol/commands/HyperLogLog.scala
@@ -1,4 +1,4 @@
-package com.twitter.finagle.redis.protocol.commands
+package com.twitter.finagle.redis.protocol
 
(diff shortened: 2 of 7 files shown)


charithe:
Travis build fails because `com.twitter#util-core_2.10;6.23.0-SNAPSHOT` cannot be resolved. All finagle-redis tests pass on my dev machine.

charithe:
Thanks @travisbrown. I fixed all the issues you pointed out.

We do use Finagle in production -- quite heavily I might add. When is your next release planned for? This particular piece with HyperLogLog is something I am currently prototyping so if it's going to delay the current release, I can wait until the next release or work off of snapshots.

luciferous:
The release is imminent, but at least a week away.

finagle-redis/src/test/scala/com/twitter/finagle/redis/commands/hyperloglog/HyperLogLogClientIntegrationSuite.scala
```diff
@@ -0,0 +1,35 @@
...
+
+import com.twitter.finagle.redis.naggati.RedisClientTest
+import com.twitter.finagle.redis.tags.{ClientTest, RedisTest}
+import com.twitter.util.{Future, Await}
+
+import org.junit.Ignore
+import org.junit.runner.RunWith
+import org.scalatest.junit.JUnitRunner
+
+@Ignore
```
luciferous: is this intentional?

finagle-redis/src/test/scala/com/twitter/finagle/redis/commands/hyperloglog/HyperLogLogClientServerIntegrationSuite.scala
```diff
@@ -0,0 +1,40 @@
...
+import com.twitter.finagle.redis.tags.{ClientServerTest, RedisTest}
+import com.twitter.finagle.redis.util.StringToChannelBuffer
+import com.twitter.util.{Await, Future}
+import org.jboss.netty.buffer.ChannelBuffer
+
+import org.junit.Ignore
+import org.junit.runner.RunWith
+import org.scalatest.junit.JUnitRunner
+
+@Ignore
```
luciferous: ignore?

finagle-redis/src/main/scala/com/twitter/finagle/redis/protocol/commands/HyperLogLog.scala
```diff
@@ -0,0 +1,49 @@
...
+  def toChannelBuffer = RedisCodec.toUnifiedFormat(Seq(CommandBytes.PFMERGE, destKey) ++ srcKeys)
+}
+
+object PFMerge {
+  def apply(args: Seq[Array[Byte]]): PFMerge = args match {
+    case head :: tail =>
+      PFMerge(ChannelBuffers.wrappedBuffer(head), tail map ChannelBuffers.wrappedBuffer)
+
+    case _ =>
+      throw ClientError("Invalid use of PFMerge")
```
luciferous: scaladoc @throws?

(side: is this how we do things in finagle-redis, otherwise I'd prefer Try/Option)

finagle-redis/src/main/scala/com/twitter/finagle/redis/protocol/commands/HyperLogLog.scala
```diff
@@ -0,0 +1,49 @@
...
+  override def toChannelBuffer = RedisCodec.toUnifiedFormat(Seq(CommandBytes.PFADD, key) ++ elements)
+}
+
+object PFAdd {
+  def apply(args: Seq[Array[Byte]]): PFAdd = args match {
+    case head :: tail =>
+      PFAdd(ChannelBuffers.wrappedBuffer(head), tail map ChannelBuffers.wrappedBuffer)
+
+    case _ =>
+      throw ClientError("Invalid use of PFAdd")
```
luciferous: scaladoc @throws?

(side: is this how we do things in finagle-redis, otherwise I'd prefer Try/Option)

finagle-redis/src/test/scala/com/twitter/finagle/redis/commands/hyperloglog/HyperLogLogClientIntegrationSuite.scala
travisbrown: Yes, we're not currently running the finagle-redis integration tests in Pants. This `@Ignore` has no effect on the SBT build, but these tests are also excluded by default in `Build.scala`.

It would be nice to have them run as proper integration tests in SBT. I've just created [an issue](https://github.com/twitter/finagle/issues/360) to track this.

finagle-redis/src/main/scala/com/twitter/finagle/redis/protocol/commands/HyperLogLog.scala
travisbrown: Yes, this is consistent with how other finagle-redis commands work. It'd be nice to clean up the API, but in my view that can wait until we spin this off into its own project in the Finagle org (and we don't have a definite timeline for that).

finagle-redis/src/main/scala/com/twitter/finagle/redis/HyperLogLogCommands.scala
```diff
@@ -0,0 +1,47 @@
+package com.twitter.finagle.redis
+
+import _root_.java.lang.{Long => JLong,Boolean => JBoolean}
```
vkostyukov: Does it actually need `_root_` prefix? Seems to me just `java.lang` should work fine.

travisbrown: Looks unnecessary to me as well, but it's in all the other `*Commands.scala` files, too.

There's a fair amount of noise and historical baggage in finagle-redis, and since it's more or less in maintenance mode until we move it to the Finagle org it's been a relatively low priority. So we have to decide between 1) doing the right thing here and being inconsistent, 2) cleaning up all instances of this at once, or 3) leaving it as is. I'm fine with either 1) or 3), but think 2) should happen in a new PR if at all.

vkostyukov: I see. Sorry. Thanks for the explanation Travis! In this case we definitely should "follow the style" at least in this PR.

luciferous:
LGTM

vkostyukov:
:+1:

stevegury:
LGTM

travisbrown:
Okay, I'm pulling this inside. Thanks all.

travisbrown:
This has been merged internally, so it'll make it into the next release. I'll close this issue when the commit gets synced to GitHub. Thanks again, @charithe!

charithe:
That's great! Thank you.

travisbrown:
This is now available [in the `develop` branch](https://github.com/twitter/finagle/commit/036c693932cc21658f2fa3d8223125b7083d58fd) on GitHub (but not yet in a public release).

PR #363, opened 2015-04-08
Make Trace's internal stack thread local

jamescway:
Trace's global stack variable is not thread safe. Changing to be thread local for thread safety.  

We are using finagle-thrift as part of tracing service calls with Zipkin.

This issue manifested itself with Jruby 1.7 and a multi-threaded application.

### DIFF
--- a/finagle-thrift/src/main/ruby/lib/finagle-thrift/trace.rb
+++ b/finagle-thrift/src/main/ruby/lib/finagle-thrift/trace.rb
@@ -2,6 +2,7 @@ module Trace
   extend self
   DEFAULT_SAMPLE_RATE = 0.001
   TRACE_ID_UPPER_BOUND = 2 ** 64
+  TRACE_STACK = :trace_stack
 
   def id
     if stack.empty?
@@ -33,7 +34,7 @@ module Trace
         saved_stack = stack.dup
         yield
       ensure
-        @stack = saved_stack
+        stack = saved_stack
       end
     end
   end
@@ -85,11 +86,11 @@ module Trace
       "TraceId(trace_id = #{@trace_id.to_s}, parent_id = #{@parent_id.to_s}, span_id = #{@span_id.to_s}, sampled = #{@sampled.to_s}, flags = #{@flags.to_s})"
     end
   end
-  
+
   # there are a total of 64 flags that can be passed down with the various tracing headers
   # at the time of writing only one is used (debug).
   #
-  # Note that using the 64th bit in Ruby requires some sign conversion since Thrift i64s are signed 
+  # Note that using the 64th bit in Ruby requires some sign conversion since Thrift i64s are signed
   # but Ruby won't do the right thing if you try to set 1 << 64
   class Flags
     # no flags set
@@ -146,8 +147,14 @@ module Trace
 
   private
 
+  # "stack" acts as a thread local variable and cannot be shared between
+  # threads.
+  def stack=(stack)
+    Thread.current[TRACE_STACK] = stack
+  end
+
   def stack
-    @stack ||= []
+    Thread.current[TRACE_STACK] ||= []
   end
 
   def tracer
--- a/finagle-thrift/src/main/ruby/.gitignore
+++ b/finagle-thrift/src/main/ruby/.gitignore
@@ -1,0 +1,1 @@
+Gemfile.lock
--- a/finagle-thrift/src/main/ruby/Gemfile
+++ b/finagle-thrift/src/main/ruby/Gemfile
@@ -1,0 +1,5 @@
+source 'https://rubygems.org'
+
+# Specify your gem's dependencies in finagle-thrift.gemspec
+gemspec
+
--- a/finagle-thrift/src/main/ruby/lib/finagle-thrift/version.rb
+++ b/finagle-thrift/src/main/ruby/lib/finagle-thrift/version.rb
@@ -1,4 +1,4 @@
 module FinagleThrift
-  VERSION = "1.4.0"
+  VERSION = "1.4.1"
 end
 
--- a/finagle-thrift/src/main/ruby/test/test_helper.rb
+++ b/finagle-thrift/src/main/ruby/test/test_helper.rb
@@ -1,0 +1,10 @@
+base_dir = File.expand_path(File.join(File.dirname(__FILE__), ".."))
+lib_dir  = File.join(base_dir, "lib")
+test_dir = File.join(base_dir, "test")
+
+$LOAD_PATH.unshift(lib_dir)
+
+require 'test/unit'
+require 'rubygems'
+require 'thrift'
+require 'finagle-thrift'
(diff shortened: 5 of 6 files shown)


finagle-thrift/src/main/ruby/lib/finagle-thrift/trace.rb
```diff
@@ -57,6 +57,10 @@ def tracer=(tracer)
     @tracer = tracer
   end
 
+  def stack=(stack)
```
mosesn: I don't know ruby, so please be patient with me :)

It looks like this is sticking the stack into a thread-local variable.  Is our guarantee for ruby tracing that you won't ever switch threads?

jamescway: @mosesn indeed it is a thread local variable.  I hope I'm understanding your question correctly, I think we are trying to do one trace object per thread.  We were getting race conditions on the stack obj.  I think Ruby does context switching from what I read [here](https://ruby-hacking-guide.github.io/thread.html), but since its thread local it shouldn't be problem.  Also, we're using jruby so they should be real threads.

jamescway:
@mosesn sorry about the delay.  We were testing this change in an integration environment and everything looks good!

finagle-thrift/src/main/ruby/lib/finagle-thrift/trace.rb
mosesn: My concern is what happens after a ruby server gets a request.  Is it ever legal to have something like:

t1: receives a request
t1: makes a request to a different server
t1: does some different work
t1: puts some work on a different thread, t2
t2: on behalf of t1, makes a request to a different server

... etc.

In this case, they should all logically be part of the same Trace, but if we are storing these as thread-locals, then they won't come along by default when we switch threads.

mosesn:
LGTM, but let's document that if you switch threads, things on the new thread will not be traced.

finagle-thrift/src/main/ruby/lib/finagle-thrift/trace.rb
jamescway: I see your point, definitely this issue could happen.  A few possible approaches here:
1.  Thread local - the architecture would require requests to be handled in T1 and if it requires additional work, it would pass the Trace data necessary to T2. 
2. Global - guarded with mutexes synchronized code.  This approach seems like a bit of a band aid and not solving the real problem.

What do you think?

I will definitely add this to the readme.

mosesn: We handle it by requiring that you use twitter's util libraries if you switch thread in java, but it won't be so simple in ruby.  I think for now we should just document the behavior, and then we can deal with it later if you actually need it.  Explicitly passing the trace data will probably be fine as a first approximation.

Grabbing a global mutex would require that you only serve one request at a time, which seems like it defeats the purpose.

jamescway:
@mosesn updated with README.md.  Should I make the PR off of develop?

finagle-thrift/src/main/ruby/README.md
```diff
@@ -0,0 +1,13 @@
+# Finable Thrift - Trace Ruby Thread Saftey
```
mosesn: s/finable thrift/finagle

No need to specify thrift

s/saftey/safety

finagle-thrift/src/main/ruby/README.md
```diff
@@ -0,0 +1,13 @@
...
+Design decisions behind thread saftey in Trace
+
+```ruby
+  def stack
+    Thread.current[TRACE_STACK] ||= []
+  end
+```
+
+## Notes
+[Trace](https://github.com/jamescway/finagle/blob/safe/finagle-thrift/src/main/ruby/lib/finagle-thrift/trace.rb) previously implemented an internal global variable "stack".  The assumption was that there would be 1 request per processes.  In a multi-thread ruby environment (i.e. jruby) this variable is now open to race conditions.  In order to fix this issue the stack variable was modified to be thread local.  The benefits are that the variable is local and only the local thread can modify it.  However, if the there is some work which has been delegated to another thread (T2) then that thread will not have access to the global.  One possible approach for delegating work to another thread might be to pass the stack to T2 to allow it to perform work.  This would then only allow T2 to access the thread.  Generally, sharing the stack between threads would be disallowed.  This leaves room for a future PR to add in this sharing functionality.
```
mosesn: Instead of one long line, could you truncate at 80 characters?

finagle-thrift/src/main/ruby/README.md
```diff
@@ -0,0 +1,13 @@
+# Finable Thrift - Trace Ruby Thread Saftey
+
+## Synopsis
+Design decisions behind thread saftey in Trace
```
mosesn: I'm not sure this makes sense from a user perspective.  Most folks reading the README probably don't care terribly about the design decision, and probably care more about how to use it.  If you want to document the design decision, probably best to do that in the code.  Could we keep the README just focusing on what is a valid use of tracing?

jamescway:
@mosesn a few Q's:
* do u guys merge internally?
* should I make a PR off of develop?
* what's the release cycle of finagle-thrift? (seems like not too often)

sorry to bug :smile: :beers:

travisbrown:
@jamescway 

* Yep, this will be tested and merged internally (see the [contributing guide](https://github.com/twitter/finagle/blob/develop/CONTRIBUTING.md#code-review) for details).
* In general that's the new recommendation, but this rebases cleanly so I wouldn't worry about it.
* All the Finagle modules are published together, and releases are currently pretty far apart (the last was in December), but we're hoping to tighten up the cycle.

ryangreenberg:
Overall changes looks fine to me with the already stated caveat that it won't work in applications that handle multiple requests concurrently in a single thread. We can come up with a solution to that problem when it arrives. Thanks for making this change.

mosesn:
Can add a commit message that conforms to https://github.com/twitter/finagle/blob/master/CONTRIBUTING.md#style?  That would be awesome!  LGTM otherwise.

finagle-thrift/src/main/ruby/README.md
```diff
@@ -0,0 +1,10 @@
+# Finagle - Trace
```
nshkrob: I don't think this belongs in a README. This is already explained in the comment below. Remove this file altogether?

(A README file should give an idea of what this code is doing and how it should be used)

jamescway:
Ok removed the readme file while preserving the reason for the change in comment and updated commit message (per style guide).  :bowtie:

nshkrob:
LGTM, I've pulled this internally. Should show up here in a few days. Thanks!

jamescway:
@nshkrob Do you think you guys will be able to do a release of finable-thrift to rubygems.org?

nshkrob:
Sure, let me look into that.

nshkrob:
@jamescway Done. Let me know if that doesn't work for you.

PR #364, opened 2015-04-10
Making PreparedStatement parameters type-safe

missingfaktor:
## Problem 

In the current Finagle MySQL, `PreparedStatement#apply` has a type `Seq[Any] => Future[Result]`. This means that one might accidentally end up passing things that cannot be written, and be in for surprises. (Happened to us quite a few times now.) This might be revealed in the warnings and the tests, but we can do much better: We can eliminate such errors at compile time itself.

## Solution

Change `PreparedToStatement#apply` to have type `Seq[Parameter] => Future[Result]`. `Parameter` is a class [that makes use of existential types](http://iveselov.info/posts/2012-08-30-existential-types.html) and packs a value `A` and an implicit instance `CanBeParameter[A]` together. `CanBeParameter` is a type-class. If you have an instance of `CanBeParameter[A]` available, that serves as an evidence that, well, a value of `A` can be a used as a parameter. 

**This PR ensures that all the types that were supported before are still supported the same way.**

`null` required special handling in a couple of places:

1. `Parameter.wrap` doesn't kick in for `null` because `null` is already a member of `Parameter` type. So we need to sanitize the `Seq[Parameter]` in `ExecuteRequest#apply`, and replace all `null`s with `NullParameter`.
2. If there is a `null` with a static type `A` for which there is a `CanBeParameter[A]` evidence available, then we might end up with, for example, a `stringCanBeParameter` instance (if `A` is `String`) and `typeCode` etc will yield wrong values. To avoid this, `Parameter.wrap` has to guard for `null` and return a `NullParameter` in that case.

This PR retains the support for `null` as is, for backwards compatibility. 

Going forward, we could stop supporting `null` altogether and avoid the complications I talked about before. We could instead support `Option` which is a more principled way to deal with these sort of things.

### DIFF
--- a/finagle-mysql/src/main/scala/com/twitter/finagle/mysql/CanBeParameter.scala
+++ b/finagle-mysql/src/main/scala/com/twitter/finagle/mysql/CanBeParameter.scala
@@ -1,0 +1,182 @@
+package com.twitter.finagle.exp.mysql
+
+import com.twitter.finagle.exp.mysql.transport.{Buffer, BufferWriter}
+
+trait CanBeParameter[-A] { outer =>
+  /**
+   * Returns the size of the given parameter in its MySQL binary representation.
+   **/
+  def sizeOf(param: A): Int
+
+  /**
+   * Retrieves the MySQL type code for the given parameter.
+   * */
+  def typeCode(param: A): Short
+
+  def write(writer: BufferWriter, param: A): Unit
+}
+
+object CanBeParameter {
+  implicit val stringCanBeParameter = {
+    new CanBeParameter[String] {
+      def sizeOf(param: String): Int = {
+        val bytes = param.getBytes(Charset.defaultCharset)
+        Buffer.sizeOfLen(bytes.size) + bytes.size
+      }
+
+      def typeCode(param: String) = Type.VarChar
+      def write(writer: BufferWriter, param: String) = writer.writeLengthCodedString(param)
+    }
+  }
+
+  implicit val booleanCanBeParameter = {
+    new CanBeParameter[Boolean] {
+      def sizeOf(param: Boolean) = 1
+      def typeCode(param: Boolean) = Type.Tiny
+      def write(writer: BufferWriter, param: Boolean) = writer.writeBoolean(param)
+    }
+  }
+
+  implicit val byteCanBeParameter = {
+    new CanBeParameter[Byte] {
+      def sizeOf(param: Byte) = 1
+      def typeCode(param: Byte) = Type.Tiny
+      def write(writer: BufferWriter, param: Byte) = writer.writeByte(param)
+    }
+  }
+
+  implicit val shortCanBeParameter = {
+    new CanBeParameter[Short] {
+      def sizeOf(param: Short) = 2
+      def typeCode(param: Short) = Type.Short
+      def write(writer: BufferWriter, param: Short) = writer.writeShort(param)
+    }
+  }
+
+  implicit val intCanBeParameter = {
+    new CanBeParameter[Int] {
+      def sizeOf(param: Int) = 4
+      def typeCode(param: Int) = Type.Long
+      def write(writer: BufferWriter, param: Int) = writer.writeInt(param)
+    }
+  }
+
+  implicit val longCanBeParameter = {
+    new CanBeParameter[Long] {
+      def sizeOf(param: Long) = 8
+      def typeCode(param: Long) = Type.LongLong
+      def write(writer: BufferWriter, param: Long) = writer.writeLong(param)
+    }
+  }
+
+  implicit val floatCanBeParameter = {
+    new CanBeParameter[Float] {
+      def sizeOf(param: Float) = 4
+      def typeCode(param: Float) = Type.Float
+      def write(writer: BufferWriter, param: Float) = writer.writeFloat(param)
+    }
+  }
+
+  implicit val doubleCanBeParameter = {
+    new CanBeParameter[Double] {
+      def sizeOf(param: Double) = 8
+      def typeCode(param: Double) = Type.Double
+      def write(writer: BufferWriter, param: Double) = writer.writeDouble(param)
+    }
+  }
+
+  implicit val byteArrayCanBeParameter = {
+    new CanBeParameter[Array[Byte]] {
+      def sizeOf(param: Array[Byte]) = Buffer.sizeOfLen(param.length) + param.length
+      def typeCode(param: Array[Byte]) = {
+        if (param.length <= 255) Type.TinyBlob
+        else if (param.length <= 65535) Type.Blob
+        else if (param.length <= 16777215) Type.MediumBlob
+        else -1
+      }
+      def write(writer: BufferWriter, param: Array[Byte]) = writer.writeLengthCodedBytes(param)
+    }
+  }
+
+  implicit val valueCanBeParameter = {
+    new CanBeParameter[Value] {
+      def sizeOf(param: Value) = param match {
+        case RawValue(_, _, true, b) => Buffer.sizeOfLen(b.length) + b.length
+        case StringValue(s)          => val bytes = s.getBytes(Charset.defaultCharset); Buffer.sizeOfLen(bytes.size) + bytes.size
+        case ByteValue(_)            => 1
+        case ShortValue(_)           => 2
+        case IntValue(_)             => 4
+        case LongValue(_)            => 8
+        case FloatValue(_)           => 4
+        case DoubleValue(_)          => 8
+        case NullValue               => 0
+        case _                       => 0
+      }
+
+      def typeCode(param: Value) = param match {
+        case RawValue(typ, _, _, _) => typ
+        case StringValue(_)         => Type.VarChar
+        case ByteValue(_)           => Type.Tiny
+        case ShortValue(_)          => Type.Short
+        case IntValue(_)            => Type.Long
+        case LongValue(_)           => Type.LongLong
+        case FloatValue(_)          => Type.Float
+        case DoubleValue(_)         => Type.Double
+        case EmptyValue             => -1
+        case NullValue              => Type.Null
+      }
+
+      def write(writer: BufferWriter, param: Value) = param match {
+        // allows for generic binary values as params to a prepared statement.
+        case RawValue(_, _, true, bytes) => writer.writeLengthCodedBytes(bytes)
+        // allows for Value types as params to prepared statements
+        case ByteValue(b)                => writer.writeByte(b)
+        case ShortValue(s)               => writer.writeShort(s)
+        case IntValue(i)                 => writer.writeInt(i)
+        case LongValue(l)                => writer.writeLong(l)
+        case FloatValue(f)               => writer.writeFloat(f)
+        case DoubleValue(d)              => writer.writeDouble(d)
+        case StringValue(s)              => writer.writeLengthCodedString(s)
+        case _                           => ()
+      }
+    }
+  }
+
+  implicit val timestampCanBeParameter = {
+    new CanBeParameter[java.sql.Timestamp] {
+      def sizeOf(param: java.sql.Timestamp) = 12
+      def typeCode(param: java.sql.Timestamp) = Type.Timestamp
+      def write(writer: BufferWriter, param: java.sql.Timestamp) = {
+        valueCanBeParameter.write(writer, TimestampValue(param))
+      }
+    }
+  }
+
+  implicit val sqlDateCanBeParameter = {
+    new CanBeParameter[java.sql.Date] {
+      def sizeOf(param: java.sql.Date) = 5
+      def typeCode(param: java.sql.Date) = Type.Date
+      def write(writer: BufferWriter, param: java.sql.Date) = {
+        valueCanBeParameter.write(writer, DateValue(param))
+      }
+    }
+  }
+
+  implicit val javaDateCanBeParameter = {
+    new CanBeParameter[java.util.Date] {
+      def sizeOf(param: java.util.Date) = 12
+      def typeCode(param: java.util.Date) = Type.DateTime
+      def write(writer: BufferWriter, param: java.util.Date) = {
+        valueCanBeParameter.write(writer, TimestampValue(new java.sql.Timestamp(param.getTime)))
+      }
+    }
+  }
+
+  implicit val nullCanBeParameter = {
+    new CanBeParameter[Null] {
+      def sizeOf(param: Null) = 0
+      def typeCode(param: Null) = Type.Null
+      def write(writer: BufferWriter, param: Null): Unit = ()
+    }
+  }
+}
--- a/finagle-mysql/src/main/scala/com/twitter/finagle/mysql/CanBeParameter.scala
+++ b/finagle-mysql/src/main/scala/com/twitter/finagle/mysql/CanBeParameter.scala
@@ -2,7 +2,7 @@ package com.twitter.finagle.exp.mysql
 
 import com.twitter.finagle.exp.mysql.transport.{Buffer, BufferWriter}
 
-trait CanBeParameter[-A] { outer =>
+trait CanBeParameter[A] { outer =>
   /**
    * Returns the size of the given parameter in its MySQL binary representation.
    **/
@@ -13,15 +13,15 @@ trait CanBeParameter[-A] { outer =>
    * */
   def typeCode(param: A): Short
 
-  def write(writer: BufferWriter, param: A): Unit
+  def write(writer: BufferWriter, param: A): BufferWriter
 }
 
 object CanBeParameter {
-  implicit val stringCanBeParameter = {
+  implicit val stringCanBeParameter: CanBeParameter[String] = {
     new CanBeParameter[String] {
       def sizeOf(param: String): Int = {
         val bytes = param.getBytes(Charset.defaultCharset)
-        Buffer.sizeOfLen(bytes.size) + bytes.size
+        Buffer.sizeOfLen(bytes.length) + bytes.length
       }
 
       def typeCode(param: String) = Type.VarChar
@@ -29,7 +29,7 @@ object CanBeParameter {
     }
   }
 
-  implicit val booleanCanBeParameter = {
+  implicit val booleanCanBeParameter: CanBeParameter[Boolean] = {
     new CanBeParameter[Boolean] {
       def sizeOf(param: Boolean) = 1
       def typeCode(param: Boolean) = Type.Tiny
@@ -37,7 +37,7 @@ object CanBeParameter {
     }
   }
 
-  implicit val byteCanBeParameter = {
+  implicit val byteCanBeParameter: CanBeParameter[Byte] = {
     new CanBeParameter[Byte] {
       def sizeOf(param: Byte) = 1
       def typeCode(param: Byte) = Type.Tiny
@@ -45,7 +45,7 @@ object CanBeParameter {
     }
   }
 
-  implicit val shortCanBeParameter = {
+  implicit val shortCanBeParameter: CanBeParameter[Short] = {
     new CanBeParameter[Short] {
       def sizeOf(param: Short) = 2
       def typeCode(param: Short) = Type.Short
@@ -53,7 +53,7 @@ object CanBeParameter {
     }
   }
 
-  implicit val intCanBeParameter = {
+  implicit val intCanBeParameter: CanBeParameter[Int] = {
     new CanBeParameter[Int] {
       def sizeOf(param: Int) = 4
       def typeCode(param: Int) = Type.Long
@@ -61,7 +61,7 @@ object CanBeParameter {
     }
   }
 
-  implicit val longCanBeParameter = {
+  implicit val longCanBeParameter: CanBeParameter[Long] = {
     new CanBeParameter[Long] {
       def sizeOf(param: Long) = 8
       def typeCode(param: Long) = Type.LongLong
@@ -69,7 +69,7 @@ object CanBeParameter {
     }
   }
 
-  implicit val floatCanBeParameter = {
+  implicit val floatCanBeParameter: CanBeParameter[Float] = {
     new CanBeParameter[Float] {
       def sizeOf(param: Float) = 4
       def typeCode(param: Float) = Type.Float
@@ -77,7 +77,7 @@ object CanBeParameter {
     }
   }
 
-  implicit val doubleCanBeParameter = {
+  implicit val doubleCanBeParameter: CanBeParameter[Double] = {
     new CanBeParameter[Double] {
       def sizeOf(param: Double) = 8
       def typeCode(param: Double) = Type.Double
@@ -85,7 +85,7 @@ object CanBeParameter {
     }
   }
 
-  implicit val byteArrayCanBeParameter = {
+  implicit val byteArrayCanBeParameter: CanBeParameter[Array[Byte]] = {
     new CanBeParameter[Array[Byte]] {
       def sizeOf(param: Array[Byte]) = Buffer.sizeOfLen(param.length) + param.length
       def typeCode(param: Array[Byte]) = {
@@ -98,19 +98,20 @@ object CanBeParameter {
     }
   }
 
-  implicit val valueCanBeParameter = {
+  implicit val valueCanBeParameter: CanBeParameter[Value] = {
     new CanBeParameter[Value] {
       def sizeOf(param: Value) = param match {
-        case RawValue(_, _, true, b) => Buffer.sizeOfLen(b.length) + b.length
-        case StringValue(s)          => val bytes = s.getBytes(Charset.defaultCharset); Buffer.sizeOfLen(bytes.size) + bytes.size
-        case ByteValue(_)            => 1
-        case ShortValue(_)           => 2
-        case IntValue(_)             => 4
-        case LongValue(_)            => 8
-        case FloatValue(_)           => 4
-        case DoubleValue(_)          => 8
-        case NullValue               => 0
-        case _                       => 0
+        case RawValue(_, _, true, b)  => Buffer.sizeOfLen(b.length) + b.length
+        case StringValue(s)           => val bytes = s.getBytes(Charset.defaultCharset); Buffer.sizeOfLen(bytes.length) + bytes.length
+        case ByteValue(_)             => 1
+        case ShortValue(_)            => 2
+        case IntValue(_)              => 4
+        case LongValue(_)             => 8
+        case FloatValue(_)            => 4
+        case DoubleValue(_)           => 8
+        case EmptyValue               => 0
+        case NullValue                => 0
+        case RawValue(_, _, false, _) => 0
       }
 
       def typeCode(param: Value) = param match {
@@ -137,12 +138,14 @@ object CanBeParameter {
         case FloatValue(f)               => writer.writeFloat(f)
         case DoubleValue(d)              => writer.writeDouble(d)
         case StringValue(s)              => writer.writeLengthCodedString(s)
-        case _                           => ()
+        case EmptyValue                  => writer
+        case NullValue                   => writer
+        case RawValue(_, _, false, _)    => writer
       }
     }
   }
 
-  implicit val timestampCanBeParameter = {
+  implicit val timestampCanBeParameter: CanBeParameter[java.sql.Timestamp] = {
     new CanBeParameter[java.sql.Timestamp] {
       def sizeOf(param: java.sql.Timestamp) = 12
       def typeCode(param: java.sql.Timestamp) = Type.Timestamp
@@ -152,7 +155,7 @@ object CanBeParameter {
     }
   }
 
-  implicit val sqlDateCanBeParameter = {
+  implicit val sqlDateCanBeParameter: CanBeParameter[java.sql.Date] = {
     new CanBeParameter[java.sql.Date] {
       def sizeOf(param: java.sql.Date) = 5
       def typeCode(param: java.sql.Date) = Type.Date
@@ -162,7 +165,7 @@ object CanBeParameter {
     }
   }
 
-  implicit val javaDateCanBeParameter = {
+  implicit val javaDateCanBeParameter: CanBeParameter[java.util.Date] = {
     new CanBeParameter[java.util.Date] {
       def sizeOf(param: java.util.Date) = 12
       def typeCode(param: java.util.Date) = Type.DateTime
@@ -172,11 +175,11 @@ object CanBeParameter {
     }
   }
 
-  implicit val nullCanBeParameter = {
+  implicit val nullCanBeParameter: CanBeParameter[Null] = {
     new CanBeParameter[Null] {
       def sizeOf(param: Null) = 0
       def typeCode(param: Null) = Type.Null
-      def write(writer: BufferWriter, param: Null): Unit = ()
+      def write(writer: BufferWriter, param: Null): BufferWriter = writer
     }
   }
 }
--- a/finagle-mysql/src/main/scala/com/twitter/finagle/mysql/CanBeParameter.scala
+++ b/finagle-mysql/src/main/scala/com/twitter/finagle/mysql/CanBeParameter.scala
@@ -2,15 +2,15 @@ package com.twitter.finagle.exp.mysql
 
 import com.twitter.finagle.exp.mysql.transport.{Buffer, BufferWriter}
 
-trait CanBeParameter[A] { outer =>
+trait CanBeParameter[-A] { outer =>
   /**
    * Returns the size of the given parameter in its MySQL binary representation.
-   **/
+   */
   def sizeOf(param: A): Int
 
   /**
    * Retrieves the MySQL type code for the given parameter.
-   * */
+   */
   def typeCode(param: A): Short
 
   def write(writer: BufferWriter, param: A): BufferWriter
--- a/finagle-mysql/src/main/scala/com/twitter/finagle/mysql/Client.scala
+++ b/finagle-mysql/src/main/scala/com/twitter/finagle/mysql/Client.scala
@@ -23,7 +23,7 @@ object Client {
 
     def prepare(sql: String): PreparedStatement = new PreparedStatement {
       def apply(ps: Parameter*): Future[Result] = factory() flatMap { svc =>
-        svc(PrepareRequest(sql)) flatMap {
+        svc(PrepareRequest(sql)).flatMap {
           case ok: PrepareOK => svc(ExecuteRequest(ok.id, ps.toIndexedSeq))
           case r => Future.exception(new Exception("Unexpected result %s when preparing %s"
             .format(r, sql)))
--- a/finagle-mysql/src/main/scala/com/twitter/finagle/mysql/Client.scala
+++ b/finagle-mysql/src/main/scala/com/twitter/finagle/mysql/Client.scala
@@ -22,7 +22,7 @@ object Client {
       }
 
     def prepare(sql: String): PreparedStatement = new PreparedStatement {
-      def apply(ps: Any*): Future[Result] = factory() flatMap { svc =>
+      def apply(ps: Parameter*): Future[Result] = factory() flatMap { svc =>
         svc(PrepareRequest(sql)) flatMap {
           case ok: PrepareOK => svc(ExecuteRequest(ok.id, ps.toIndexedSeq))
           case r => Future.exception(new Exception("Unexpected result %s when preparing %s"
--- a/finagle-mysql/src/main/scala/com/twitter/finagle/mysql/Parameter.scala
+++ b/finagle-mysql/src/main/scala/com/twitter/finagle/mysql/Parameter.scala
@@ -1,0 +1,37 @@
+package com.twitter.finagle.exp.mysql
+
+import com.twitter.finagle.exp.mysql.transport.BufferWriter
+import language.implicitConversions
+
+trait Parameter {
+  type A
+  def value: A
+  def evidence: CanBeParameter[A]
+
+  final def writeTo(writer: BufferWriter): Unit = {
+    evidence.write(writer, value)
+  }
+
+  final def size = evidence.sizeOf(value)
+  final def typeCode = evidence.typeCode(value)
+}
+
+object Parameter {
+  implicit def wrap[_A](_value: _A)(implicit _evidence: CanBeParameter[_A]): Parameter = {
+    if (_value == null) {
+      NullParameter
+    } else {
+      new Parameter {
+        type A = _A
+        def value: A = _value
+        def evidence: CanBeParameter[A] = _evidence
+      }
+    }
+  }
+
+  object NullParameter extends Parameter {
+    type A = Null
+    def value = null
+    def evidence = CanBeParameter.nullCanBeParameter
+  }
+}
--- a/finagle-mysql/src/main/scala/com/twitter/finagle/mysql/Parameter.scala
+++ b/finagle-mysql/src/main/scala/com/twitter/finagle/mysql/Parameter.scala
@@ -3,7 +3,11 @@ package com.twitter.finagle.exp.mysql
 import com.twitter.finagle.exp.mysql.transport.BufferWriter
 import language.implicitConversions
 
-trait Parameter {
+/**
+ * A value of type `A` can implicitly convert to a `Parameter` if an evidence `CanBeParameter[A]` is
+ * available in scope. This type is not be instantiated in any other manner.
+ */
+sealed trait Parameter {
   type A
   def value: A
   def evidence: CanBeParameter[A]
@@ -12,8 +16,8 @@ trait Parameter {
     evidence.write(writer, value)
   }
 
-  final def size = evidence.sizeOf(value)
-  final def typeCode = evidence.typeCode(value)
+  final def size: Int = evidence.sizeOf(value)
+  final def typeCode: Short = evidence.typeCode(value)
 }
 
 object Parameter {
--- a/finagle-mysql/src/main/scala/com/twitter/finagle/mysql/Parameter.scala
+++ b/finagle-mysql/src/main/scala/com/twitter/finagle/mysql/Parameter.scala
@@ -17,14 +17,14 @@ trait Parameter {
 }
 
 object Parameter {
-  implicit def wrap[_A](_value: _A)(implicit _evidence: CanBeParameter[_A]): Parameter = {
-    if (_value == null) {
+  implicit def wrap[A0](value0: A0)(implicit evidence0: CanBeParameter[A0]): Parameter = {
+    if (value0 == null) {
       NullParameter
     } else {
       new Parameter {
-        type A = _A
-        def value: A = _value
-        def evidence: CanBeParameter[A] = _evidence
+        type A = A0
+        def value: A = value0
+        def evidence: CanBeParameter[A] = evidence0
       }
     }
   }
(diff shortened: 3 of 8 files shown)


finagle-mysql/src/test/scala/com/twitter/finagle/mysql/ParameterTest.scala
```diff
@@ -0,0 +1,28 @@
...
+import com.twitter.finagle.exp.mysql.Parameter
+import com.twitter.finagle.exp.mysql.Parameter.NullParameter
+import org.junit.runner.RunWith
+import org.scalatest.FunSuite
+import org.scalatest.junit.JUnitRunner
+
+@RunWith(classOf[JUnitRunner])
+class ParameterTest extends FunSuite {
+  test("Parameter coercion") {
+    locally {
```
mosesn: instead of locally blocks, can we just separate these out into clearly named tests?  that will help us understand what a test failure means if it happens.

finagle-mysql/src/test/scala/com/twitter/finagle/mysql/unit/RequestTest.scala
```diff
@@ -2,6 +2,7 @@ package com.twitter.finagle.exp.mysql
 
 import java.sql.{Timestamp, Date => SQLDate}
 import java.util.{Calendar, Date, TimeZone}
+import com.twitter.finagle.exp.mysql.Parameter.NullParameter
```
mosesn: please keep imports sorted

finagle-mysql/src/main/scala/com/twitter/finagle/mysql/Parameter.scala
```diff
@@ -0,0 +1,37 @@
...
+  final def writeTo(writer: BufferWriter): Unit = {
+    evidence.write(writer, value)
+  }
+
+  final def size = evidence.sizeOf(value)
+  final def typeCode = evidence.typeCode(value)
+}
+
+object Parameter {
+  implicit def wrap[_A](_value: _A)(implicit _evidence: CanBeParameter[_A]): Parameter = {
```
travisbrown: I'm nitpicking here, but maybe we could go with the `A0`, `value0` convention since `_` is already so massively overloaded in Scala? I'm not sure there's precedent for either convention in Finagle right now.

finagle-mysql/src/test/scala/com/twitter/finagle/mysql/ParameterTest.scala
missingfaktor: @mosesn Good suggestion. Will do.

finagle-mysql/src/main/scala/com/twitter/finagle/mysql/Request.scala
```diff
@@ -258,6 +231,20 @@ case class ExecuteRequest(
     }
 }
 
+object ExecuteRequest {
+  def apply(stmtId: Int, params: IndexedSeq[Parameter] = IndexedSeq.empty, hasNewParams: Boolean = true, flags: Byte = 0) = {
```
mosesn: please keep line length < 100.

if you need to do this sanitization for every request, you should disallow creation otherwise.  does this need to be here though?  could we just put it in the constructor of ExecuteRequest?

finagle-mysql/src/main/scala/com/twitter/finagle/mysql/Request.scala
```diff
@@ -151,33 +151,31 @@ case class HandshakeResponse(
  * a prepared statement.
  * [[http://dev.mysql.com/doc/internals/en/com-stmt-execute.html]]
  */
-case class ExecuteRequest(
-  stmtId: Int,
-  params: IndexedSeq[Any] = IndexedSeq.empty,
-  hasNewParams: Boolean = true,
-  flags: Byte = 0
+class ExecuteRequest(
```
mosesn: imo let's keep this a case class.  why does it need to change?

finagle-mysql/src/main/scala/com/twitter/finagle/mysql/Request.scala
```diff
@@ -151,33 +151,31 @@ case class HandshakeResponse(
...
-    private[this] def makeNullBitmap(parameters: IndexedSeq[Any]): Array[Byte] = {
+    private[this] def makeNullBitmap(parameters: IndexedSeq[Parameter]): Array[Byte] = {
       val bitmap = new Array[Byte]((parameters.size + 7) / 8)
       val ps = parameters.zipWithIndex
-      for ((p, idx) <- ps if isNull(p)) {
-        val bytePos = idx / 8
-        val bitPos = idx % 8
-        val byte = bitmap(bytePos)
-        bitmap(bytePos) = (byte | (1 << bitPos)).toByte
+      ps foreach {
```
mosesn: prefer ps.foreach

finagle-mysql/src/test/scala/com/twitter/finagle/mysql/unit/RequestTest.scala
missingfaktor: @mosesn Tried removing that line and re-auto-importing. For some reason, IDEA keeps putting it on the same line! Will move it manually and push.

finagle-mysql/src/main/scala/com/twitter/finagle/mysql/Parameter.scala
missingfaktor: @travisbrown I am not sure there is a precedent for either convention in Scala, since it's not a very widely used pattern. But I agree that `A0` might be a better convention that `_A`. Will change.

finagle-mysql/src/main/scala/com/twitter/finagle/mysql/Request.scala
```diff
@@ -192,40 +190,15 @@ case class ExecuteRequest(
      * Returns sizeof all the parameters according to
      * mysql binary encoding.
      */
-    private[this] def sizeOfParameters(parameters: IndexedSeq[Any]): Int =
-      parameters.foldLeft(0) { (sum, param) =>
-        sum + Type.sizeOf(param)
-      }
+    private[this] def sizeOfParameters(parameters: IndexedSeq[Parameter]): Int =
+      parameters.foldLeft(0)(_ + _.size)
```
mosesn: feel free to ignore: maybe simpler as

```scala
parameters.map(_.size).sum
```

finagle-mysql/src/main/scala/com/twitter/finagle/mysql/Request.scala
missingfaktor: Okay re: line length. 

Putting it in the constructor will be messy because of how Scala does constructors. Every parameter to class is accessible in its entire body, and also it's immutable. We will have something like:

```scala
class ExecuteRequest(..., params: Seq[Parameter], ...) {
  val sanitizedParams = sanitize(params)

  // but params is still accessible here, what if we accidentally use that?
}
```

finagle-mysql/src/main/scala/com/twitter/finagle/mysql/CanBeParameter.scala
```diff
@@ -0,0 +1,182 @@
...
+   * Returns the size of the given parameter in its MySQL binary representation.
+   **/
+  def sizeOf(param: A): Int
+
+  /**
+   * Retrieves the MySQL type code for the given parameter.
+   * */
+  def typeCode(param: A): Short
+
+  def write(writer: BufferWriter, param: A): Unit
```
mosesn: how about we make this return `BufferWriter`, like the `writeX` methods in BufferWriter?

finagle-mysql/src/main/scala/com/twitter/finagle/mysql/CanBeParameter.scala
```diff
@@ -0,0 +1,182 @@
...
+trait CanBeParameter[-A] { outer =>
+  /**
+   * Returns the size of the given parameter in its MySQL binary representation.
+   **/
+  def sizeOf(param: A): Int
+
+  /**
+   * Retrieves the MySQL type code for the given parameter.
+   * */
+  def typeCode(param: A): Short
```
mosesn: this doesn't seem to need to take an argument.

finagle-mysql/src/main/scala/com/twitter/finagle/mysql/Request.scala
missingfaktor: @mosesn Because of the sanitization that's needed beforehand. With a regular class, I could mark the constructor private. With a case class, I still could do that, but the `apply` auto-generated in the companion object will be publicly available, and it cannot even be overridden.

finagle-mysql/src/main/scala/com/twitter/finagle/mysql/Request.scala
missingfaktor: Okay.

finagle-mysql/src/main/scala/com/twitter/finagle/mysql/Request.scala
missingfaktor: Yes. Also just a tiny bit more expensive. :)

finagle-mysql/src/main/scala/com/twitter/finagle/mysql/CanBeParameter.scala
```diff
@@ -0,0 +1,182 @@
...
+
+  def write(writer: BufferWriter, param: A): Unit
+}
+
+object CanBeParameter {
+  implicit val stringCanBeParameter = {
+    new CanBeParameter[String] {
+      def sizeOf(param: String): Int = {
+        val bytes = param.getBytes(Charset.defaultCharset)
+        Buffer.sizeOfLen(bytes.size) + bytes.size
```
mosesn: prefer bytes.length, doesn't create an extra allocation

finagle-mysql/src/main/scala/com/twitter/finagle/mysql/CanBeParameter.scala
missingfaktor: @mosesn That seems to give a wrong impression that this is a pure function, which it is not. It's executed purely for its side-effects. And if it's just going to return the same writer at the end, the return doesn't mean much. If a function is side-effecting it should show in its type-signature, don't you think?

finagle-mysql/src/main/scala/com/twitter/finagle/mysql/CanBeParameter.scala
```diff
@@ -0,0 +1,182 @@
...
+  /**
+   * Retrieves the MySQL type code for the given parameter.
+   * */
+  def typeCode(param: A): Short
+
+  def write(writer: BufferWriter, param: A): Unit
+}
+
+object CanBeParameter {
+  implicit val stringCanBeParameter = {
```
mosesn: please add type annotations to public methods (here and elsewhere)

finagle-mysql/src/main/scala/com/twitter/finagle/mysql/Parameter.scala
```diff
@@ -0,0 +1,37 @@
+package com.twitter.finagle.exp.mysql
+
+import com.twitter.finagle.exp.mysql.transport.BufferWriter
+import language.implicitConversions
+
+trait Parameter {
+  type A
+  def value: A
+  def evidence: CanBeParameter[A]
```
vkostyukov: Why do we need this `evidence` here? Can we propagate via closure? Something like this:

```scala
trait Parameter {
  type A
  def value: A
  def writeTo(writer: BufferWritter): Unit
}

object Parameter {
  implicit def wrap[_A](_value: _A)(implicit _evidence: CanBeParameter[_A]): Parameter = new Paramter {
    type A = _A
    def value: A = _value
    def writeTo(writer: BufferWritter): Unit = {
      evidence.write(writer, value)
    }
  }
}
```

My point is that we don't really need to carry that evidance in the parameter.

finagle-mysql/src/main/scala/com/twitter/finagle/mysql/CanBeParameter.scala
missingfaktor: It does, for byte arrays and `Value`.

finagle-mysql/src/main/scala/com/twitter/finagle/mysql/CanBeParameter.scala
missingfaktor: @mosesn Good point. I changed it in many places, but missed this one. (In current Finagle MySQL, `Array#size` is more prevalent than `Array#length`.

finagle-mysql/src/main/scala/com/twitter/finagle/mysql/CanBeParameter.scala
missingfaktor: @mosesn Okay.

finagle-mysql/src/main/scala/com/twitter/finagle/mysql/Parameter.scala
missingfaktor: @vkostyukov The members of `Parameter` will increase with the number of methods in `CanBeParameter`, and all of them will be pure delegators. What's the value?

finagle-mysql/src/main/scala/com/twitter/finagle/mysql/CanBeParameter.scala
mosesn: oh, I see, you need it for disambiguating between TinyBlob, etc.  Is that correct behavior?  Will mysql be OK if you pass it mismatched kinds of blobs?

finagle-mysql/src/main/scala/com/twitter/finagle/mysql/CanBeParameter.scala
```diff
@@ -0,0 +1,182 @@
...
+      }
+      def write(writer: BufferWriter, param: Array[Byte]) = writer.writeLengthCodedBytes(param)
+    }
+  }
+
+  implicit val valueCanBeParameter = {
+    new CanBeParameter[Value] {
+      def sizeOf(param: Value) = param match {
+        case RawValue(_, _, true, b) => Buffer.sizeOfLen(b.length) + b.length
+        case StringValue(s)          => val bytes = s.getBytes(Charset.defaultCharset); Buffer.sizeOfLen(bytes.size) + bytes.size
```
mosesn: prefer bytes.length

finagle-mysql/src/main/scala/com/twitter/finagle/mysql/CanBeParameter.scala
```diff
@@ -0,0 +1,182 @@
...
+        case ByteValue(_)            => 1
+        case ShortValue(_)           => 2
+        case IntValue(_)             => 4
+        case LongValue(_)            => 8
+        case FloatValue(_)           => 4
+        case DoubleValue(_)          => 8
+        case NullValue               => 0
+        case _                       => 0
```
mosesn: why provide a fallback case here but not in typeCode?

finagle-mysql/src/main/scala/com/twitter/finagle/mysql/CanBeParameter.scala
```diff
@@ -0,0 +1,182 @@
...
+        case RawValue(_, _, true, bytes) => writer.writeLengthCodedBytes(bytes)
+        // allows for Value types as params to prepared statements
+        case ByteValue(b)                => writer.writeByte(b)
+        case ShortValue(s)               => writer.writeShort(s)
+        case IntValue(i)                 => writer.writeInt(i)
+        case LongValue(l)                => writer.writeLong(l)
+        case FloatValue(f)               => writer.writeFloat(f)
+        case DoubleValue(d)              => writer.writeDouble(d)
+        case StringValue(s)              => writer.writeLengthCodedString(s)
+        case _                           => ()
```
mosesn: is silence the correct behavior here?

what about nulls?

finagle-mysql/src/main/scala/com/twitter/finagle/mysql/CanBeParameter.scala
```diff
@@ -0,0 +1,182 @@
+package com.twitter.finagle.exp.mysql
+
+import com.twitter.finagle.exp.mysql.transport.{Buffer, BufferWriter}
+
+trait CanBeParameter[-A] { outer =>
```
mosesn: why do you want this variance annotation?

finagle-mysql/src/main/scala/com/twitter/finagle/mysql/Client.scala
```diff
@@ -22,7 +22,7 @@ object Client {
       }
 
     def prepare(sql: String): PreparedStatement = new PreparedStatement {
-      def apply(ps: Any*): Future[Result] = factory() flatMap { svc =>
+      def apply(ps: Parameter*): Future[Result] = factory() flatMap { svc =>
```
mosesn: prefer factory().flatMap

finagle-mysql/src/main/scala/com/twitter/finagle/mysql/Parameter.scala
```diff
@@ -0,0 +1,37 @@
+package com.twitter.finagle.exp.mysql
+
+import com.twitter.finagle.exp.mysql.transport.BufferWriter
+import language.implicitConversions
+
+trait Parameter {
+  type A
```
mosesn: can we make this a generic instead?

finagle-mysql/src/main/scala/com/twitter/finagle/mysql/Parameter.scala
```diff
@@ -0,0 +1,37 @@
...
+  def value: A
+  def evidence: CanBeParameter[A]
+
+  final def writeTo(writer: BufferWriter): Unit = {
```
mosesn: can we return a BufferWriter instead?

finagle-mysql/src/main/scala/com/twitter/finagle/mysql/CanBeParameter.scala
missingfaktor: @mosesn Okay.

finagle-mysql/src/main/scala/com/twitter/finagle/mysql/Parameter.scala
mosesn: instead of the _ prefixed things, can we rename the variables?

finagle-mysql/src/main/scala/com/twitter/finagle/mysql/CanBeParameter.scala
missingfaktor: @mosesn `Value` is a sealed trait and `typeCode` covers all of the cases explicitly. There is no case that's left out. In case of `sizeOf` I am just keeping the existing behavior as-is.

finagle-mysql/src/main/scala/com/twitter/finagle/mysql/Client.scala
missingfaktor: @mosesn It's been that way since before.

finagle-mysql/src/main/scala/com/twitter/finagle/mysql/CanBeParameter.scala
missingfaktor: @mosesn I have tried to mirror the existing behavior as faithfully as possible. I think other values are skipped in current implementation too. As for `null`s, we are not handling them here. Should they be handled here? Handling `null`s everywhere on a case-by-case basis seems messy, and something to be avoided.

finagle-mysql/src/main/scala/com/twitter/finagle/mysql/CanBeParameter.scala
missingfaktor: @mosesn If I have a custom data type `X` such that `X <: Y`, and there's evidence `CanBeParameter[Y]` available, I would want to be able to pass `X`. The fact that it typechecks means such a usage is valid. Why not have the annotation?

finagle-mysql/src/main/scala/com/twitter/finagle/mysql/Parameter.scala
missingfaktor: @mosesn It wouldn't make sense. It's supposed to be used as an existential type.

finagle-mysql/src/main/scala/com/twitter/finagle/mysql/Parameter.scala
missingfaktor: @mosesn Please refer to my reply to a similar comment before.

finagle-mysql/src/main/scala/com/twitter/finagle/mysql/Parameter.scala
vkostyukov: My opinion is that we probably want to keep the API surface as small as possible. And also, once you have an instance of `Parameter` you don't need an evidence that it can be parameter, since it's already it :)

As far as I know it's a common practice do not store evidences but wrap them into a closure. Although, it's probably just a nitpick. Feel free to ignore.

mosesn:
From a high level, I'm not sure how I feel about this approach.  We want to keep our APIs java friendly in general, and implicits are typically not the easiest way to do it.  With that said, it's definitely a cool API in scala.

Maybe the right thing is to separate this out into a separate API for scala, but not get rid of the old java API?  This also has the benefit of not breaking API/ABI compatibility.

I'm also not sure I'm in love with the Parameter API.  I might be misunderstanding, but it seems like the path to upgrade from Any => Parameter is to import the Parameter.wrap method into scope, and then hope that clients used Seq literals?  That seems awfully fragile, and necessary implicit methods don't make for the easiest to use or understand code.

finagle-mysql/src/main/scala/com/twitter/finagle/mysql/Client.scala
mosesn: right, the style has changed.  since you're changing the line, could you please fix it?  we're changing the style lazily, so we ask that folks fix style as they change code.

finagle-mysql/src/main/scala/com/twitter/finagle/mysql/CanBeParameter.scala
mosesn: Almost all of the parameter types are very specific types that can't be subclasses, like Int.  What is it useful for?

finagle-mysql/src/main/scala/com/twitter/finagle/mysql/CanBeParameter.scala
mosesn: I meant NullParameters, but yes, you're right that it looks like this is how they're done elsewhere too.  Strange!

finagle-mysql/src/main/scala/com/twitter/finagle/mysql/CanBeParameter.scala
mosesn: prefer consistency over existing behavior imo

finagle-mysql/src/main/scala/com/twitter/finagle/mysql/CanBeParameter.scala
mosesn: I think consistency is better here.  If we're going to change the `write*` style methods, let's do them all at once.  Having some of them which follow one convention, and others that follow another is a bummer.  Also, this matches the way that the `java.nio.Buffer` methods work, so it's not unprecedented.

finagle-mysql/src/main/scala/com/twitter/finagle/mysql/Request.scala
mosesn: maybe just keep a `private[this] val sanitized = ....` on the case class?

finagle-mysql/src/main/scala/com/twitter/finagle/mysql/Request.scala
mosesn: yes, this is cleaner imo.

finagle-mysql/src/main/scala/com/twitter/finagle/mysql/CanBeParameter.scala
roanta: I'm not sure we should have the variance either. Even if we have a X <: Y relationship, we should still be forced to think about how that serializes into a Parameter (e.g. sizeOf, typeCode, etc. are likely to be different).

roanta:
Thanks for the PR! This is very cool!

@mosesn I haven't tried it out, but I'm under the impression that this change should be transparent and the translation from Any => Parameter should "just work". At least the changes to the unit tests suggest this.

re: java API. The original API isn't very Java friendly to begin with. We could always offer a parallel java API that passes the implicits explicitly. I think the type safety is definitely worth the compromise.

mosesn:
I definitely agree typesafety is worth it, and I definitely want this PR to make it in, but I'm not sure if this is its final form or if it has some more evolving to do.  With that said, I also want to make sure we have a good java compat story, so if we're dropping the old API, we should at least add some java compatibility tests.

finagle-mysql/src/main/scala/com/twitter/finagle/mysql/Request.scala
missingfaktor: @mosesn That will expose unsanitized params in the (auto-generated) `unapply`.

finagle-mysql/src/main/scala/com/twitter/finagle/mysql/CanBeParameter.scala
missingfaktor: Okay, agreed. I will change this.

finagle-mysql/src/main/scala/com/twitter/finagle/mysql/CanBeParameter.scala
missingfaktor: @mosesn Sorry, I am out of my depth here. I have tried to retain the existing behavior here as well.

finagle-mysql/src/main/scala/com/twitter/finagle/mysql/CanBeParameter.scala
missingfaktor: Okay.

finagle-mysql/src/main/scala/com/twitter/finagle/mysql/CanBeParameter.scala
missingfaktor: Okay. I will try making it more explicit at the least, what cases specifically are skipped. :)

finagle-mysql/src/main/scala/com/twitter/finagle/mysql/CanBeParameter.scala
missingfaktor: @mosesn I was thinking this could be useful for custom data types. But @roanta here has raised some very valid points. I will make the type-class invariant.

finagle-mysql/src/main/scala/com/twitter/finagle/mysql/Client.scala
missingfaktor: Okay. :)

missingfaktor:
@mosesn @travisbrown @vkostyukov @roanta Thanks a lot for the reviews, guys! I very much appreciate you giving your time for this.

@mosesn What @roanta said. The changes should be transparent, and would "just work". Due to how implicit lookup works in Scala, even additional imports would not be required. Example:

```scala
> finagle-mysql/console
[info] Starting scala interpreter...
[info]
Welcome to Scala version 2.10.4 (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_31).
Type in expressions to have them evaluated.
Type :help for more information.

scala> import com.twitter.finagle.exp.mysql._
import com.twitter.finagle.exp.mysql._

scala> def foo(args: Parameter*) = args
foo: (args: com.twitter.finagle.exp.mysql.Parameter*)Seq[com.twitter.finagle.exp.mysql.Parameter]

scala> foo(3, "hello")
res0: Seq[com.twitter.finagle.exp.mysql.Parameter] = WrappedArray(com.twitter.finagle.exp.mysql.Parameter$$anon$1@152f97f0, com.twitter.finagle.exp.mysql.Parameter$$anon$1@3966c6d0)

scala> foo(3, "hello", 'unsupported_type)
<console>:12: error: could not find implicit value for parameter evidence0: com.twitter.finagle.exp.mysql.CanBeParameter[Symbol]
              foo(3, "hello", 'unsupported_type)
                              ^
```

As for Java API, I again have to agree with @roanta. We could have a parallel Java API to make this work, if a Java compat is desirable.

mosesn:
Sorry, where exactly does it find the implicit here?  I don't understand.  Also, that doesn't address my concern is that since the API we're dealing with is Seq[Any], if someone has already created a Seq[Any], and then tries to pass it in separately, it will break.

vkostyukov:
It search in a companion object `Parameter`.  

Re `Seq[Any]`: It's a good concern. It seems to me that it's not possible to convert `Seq[Any]` into `Seq[Parameter]` since the type information is lost. Although, It might be fixed by replacing `Seq[Any]` with `Seq[Parameter]` in user's code: `val seq: Seq[Parameter] = Seq(1, "foo", bool)`.

mosesn:
OK, so to recap: the path to upgrade from Any => Parameter is to make sure the Parameter.wrap method is in scope, which will then update any literal Seqs automatically.  If anyone isn't using a literal Seq (ie even `val x: Seq[Any] = Seq(3, 4, 5, "OK")` is no good) then they will need to make code changes to do this.  If they pass those arguments through a few places, they will have to make multiple code changes.

If we're forcing folks to change their code anyway, maybe we can do better?  My impression is that prepared statements require a specific set of parameters, which have types that are known in advance.  Can we require that they are the right types, beyond just that they are parameter types?

missingfaktor:
> the path to upgrade from Any => Parameter is to make sure the Parameter.wrap method is in scope

No. Scala looks in the companion object of `Parameter` for this conversion automatically, and one has to do nothing to bring the implicit conversion in scope.

> If anyone isn't using a literal Seq (ie even val x: Seq[Any] = Seq(3, 4, 5, "OK") is no good) then they will need to make code changes to do this.

Yes, though the code change should be fairly minimal.

> My impression is that prepared statements require a specific set of parameters, which have types that are known in advance. Can we require that they are the right types, beyond just that they are parameter types?

I am not sure what you mean. That's exactly what this PR is trying to do: To restrict that only things that can be used a parameter, are. `CanBeParameter` and `Parameter` is the machinery used to do it.

mosesn:
OK, I understand now.  Even if we don't import Parameter explicitly, because it's the type that we're trying to convert to, scala tries to find an implicit conversion in its companion object.  Very neat!

```scala
scala> import com.twitter.finagle.exp.mysql.ExecuteRequest
import com.twitter.finagle.exp.mysql.ExecuteRequest

scala> ExecuteRequest(3, IndexedSeq(3, "hello"))
res0: com.twitter.finagle.exp.mysql.ExecuteRequest = com.twitter.finagle.exp.mysql.ExecuteRequest@6a734dd1
```

With that said, I still think we can do better.  When we make a prepared statement, we know that we want to fill in the gaps in the prepared statements with things of a precise type.  For example, we might know that we need a String, an Int, and a Boolean.  In the proposed Parameters api, we can enforce that we get valid mysql Parameters, but not the number, the specific type, or the order.

Could we do something like:

```scala
object TupledParameters[A] {
  implicit def 1tuple[A : Parameter](tup: (A)): TupledParameters[(A)] = ???
  implicit def 2tuple[A : Parameter, B : Parameter](tup: (A, B)): TupledParameters[(A, B)] = ???
}

class ExecuteRequest[A : TupledParameters](stmtId: Int, params: A,...) { ... }

trait PreparedStatement {
  def apply[A : TupledParameters](params: A): Future[Result]
}
```

?

finagle-mysql/src/main/scala/com/twitter/finagle/mysql/CanBeParameter.scala
roanta: It probably wasn't the best idea to derive the blob type from the size of the array, but we should revisit this separately. I think retaining current behavior is a good idea!

@mosesn The blobs are length encoded, but they still need to match your schema. I suspect mysql will complain otherwise.

finagle-mysql/src/main/scala/com/twitter/finagle/mysql/Parameter.scala
```diff
@@ -0,0 +1,37 @@
+package com.twitter.finagle.exp.mysql
+
+import com.twitter.finagle.exp.mysql.transport.BufferWriter
+import language.implicitConversions
+
```
roanta: Some scaladoc that briefly explains what Parameter is would be nice. Note, that all usage of Parameter should be via CanBeParameter.

finagle-mysql/src/main/scala/com/twitter/finagle/mysql/Parameter.scala
```diff
@@ -0,0 +1,37 @@
...
+trait Parameter {
```
roanta: Would it make sense to seal this? Or is it not possible because we instantiate it anonymously in wrap?

finagle-mysql/src/main/scala/com/twitter/finagle/mysql/CanBeParameter.scala
```diff
@@ -0,0 +1,185 @@
+package com.twitter.finagle.exp.mysql
+
+import com.twitter.finagle.exp.mysql.transport.{Buffer, BufferWriter}
+
+trait CanBeParameter[A] { outer =>
+  /**
+   * Returns the size of the given parameter in its MySQL binary representation.
+   **/
```
roanta: style nit: our scaladoc blocks are closed with */ instead of **/

finagle-mysql/src/main/scala/com/twitter/finagle/mysql/Request.scala
```diff
@@ -258,6 +231,24 @@ case class ExecuteRequest(
     }
 }
 
+object ExecuteRequest {
+  def apply(stmtId: Int,
```
roanta: style nit, drop params to new line:

    def apply(
       stmtId: Int,
       ...
    ): ExecuteRequest = ...

roanta:
Just some minor nits, this looks great to me! I think we should pull it in.

> For example, we might know that we need a String, an Int, and a Boolean

@mosesn That information is only available at runtime. I think this is the best we can do with the info we have. We would need some sort of schema annotation along with code gen to do what your describing at compile time.

missingfaktor:
Hi @mosesn. That would require a change in how you declare queries – You will have to declare the types of parameters and such with the sql statements. I have been toying with this idea in one of our projects. This is what it looks like (The API is inspired by Doobie, but the underlying mechanism is a lot simpler and is less feature-rich):

```scala
val _query: SelectOpt[Project, (String, Int)] = """
  |select * from projects where name = ? and data_sensitivity_level = ?
""".stripMargin.selectOpt[Project, (String, Int)]
```

This would look for `CanBeParameter` evidences for all components of the tuple of parameters. 

As I said, I am still toying with the idea. It's pretty basic in its current shape.

IMO the changes I have implemented as part of this PR are a good first step. The users will need some code change but that should be fairly minimal. Later we can try to improve things even further.

missingfaktor:
@roanta I have addressed your review comments in my latest commit.

missingfaktor:
@roanta @mosesn I had to restore the variance annotation. :( In its absence, for instance, a value of type `CanBeParameter[Value]` cannot be used when a `CanBeParameter[StringValue]` is needed. (Note that `StringValue <: Value`.) This leads to errors like the one shown below: 

```scala
could not find implicit value for parameter evidence0: com.twitter.finagle.exp.mysql.CanBeParameter[com.twitter.finagle.exp.mysql.StringValue]
```

finagle-mysql/src/main/scala/com/twitter/finagle/mysql/Parameter.scala
missingfaktor: @roanta I usually only use `sealed` when declaring ADTs/GADTs, but using it to prevent extension of `Parameter` from outside this unit seems like a great idea. 

Anonymous instantiation is okay as long as it's in the same unit.

mosesn:
@roanta we can do better by declaring in `prepare` what the types we need are, like how @missingfaktor described it.  Then we can enforce it at compile-time, sort of like with string interpolation.

@missingfaktor since we're breaking ABI compat, I'd rather just do it all in one fell swoop.  I don't think it's that much extra work to add the type parameter to the prepare call site in the API, and I'd be willing to do the migration for twitter.  I think it'll be easier to only have to do one migration rather than two (although it would definitely be a bigger migration than just the one that's proposed right now).

Oh yeah, the Value is a good use of the variance.  Good call.

Anyway, I think this review looks fine, but I have a strong preference for the CanBeParameter tuples.  I'm not going to block this review though.

roanta:
I see, annotating at the `prepare` call site may be enough. I agree with @missingfaktor that we should pursue this incrementally and this is a good first step.

@missingfaktor re: the variance annotation. Can we instead expand the CanBeParameter[Value] into separate instance of CanBeParameter? I really think this is more correct (both for `Value` and other parameter types moving forward).

alexflav23:
@mosesn @missingfaktor I realise this is somewhat off topic, yet I think you may be interested in knowing we have a fully type safe reactive DSL built atop Finagle for MySQL. Have a look here: https://github.com/websudos/morpheus/blob/develop/morpheus-mysql/src/test/scala/com/websudos/morpheus/mysql/query/MySQLUpdateQueryTest.scala.

It's not yet production ready, but we'd welcome feedback and contributions.

missingfaktor:
@roanta Splitting `CanBeParameter[Value]` into separate instances doesn't seem right, because `Value` is the type constructor here, and `StringValue`, `LongValue` etc are simply data constructors. It just happens to be the case that Scala implements ADTs with subtyping, so `StringValue` etc also become types on their own.

Also consider this: When we add a support for `Option[_]` in future (which I hope we do), we will want the following to just work: 

```scala
client.prepare("...stuff...").apply(2, 4, Some("howdy"))
```

missingfaktor:
> @mosesn @missingfaktor I realise this is somewhat off topic, yet I think you may be interested in knowing we have a fully type safe reactive DSL built atop Finagle for MySQL. Have a look here: https://github.com/websudos/morpheus/blob/develop/morpheus-mysql/src/test/scala/com/websudos/morpheus/mysql/query/MySQLUpdateQueryTest.scala. It's not yet production ready, but we'd welcome feedback and contributions.

@alexflav23 Hi Alex. That seems to take a good deal of inspiration from Rogue and Slick, my two favorite database libraries. Looks pretty great! Thanks for bringing it to notice.

roanta:
@missingfaktor I don't see any good reason to hide the representation of `Value`. It's not really an ADT; there's nothing abstract about it. We can't do anything interesting without exposing it's representations. But maybe I'm missing something.

Support Option[_] seems like a good idea, but are you sure that -A will give us what we want?

    scala> trait CanBeParameter[-A]
    defined trait CanBeParameter
    
    scala> def wrap[A: CanBeParameter](value: A) = ???
    wrap: [A](value: A)(implicit evidence$1: CanBeParameter[A])Nothing

    scala> implicit val stringCanBe = new CanBeParameter[String] {}
    stringCanBe: CanBeParameter[String] = $anon$1@4f913b29

    scala> wrap("hello")
    scala.NotImplementedError: an implementation is missing
    ...

    scala> wrap(Some("hello"))
    <console>:11: error: could not find implicit value for evidence parameter of type   
    CanBeParameter[Some[String]]
              wrap(Some("hello"))

missingfaktor:
@roanta I meant algebraic data type, not abstract data type. Sorry for the confusion.

Some REPL scribbles to clarify my point about `Option[_]`:

```scala
scala> trait CBP1[A]
defined trait CBP1

scala> implicit val forString = new CBP1[String] {}
forString: CBP1[String] = $anon$1@4d591d15

scala> implicit def forOpt[A : CBP1] = new CBP1[Option[A]] {}
forOpt: [A](implicit evidence$1: CBP1[A])CBP1[Option[A]]

scala> def wrap[A : CBP1](x: A) = ???
wrap: [A](x: A)(implicit evidence$1: CBP1[A])Nothing

scala> wrap(Some("hello"))
<console>:12: error: could not find implicit value for evidence parameter of type CBP1[Some[String]]
              wrap(Some("hello"))
                  ^

scala> trait CBP2[-A]
defined trait CBP2

scala> implicit val forString2 = new CBP2[String] {}
forString2: CBP2[String] = $anon$1@3b0143d3

scala> implicit def forOpt2[A : CBP1] = new CBP2[Option[A]] {}
forOpt2: [A](implicit evidence$1: CBP1[A])CBP2[Option[A]]

scala> def wrap2[A : CBP2](x: A) = ???
wrap2: [A](x: A)(implicit evidence$1: CBP2[A])Nothing

scala> wrap2(Some("hello"))
scala.NotImplementedError: an implementation is missing
  at scala.Predef$.$qmark$qmark$qmark(Predef.scala:225)
  at .wrap2(<console>:13)
  ... 33 elided
```

roanta:
I see, that's neat! Unless anyone else has anymore comments, I'll work to pull this in this week.

missingfaktor:
Thank you again @roanta, @mosesn, and others for the reviews! :)

finagle-mysql/src/main/scala/com/twitter/finagle/mysql/Parameter.scala
```diff
@@ -0,0 +1,41 @@
...
+  final def writeTo(writer: BufferWriter): Unit = {
+    evidence.write(writer, value)
+  }
+
+  final def size: Int = evidence.sizeOf(value)
+  final def typeCode: Short = evidence.typeCode(value)
+}
+
+object Parameter {
+  implicit def wrap[A0](value0: A0)(implicit evidence0: CanBeParameter[A0]): Parameter = {
```
luciferous: Any way to make this `implicit class` and drop the implicitConversion import?

missingfaktor: There are two things here that make that not possible: A conditional inside the `wrap`, and a type member. `implicit class`es can be substituted for implicit conversions in most cases, but this isn't one of them.

luciferous: What I mean is:

```scala
implicit class Wrap[A0](val value0: A0) extends AnyVal {
  def asParam(implicit evidence0: CanBeParameter[A0]): Parameter =
    if (value0 == null) {
      NullParameter
    } else {
      new Parameter {
        type A = A0
        def value: A = value0
        def evidence: CanBeParameter[A] = evidence0
      }
    }
}
```

missingfaktor: I see. With this, though, the client is burdened with code like following:

```scala
client.prepare("...").apply(3.asParam, "jd".asParam)
```

I think it adds no value and hampers readability.

missingfaktor: Thanks @luciferous.

luciferous: @missingfaktor Thank _you_. This is good work.

roanta:
We merged this internally, it should show up in the `develop` branch shortly!

Note: We added  `def unsafeWrap(value: Any): Parameter` to the Parameter companion object to ease migration for some of our services. They had rich wrappers around finagle-mysql that required changing their types all the way through. I imagine this might help others with the migration too.

missingfaktor:
@roanta Yay! Thanks for the merge. :fireworks: Any estimate when this might be a part of official release?

How does `unsafeWrap` work? Does it have a registry of all the `CanBeParameter` instances?

roanta:
Unfortunately, we just missed the 6.25 release. I don't have an exact ETA when we will do the next release, but we're trying to release more frequently.

re: `unsafeWrap` it just pattern matches on the `value` and preserves the previous behavior where unknown params are logged and written as sql null.

PR #365, opened 2015-04-10
Expose a means to configure MaskCancelFilter in the Finagle 6 APIs

evnm:
Problem

The "cancel on hangup" feature afforded by `c.t.f.filter.MaskCancelFilter`
is not configurable with the Finagle 6 APIs. This represents a lack of
parity with ServerBuilder, which offers a `cancelOnHangup()` function.

Solution

Make the `MaskCancelFilter` Param/module trimmings public.

### DIFF
--- a/finagle-core/src/main/scala/com/twitter/finagle/filter/MaskCancelFilter.scala
+++ b/finagle-core/src/main/scala/com/twitter/finagle/filter/MaskCancelFilter.scala
@@ -8,8 +8,8 @@ object MaskCancelFilter {
 
   // TODO: we should simply transform the stack for boolean
   // stackables like this.
-  private[finagle] case class Param(yesOrNo: Boolean)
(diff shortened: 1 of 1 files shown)


tonyd3:
+1 
I'm having problems with enabling MaskCancelFilter using the new API. 
MaskCancelFilter.Param is inaccessible, so the following example fails.
 val service = Server()
  .configured(Label("thrift"))
  .configured(MaskCancelFilter.Param(true))
  .serveIface("0.0.0.0:" + port, serviceClass)

kevinoliver:
Thanks! This was fixed on the `develop` branch — https://github.com/twitter/finagle/blob/develop/finagle-core/src/main/scala/com/twitter/finagle/filter/MaskCancelFilter.scala

Hopefully we will do a release in the next couple weeks.

evnm:
Thanks, Kevin.

Issue #366, opened 2015-04-11
LoadService not support GBK charset


tianxiao-ma:
when LoadService encounter a file named `META-INF/services/upload.xsd` in a jar, it throw a `java.nio.charsetMalformedInputException` exception. 

Because my linux environment's default charset is GBK, so LoadService use GBK to read files in jar, but the `META-INF/services/upload.xsd` file is encoded with `UTF-8`, so a exception is thrown.

Can let LoadService use `ISO8859-1` as its default charset to read files? or can let the user determine which charset should use?

mosesn:
Could you paste the stack trace?  I think the bug is that https://github.com/twitter/finagle/blob/master/finagle-core/src/main/scala/com/twitter/finagle/util/LoadService.scala#L115 should have a second argument, "UTF-8", but a stack trace would help nail it down.

https://github.com/twitter/finagle/blob/master/finagle-core/src/main/scala/com/twitter/finagle/util/LoadService.scala#L180 is a possible candidate too.

tianxiao-ma:
@mosesn 

sorry, I'm late.

yes, the problem is the second parameter. And [https://github.com/twitter/finagle/blob/master/finagle-core/src/main/scala/com/twitter/finagle/util/LoadService.scala#L93](https://github.com/twitter/finagle/blob/master/finagle-core/src/main/scala/com/twitter/finagle/util/LoadService.scala#L93) is hard code 'UTF-8'.

I think `ISO8859-1` is a better default charset for `LoadService`.

my stack track is here.
![stack](https://cloud.githubusercontent.com/assets/3139359/7134534/4d0630c4-e2d0-11e4-8ef7-ee3f2fbffa55.png)

The `data-resolve.xsd` file is encoded use `UTF-8`, and the linux server's default charset I run `finagle` is `GBK`, so the exception is throwed.

mosesn:
Could you elaborate on why you think ISO8859-1 is a better choice?  And how do you feel about making a PR fixing this?

tianxiao-ma:
@mosesn 
It's about how `ISO8859-1` do encode and decode. For short,  `ISO8859-1` is single-byte fixed length encoding, and `GBK` or `UTF-8` is variable-length encoding. For detail, please see [http://stackoverflow.com/questions/7048745/what-is-the-difference-between-utf-8-and-iso-8859-1](http://stackoverflow.com/questions/7048745/what-is-the-difference-between-utf-8-and-iso-8859-1)

[...]

PR #367, opened 2015-04-15
enforce LoadService use UTF-8 to read files

tianxiao-ma:
`LoadService` use `ISO8859-1` to read file.

Before this commit, the `LoadService` use `UTF-8` or system default charset to read files, when the file is encoded by `UTF-8` and the system default charset is `GBK`, or the revert case, `LoadService` will throw exception. Please see issue #366

### DIFF
--- a/finagle-core/src/main/scala/com/twitter/finagle/util/LoadService.scala
+++ b/finagle-core/src/main/scala/com/twitter/finagle/util/LoadService.scala
@@ -90,7 +90,7 @@ private object ClassPath {
       if (f.isDirectory && f.canRead)
         browseDir(f, loader, prefix + f.getName + "/", buf)
       else for (iface <- ifaceOfName(prefix + f.getName)) {
-        val source = Source.fromFile(f, "ISO8859-1")
+        val source = Source.fromFile(f, "UTF-8")
         val lines = readLines(source)
         buf += Info(prefix + f.getName, iface, lines)
       }
@@ -112,7 +112,7 @@ private object ClassPath {
         if !(ignoredPackages exists (n startsWith _))
         iface <- ifaceOfName(n)
       } {
-        val source = Source.fromInputStream(jarFile.getInputStream(e), "ISO8859-1")
+        val source = Source.fromInputStream(jarFile.getInputStream(e), "UTF-8")
         val lines = readLines(source)
         buf += Info(n, iface, lines)
       }
@@ -177,7 +177,7 @@ object LoadService {
 
     val classNamesFromRessources = for {
       rsc <- loader.getResources("META-INF/services/" + ifaceName).asScala
-      line <- ClassPath.readLines(Source.fromURL(rsc, "ISO8859-1"))
+      line <- ClassPath.readLines(Source.fromURL(rsc, "UTF-8"))
     } yield line
 
     (classNames ++ classNamesFromRessources).distinct flatMap { className =>
--- a/finagle-core/src/main/scala/com/twitter/finagle/util/LoadService.scala
+++ b/finagle-core/src/main/scala/com/twitter/finagle/util/LoadService.scala
@@ -90,7 +90,7 @@ private object ClassPath {
       if (f.isDirectory && f.canRead)
         browseDir(f, loader, prefix + f.getName + "/", buf)
       else for (iface <- ifaceOfName(prefix + f.getName)) {
-        val source = Source.fromFile(f, "UTF-8")
+        val source = Source.fromFile(f, "ISO8859-1")
         val lines = readLines(source)
         buf += Info(prefix + f.getName, iface, lines)
       }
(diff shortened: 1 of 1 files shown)


mosesn:
I don't think this will work for variable-length UTF-8 characters.  I don't think it will work for GBK when a GBK character is two bytes either.  I think the simplest way forward is to just keep it UTF-8, and require that all of the files that finagle uses be in UTF-8.

What do you think?

tianxiao-ma:
@mosesn 
`require that all of the files that finagle uses be in UTF-8` is impossible. My project include some jar which is created by other people, and in these jar, some file will read by finagle's LoadService, like the file I list in #366 .

`ISO8859-1` worked, I have try it in my project.

mosesn:
@tianxiao-ma the overlap between `ISO8859-1` 00 => 7F, is ASCII.  `UTF-8` also encodes 00 => 7F the same way.  As far as I can tell, `GBK` 00 => 7F is not ASCII, so `ISO8859-1` is inappropriate, unless `UTF-8` is also appropriate.

Looking at #366 it looks like the file you're trying to read is in `UTF-8`.  Could you elaborate on your problem?  I'm not sure that I understand it.

tianxiao-ma:
@mosesn 
Yes,  the file I list in #366 is encoded as `UTF-8`, but the charset `LoadService` used to read this file is `GBK`, so the exception throws. This is because in current implementation, `LoadService` will use system's default charset to read files, and the system's default charset is `GBK`.

What is my require is to have a way change the charset used by `LoadService`, but in finagle's current version, there is no such way. 

### About the `ISO8859-1` and `UTF-8`
The `ISO8859-1` use a one byte fixed-length scheme to encode and decode text, so it can work for any file, since the minimal unit for a file is `byte`.

But `UTF-8` or `GBK` will use different algorithm to encode and decode text, so the file encoded by `UTF-8` can't be decoded by `GBK`. For example, the chinese word `中国`, `UTF-8` may encoded it into 4 bytes, but `GBK` my encoded it into 5 bytes. So, if we use `GBK` to read a file it a file which is encoded as `UTF-8`, a exception will be throw.

mosesn:
Right, but just because an exception doesn't throw doesn't mean it will be correct or meaningful.

NB: java also requires that these files be in `UTF-8`.

> A service provider is identified by placing a provider-configuration file in the resource directory META-INF/services. The file's name is the fully-qualified binary name of the service's type. The file contains a list of fully-qualified binary names of concrete provider classes, one per line. Space and tab characters surrounding each name, as well as blank lines, are ignored. The comment character is '#' ('\u0023', NUMBER SIGN); on each line all characters following the first comment character are ignored. The file must be encoded in UTF-8.



https://docs.oracle.com/javase/7/docs/api/java/util/ServiceLoader.html

tianxiao-ma:
So,  Can you let LoadService use 'UTF-8' to read these file, not the system's default charset?

mosesn:
Yes!  I think your change is great, except that we should use `UTF-8` instead of `ISO8859-1`.

tianxiao-ma:
I think it's need to let developer has a method to set the charset used by `LoadService`

mosesn:
@tianxiao-ma Could you elaborate?  Finagle exports files in `UTF-8` and java requires that they be in `UTF-8` too.  If you put them in something that's not `UTF-8` then you're breaking the contract with java and finagle.

tianxiao-ma:
@mosesn 
Use `UTF-8` is right, but in current version of `LoadService`, it will use system default charset to read files. And the system's default charset might not be `UTF-8`.

I think you should enforce  `LoadService` use `UTF-8` to read files, like [https://github.com/twitter/finagle/blob/master/finagle-core/src/main/scala/com/twitter/finagle/util/LoadService.scala#L93](https://github.com/twitter/finagle/blob/master/finagle-core/src/main/scala/com/twitter/finagle/util/LoadService.scala#L93)

mosesn:
I absolutely agree!  I think you should change this pull request to use UTF-8 instead of ISO8859-1.  How does that sound?

tianxiao-ma:
@mosesn 
I have update the pull request, please check it

mosesn:
LGTM

finagle-core/src/main/scala/com/twitter/finagle/util/LoadService.scala
```diff
@@ -177,7 +177,7 @@ object LoadService {
 
     val classNamesFromRessources = for {
       rsc <- loader.getResources("META-INF/services/" + ifaceName).asScala
-      line <- ClassPath.readLines(Source.fromURL(rsc))
+      line <- ClassPath.readLines(Source.fromURL(rsc, "UTF-8"))
```
kevinoliver: same.

kevinoliver:
@tianxiao-ma after the small requested change, i'm happy to merge this in.

dschobel:
LGTM

tianxiao-ma:
my first successful pull request, thank you all. @mosesn @kevinoliver @dschobel

finagle-core/src/main/scala/com/twitter/finagle/util/LoadService.scala
tianxiao-ma: @kevinoliver can you explain more? how to change?

kevinoliver: apologies. didn't realize the Source api doesn't expose that.

i'll try to get this merged in soon.

kevinoliver:
Ok this is now merged locally and should make it to the develop branch soon. Thanks @tianxiao-ma.

tianxiao-ma:
^_^

PR #368, opened 2015-04-16
fingle-mysql: transaction support

fwbrasil:
**Problem**

Currently it is not possible to run multiple statements using a 
single transaction.

**Solution**

Add the `transactional` method that executes a function within a transaction. 
The function's input is a client that has only one underlying database 
connection with an open transaction. The provided client instance 
must be used for all database interactions and never leak to outside 
the transactional block scope.

The transaction is automatically committed in case of a success and 
automatically rolled back in case of a failure.

**Result**

Users are able to run multiple statements using a single transaction.

### DIFF
--- a/finagle-mysql/src/main/scala/com/twitter/finagle/mysql/Client.scala
+++ b/finagle-mysql/src/main/scala/com/twitter/finagle/mysql/Client.scala
@@ -2,7 +2,7 @@ package com.twitter.finagle.exp.mysql
 
 import com.twitter.finagle.ServiceFactory
 import com.twitter.finagle.stats.{NullStatsReceiver, StatsReceiver}
-import com.twitter.util.{Closable, Future, Time}
+import com.twitter.util.{Closable, Future, Time, Throw, Return}
 import java.util.logging.Level
 
 object Client {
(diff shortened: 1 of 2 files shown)


fwbrasil:
cc/ @rodolfo42 @missingfaktor @folone

mosesn:
@fwbrasil I think we already added support for transactions in the `develop` [branch](https://github.com/twitter/finagle/blob/develop/finagle-mysql/src/main/scala/com/twitter/finagle/mysql/Client.scala#L73).

Does that look like it takes care of your concerns?

finagle-mysql/src/main/scala/com/twitter/finagle/mysql/Client.scala
```diff
@@ -33,6 +33,23 @@ object Client {
       }
     }
 
+    def transactional[R](f: Client => Future[R]): Future[R] = 
+      factory().flatMap { svc =>
+        val client = apply(ServiceFactory.const(svc))
```
roanta: As @mosesn mentioned, we recently added this internally and it's available as part of the develop branch.

Note though, this may have surprising results. Because of how the mysql client operates on the this `ServiceFactory` you aren't guaranteed exclusivity to this `Service` (i.e. connection in this case). You need to override the `Service#close` method to completely control its lifecycle. Take a look at how we implemented it in the `develop` branch.

fwbrasil:
@mosesn @roanta Thanks for the feedback, I'll take a look at the `develop` branch. 

It is interesting to see that you are working on improving the `finagle-mysql` module, it has been difficult to use it in our projects. Out of curiosity: do you consider it production ready?

roanta:
I would say it's production ready; we have teams using it in production at twitter.

What kind of difficulties have you guys run into? Is it mostly related to the bare API?

PR #369, opened 2015-04-16
 finagle-core: make ApertureLoadBandBalancer public

tonyd3:
Problem
I want to use the Aperture LoadBalancerFactory with custom values.
The class is private, so we are unable to create a new object that instantiates the ApertureLoadBandBalancer class.
I also want to experiment with creating load balancers with the Aperture trait mixed in, however I am unable to do that because it's private as well.
    
Solution
Make both the class and the trait public.
    
Result
You should be able to instantiate ApertureLoadBandBalancer and also create a custom load balancer with the Aperture trait mixed in.


tonyd3:
created against the wrong branch.

PR #370, opened 2015-04-16
finagle-core: make ApertureLoadBandBalancer public

tonyd3:
Problem
I want to use the Aperture LoadBalancerFactory with custom values.
The class is private, so we are unable to create a new object that instantiates the ApertureLoadBandBalancer class.
I also want to experiment with creating load balancers with the Aperture trait mixed in, however I am unable to do that because it's private as well.

Solution
Make both the class and the trait public.

Result
You should be able to instantiate ApertureLoadBandBalancer and also create a custom load balancer with the Aperture trait mixed in.

### DIFF
--- a/finagle-core/src/main/scala/com/twitter/finagle/loadbalancer/Aperture.scala
+++ b/finagle-core/src/main/scala/com/twitter/finagle/loadbalancer/Aperture.scala
@@ -67,7 +67,6 @@ object ApertureBalancerFactory extends WeightedLoadBalancerFactory {
 }
 
 
-
 /**
  * The aperture load-band balancer balances load to the smallest
  * subset ("aperture") of services so that:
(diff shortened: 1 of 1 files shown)


mosesn:
LGTM, but I'm curious about what the ApertureBalancerFactory doesn't expose that you want to twiddle.

tonyd3:
In this case, just max effort.

mosesn:
What do you mean by max effort?

luciferous:
@mosesn https://github.com/twitter/finagle/pull/370/files#diff-427d6c5f1956c47f025da1ba514f27d4R110

luciferous:
@roanta thoughts

roanta:
I think this should be configurable via the `Balancers` object now.

https://github.com/twitter/finagle/blob/develop/finagle-core/src/main/scala/com/twitter/finagle/loadbalancer/Balancers.scala#L117

luciferous:
@roanta That seems correct.

Issue #376, opened 2015-05-05
Document Name.Bound.path


olix0r:
I'm looking at the `Name` implementation; and I can't really recall why we added the [`path` field to `Name.Bound`](https://github.com/twitter/finagle/blob/develop/finagle-core/src/main/scala/com/twitter/finagle/Name.scala#L36-L47).

We should add some documentation explaining the role of this field.  I'm happy to do it, except I don't really know what to say ;)

olix0r:
Specifically: is this a 'residual' path or is this something else?

If I recall correctly, this path should only consist of the unbound residual path components (and not the bound path), while the id field (if it is a path) should identify the bound path.

dschobel:
routed to #traffic

jaked:
It's the residual path. according to @olix0r - 162.days:

  https://github.com/twitter/finagle/commit/25a24417

olix0r:
I had a feeling this was my fault.  I'll submit a PR to clarify the docs.  Thanks.

PR #378, opened 2015-05-05
Improve Name.Bound documentation.

olix0r:
Problem

The 'id' and 'path' fields of Name.Bound were insufficiently documented.

Solution
    
Clarify that 'path' should only contain unbound residual path components and that 'id' should only reflect the bound path.


olix0r:
Fixes #376

dschobel:
pulling in, thanks for the docs!

finagle-core/src/main/scala/com/twitter/finagle/Name.scala
```diff
@@ -39,7 +39,11 @@ object Name {
    *
    * Equality of two Names is delegated to `id`. Two Bound instances
    * are equal whenever their `id`s are. `id` identifies the `addr`
-   * and not the `path`.
+   * and not the `path`.  If the `id` is a [[com.twitter.finagle.Name.Path
+   * Path]], it should only contain *bound*--not residual--path components.
```
mosesn: do we ever define what residual means?

mosesn:
For future reference, it would be rad if you included the Problem / Solution / Result in the commit message directly, since we just snarf your commit message in directly with a signed-off-by message.

olix0r:
Sorry, it was in there just without the words 'problem' and 'solution'
(otherwise the content was the same).  Will keep this in mind next time.

On Tue, May 5, 2015 at 12:48 PM, Moses Nakamura <notifications@github.com>
wrote:

> For future reference, it would be rad if you included the Problem /
> Solution / Result in the commit message directly, since we just snarf your
> commit message in directly with a signed-off-by message.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/twitter/finagle/pull/378#issuecomment-99253473>.
>



-- 
Oliver Gould <ver@buoyant.io>

mosesn:
No worries!

finagle-core/src/main/scala/com/twitter/finagle/Name.scala
```diff
@@ -39,7 +39,11 @@ object Name {
...
+   *
+   * The `path` contains unbound residual path components that were not
+   * processed duing name resolution.
```
dschobel: *during

finagle-core/src/main/scala/com/twitter/finagle/Name.scala
mosesn: er, to be more precise, I don't know what a residual path is.  Can we add a description of what a residual path is, and what it might be used for?

olix0r: I was hoping *bound* and *unbound* were the operative words and residual was simply an adjective.  Basically, given a path like `/s/buoyant/proxy/s/svc/app/path/to/resource`, a Namer may only bind a prefix like `/s/buoyant/proxy` and leave the residual path, `/s/svc/app/path/to/resource` to be bound later (i.e. by the proxy).  When the proxy binds the remaining name, it may only bind `/s/svc/app`, leaving `/path/to/resource` to be handled by the application in whatever way it deems fit.

I'll try to figure out the best way to fit this into the docs.

travisbrown:
Is this ready to go except for the typo, @dschobel, @olix0r? If so I can just pull it in and add the fix before internal review.

dschobel:
I'd like it better if we included @olix0r's comment of what a residual path is but I think the posted version is still an improvement for experts so I'm happy to have it land.

olix0r:
sorry i haven't popped stack back to this yet.  i can address this tomorrow

olix0r:
I added a note about why residual paths are even a thing. Let me know if it needs more clarification.

finagle-core/src/main/scala/com/twitter/finagle/Name.scala
```diff
@@ -23,6 +23,11 @@ import com.twitter.finagle.util.Showable
  * (e.g. `Http.newClient(/s/org/servicename)`). These APIs use `Resolver` under
  * the hood to resolve the destination names into the `Name` representation
  * of the appropriate cluster.
+ *
+ * As names are are bound, a [[com.twitter.finagle.Namer Namer]] may
```
dschobel: s/are^2/are/

dschobel:
lgtm modulo typo

mosesn:
lgtm

luciferous:
Closed by de6c058d25a840098e8cd99e9075609315aa7c6e

PR #380, opened 2015-05-13
add scripting feature to finagle-redis, with tests

timxzl:
add scripting feature to finagle-redis, with a rich set of testcases
supported commands: EVAL, EVALSHA, SCRIPT EXISTS, SCRIPT FLUSH, SCRIPT LOAD


timxzl:
it seems that I may have used a wrong base repo. will close this one and open a new one.

PR #387, opened 2015-06-19
Enable finagle-swift.

rojanu:
Use facebook swift annotation instead of creating new one.
Fix minor bug fix in swift proxy when no argument is defined.


travisbrown:
I've merged these changes over in the new stand-alone [finagle-swift](https://github.com/finagle/finagle-swift/commit/7d7d16f5745842fafbc532787959cf6017da69f5) project—thanks, @rojanu.

PR #390, opened 2015-07-02
LoadService may fail with StackOverflowError's

thirstycrow:
When some jars in classpath reference each other by the Class-Path manifest header,
the LoadService may result in infinite loop and fail with a StackOverflowError.

Solution

Keep a track of the browsed files, and do nothing if the referenced jar has
already been browsed.

### DIFF
--- a/finagle-core/src/main/scala/com/twitter/finagle/util/LoadService.scala
+++ b/finagle-core/src/main/scala/com/twitter/finagle/util/LoadService.scala
@@ -68,7 +68,13 @@ private object ClassPath {
     ents
   }
 
-  private[finagle] def browseUri(uri: URI, loader: ClassLoader, buf: mutable.Buffer[Info], history: mutable.Set[String] = mutable.Set[String]()) {
+  private[finagle] def browseUri(
+    uri: URI,
+    loader: ClassLoader,
+    buf: mutable.Buffer[Info],
+    history: mutable.Set[String] = mutable.Set[String]()
+  ): Unit = {
+
     if (uri.getScheme != "file")
       return
     val f = new File(uri)
--- a/finagle-core/src/main/scala/com/twitter/finagle/util/LoadService.scala
+++ b/finagle-core/src/main/scala/com/twitter/finagle/util/LoadService.scala
@@ -68,17 +68,22 @@ private object ClassPath {
     ents
   }
 
-  private[finagle] def browseUri(uri: URI, loader: ClassLoader, buf: mutable.Buffer[Info]) {
+  private[finagle] def browseUri(uri: URI, loader: ClassLoader, buf: mutable.Buffer[Info], history: mutable.Set[String] = mutable.Set[String]()) {
     if (uri.getScheme != "file")
       return
     val f = new File(uri)
     if (!(f.exists() && f.canRead))
       return
+    val path = f.getCanonicalPath
+    if (history.contains(path))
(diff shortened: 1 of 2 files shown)


finagle-core/src/main/scala/com/twitter/finagle/util/LoadService.scala
```diff
@@ -68,17 +68,22 @@ private object ClassPath {
     ents
   }
 
-  private[finagle] def browseUri(uri: URI, loader: ClassLoader, buf: mutable.Buffer[Info]) {
+  private[finagle] def browseUri(uri: URI, loader: ClassLoader, buf: mutable.Buffer[Info], history: mutable.Set[String] = mutable.Set[String]()) {
```
mosesn: this line is very long, can we chop it up?

mosesn: let's explicitly return `): Unit = {`

finagle-core/src/main/scala/com/twitter/finagle/util/LoadService.scala
```diff
@@ -68,17 +68,22 @@ private object ClassPath {
...
     if (uri.getScheme != "file")
       return
     val f = new File(uri)
     if (!(f.exists() && f.canRead))
       return
+    val path = f.getCanonicalPath
```
mosesn: instead of turning these into strings, why not just keep them `File`s?  Java supports rich equality / hashCode for them, so they also allow things like checking if two pathnames are the same on case-insensitive filesystems.

mosesn:
Thanks for the contribution!

Can we add a test?  How have you tested this?

finagle-core/src/main/scala/com/twitter/finagle/util/LoadService.scala
thirstycrow: The concerning here is about symbolic link.

thirstycrow:
I made this change as a quick fix for the problem to use finagle clients with resin 4 pro container. It works in our scenario. No enough time for testing code at the moment, but I will do it as soon as possible.

finagle-core/src/main/scala/com/twitter/finagle/util/LoadService.scala
mosesn: Could you elaborate?

thirstycrow: .
├── p -> .
└── test.jar

thirstycrow: Assume we have a test.jar, which references dot/test.jar by the Class-Path manifest attribute, where dot is a symbolic link to the current directory. In this case, test.jar will be processed repeatedly, because test.jar is not equal to dot/test.jar as File objects, and dot/test.jar is not equals to dot/dot/test.jar, ...

There may never be such a case in real systems, but generally speaking, compare files by canonical path is a safer option.

vkostyukov:
@mosesn since we have tests now, can we mark it shipable?

mosesn:
LGTM

vkostyukov:
I'm pulling this to review internally. Will close the PR once it's merged.

vkostyukov:
This has been merged internally. Thanks @thirstycrow!

PR #396, opened 2015-07-18
Introduce ServiceFactorySocketAddress so that Names may bind local services.

olix0r:
Problem:

When building a client using a Name--i.e. via Dtabs--there is no mechanism to
cause a client to bind to a local Service reference.  Consider the following example:

A program--zoo--that exposes exposes an http endpoint /api/birds; and satisfying these requests requires a downstream client to a service--/s/birds.

We may typically run zoo with a base dtab like:

  /s => /$/io.bouyant.sd

/s/birds may be served by a downstream service, but it may also be
served from directly within the process (e.g. by an in-memory buffer).
This situation may arise when migrating functionality between
processes (e.g. decomposing a service) or mocking downstream
clients. In these cases, we want these requests routed in-process with
a dentry like:

  /s/birds => /$/io.buoyant.birds.local

The workaround today is to have zoo bind to an ephemeral port and
control redirects to that ephemeral port:

  /localPort => /$/inet/127.1
  /s/birds => /localPort/${birdsPort}

However, this approach incurs a full roundtrip through the networking stack.

Solution:

Introduce a SocketAddress type that encapsulates a ServiceFactory.
StdStackClient's default endpointer bypasses the client's Transporter when such
an address is encountered.

Result:

Local services may be named.

### DIFF
--- a/finagle-core/src/main/scala/com/twitter/finagle/Service.scala
+++ b/finagle-core/src/main/scala/com/twitter/finagle/Service.scala
@@ -223,6 +223,9 @@ abstract class ServiceFactoryProxy[-Req, +Rep](_self: ServiceFactory[Req, Rep])
   def self: ServiceFactory[Req, Rep] = _self
 }
 
+case class ServiceFactorySocketAddress[-Req, +Rep](factory: ServiceFactory[Req, Rep])
+  extends SocketAddress
+
 object FactoryToService {
   val role = Stack.Role("FactoryToService")
 
--- a/finagle-core/src/main/scala/com/twitter/finagle/Service.scala
+++ b/finagle-core/src/main/scala/com/twitter/finagle/Service.scala
@@ -223,7 +223,7 @@ abstract class ServiceFactoryProxy[-Req, +Rep](_self: ServiceFactory[Req, Rep])
   def self: ServiceFactory[Req, Rep] = _self
 }
 
-case class ServiceFactorySocketAddress[-Req, +Rep](factory: ServiceFactory[Req, Rep])
+case class ServiceFactorySocketAddress[Req, Rep](factory: ServiceFactory[Req, Rep])
   extends SocketAddress
 
 object FactoryToService {
--- a/finagle-core/src/main/scala/com/twitter/finagle/client/StackClient.scala
+++ b/finagle-core/src/main/scala/com/twitter/finagle/client/StackClient.scala
@@ -340,9 +340,14 @@ trait StdStackClient[Req, Rep, This <: StdStackClient[Req, Rep, This]]
       val parameters = Seq(implicitly[Stack.Param[Transporter.EndpointAddr]])
       def make(prms: Stack.Params, next: Stack[ServiceFactory[Req, Rep]]) = {
         val Transporter.EndpointAddr(addr) = prms[Transporter.EndpointAddr]
-        val endpointClient = copy1(params=prms)
-        val transporter = endpointClient.newTransporter()
-        Stack.Leaf(this, ServiceFactory(() => transporter(addr).map(endpointClient.newDispatcher)))
+        val factory = addr match {
+          case ServiceFactorySocketAddress(sf: ServiceFactory[Req, Rep]) => sf
+          case _ =>
+            val endpointClient = copy1(params=prms)
+            val transporter = endpointClient.newTransporter()
+            ServiceFactory(() => transporter(addr).map(endpointClient.newDispatcher))
+        }
+        Stack.Leaf(this, factory)
       }
     }
 
--- a/finagle-core/src/test/scala/com/twitter/finagle/client/StackClientTest.scala
+++ b/finagle-core/src/test/scala/com/twitter/finagle/client/StackClientTest.scala
@@ -399,7 +399,7 @@ class StackClientTest extends FunSuite
       client.configured[param.Label]((param.Label("foo"), param.Label.param))
   }
 
-  test("ServiceFactorySocketAddress bypasses transporter") {
+  test("StackClient binds to a local service via ServiceFactorySocketAddress") {
     val reverser = Service.mk[String, String] { in => Future.value(in.reverse) }
     val sf = ServiceFactory[String, String](() => Future.value(reverser))
     val sa = ServiceFactorySocketAddress[String, String](sf)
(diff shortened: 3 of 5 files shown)


olix0r:
per discussion with @mariusae

finagle-core/src/main/scala/com/twitter/finagle/Service.scala
```diff
@@ -223,6 +223,9 @@ abstract class ServiceFactoryProxy[-Req, +Rep](_self: ServiceFactory[Req, Rep])
   def self: ServiceFactory[Req, Rep] = _self
 }
 
+case class ServiceFactorySocketAddress[-Req, +Rep](factory: ServiceFactory[Req, Rep])
```
mosesn: what do the variance annotations buy us here?

olix0r: nothing.  copy-pasta. will remove.

finagle-core/src/main/scala/com/twitter/finagle/client/StackClient.scala
```diff
@@ -340,9 +340,14 @@ trait StdStackClient[Req, Rep, This <: StdStackClient[Req, Rep, This]]
       val parameters = Seq(implicitly[Stack.Param[Transporter.EndpointAddr]])
       def make(prms: Stack.Params, next: Stack[ServiceFactory[Req, Rep]]) = {
         val Transporter.EndpointAddr(addr) = prms[Transporter.EndpointAddr]
-        val endpointClient = copy1(params=prms)
-        val transporter = endpointClient.newTransporter()
-        Stack.Leaf(this, ServiceFactory(() => transporter(addr).map(endpointClient.newDispatcher)))
+        val factory = addr match {
+          case ServiceFactorySocketAddress(sf: ServiceFactory[Req, Rep]) => sf
```
mosesn: since these generics are erased, this is implicitly a cast.  could we instead make it:

```scala
  case ServiceFactorySocketAddress(sf: ServiceFactory[_, _]) => sf.asInstanceOf[ServiceFactory[Req, Rep]]
```

so that the cast becomes explicit?

finagle-core/src/test/scala/com/twitter/finagle/client/StackClientTest.scala
```diff
@@ -397,4 +398,16 @@ class StackClientTest extends FunSuite
     val client3: StackClient[String, String] =
       client.configured[param.Label]((param.Label("foo"), param.Label.param))
   }
+
+  test("ServiceFactorySocketAddress bypasses transporter") {
```
mosesn: can we either verify that the transporter isn't called, or change the name of this test?

mosesn:
Are you planning on adding a new Resolver in a future PR?

I'm a little worried about adding all of these different kinds of SocketAddress (like the Weighted one we added before) because I think we have some code which looks explicitly for InetSocketAddress.  Since the SocketAddress interface is so thin, does it make sense to start providing typeclasses for the extra behavior that we want from some SocketAddresses?  (I actually don't know what we're looking for).

From a quick search around:

```scala
 moses@Mosess-MacBook-Air ⮀ ~/projects/finagle ⮀ ⭠ master ⮀ ag ': InetSocketAddress =>' ./finagle-core
finagle-core/src/main/scala/com/twitter/finagle/builder/ClientBuilder.scala
642:      case inet: InetSocketAddress => Ssl.client(hostname, inet.getPort)
655:      case inet: InetSocketAddress => Ssl.client(sslContext, inet.getHostName, inet.getPort)
666:      case inet: InetSocketAddress => Ssl.client(sslContext, hostname.getOrElse(inet.getHostName), inet.getPort)
676:      case inet: InetSocketAddress => Ssl.clientWithoutCertificateValidation(inet.getHostName, inet.getPort)

finagle-core/src/main/scala/com/twitter/finagle/Codec.scala
122:    case ia: InetSocketAddress => ia

finagle-core/src/main/scala/com/twitter/finagle/dispatch/ClientDispatcher.scala
19:    case ia: InetSocketAddress => ia

finagle-core/src/main/scala/com/twitter/finagle/tracing/DestinationTracing.scala
21:            case ia: InetSocketAddress =>
27:            case ia: InetSocketAddress =>
66:      case ia: InetSocketAddress =>

finagle-core/src/main/scala/com/twitter/finagle/WeightedSocketAddress.scala
35:      case sa: InetSocketAddress => Some(sa, weight)
```

None of these is relevant for the ServiceFactorySocketAddress, but a few could be relevant for the WeightedSocketAddress.

mariusae:
I think the right thing to do here is, in effect, not use InetSocketAddr subclasses directly, but rather create our own wrapper type, ideally a sealed ADT.

olix0r:
also, i should point out that this obviously can't be type-safe end-to-end; and this is Okay. Today, we have no assurances with regard to the protocol of a named service and so we can't differentiate i.e. http and thrift interfaces at the name-level.  So this at least isn't a regression feature-wise, though we can encounter runtime cast exceptions.

@mosesn I added a ServiceNamer trait that makes it easy to build Service-backed namers. I don't think it's necessary to wire this through the Resolver interface unless there's a compelling reason.

roanta:
@olix0r The problem statement doesn't explain why/how this would be useful. Can you please add more details about the motivation for this?

olix0r:
@roanta consider the following:

Let's say I have a program--_zoo_--that exposes exposes an http
endpoint `/api/birds`; and satisfying these requests requires a
downstream client to a service--`/s/birds`.

We may typically run _zoo_ with a base dtab like:
```
/s => /$/io.bouyant.sd
```

`/s/birds` may be served by a downstream service, but it may also be
served from directly within the process (e.g. by an in-memory buffer).
This situation may arise when migrating functionality between
processes (e.g. decomposing a service) or mocking downstream
clients. In these cases, we want these requests routed in-process with
a dentry like:
```
/s/birds => /$/io.buotant.birds.local
```

The workaround today is to have _zoo_ bind to an ephemeral port and
control redirects to that ephemeral port:
```
/localPort => /$/inet/127.1
/s/birds => /localPort$/${birdsPort}
```

However, this approach incurs a full serialization roundtrip.  The
proposed change allows us to bypass the entire networking stack for
local requests.

mosesn:
@mariusae yeah, I think you're right.  I'll file a ticket internally.

:shipit: this looks fine to me, although I still don't love the implicit cast.

olix0r:
@mosesn made it invariant.  It *could* vary to let subclasses maintain Service-type flexibility but I think this isn't a concern practically.

luciferous:
@olix0r I don't think there's harm in having a verbose problem statement. If there's no objection, maybe we can update the problem statement with the information in your last comment.

roanta:
@olix0r cool, this seems like a nice patch.

vkostyukov:
This has been merged internally (by @luciferous). Thanks @olix0r!

PR #401, opened 2015-07-23
Server-side wire tracing.

olix0r:
Problem:

stack.  There is no analog on the server-side (i.e. a top-of-the-stack
indicator).

Furthermore, it looks like WireTracingFilter.module was using the
ClientTracingFilter.role instead of WireTracingFilter.role. This may
cause client tracing issues?

Solution:

Introduce WireTracingFilter.serverModule at the top of the default
server stack.  This module installs a filter that annotates WireRecv
on receipt and WireSend on success.

Result:

Tracers have a definitive 'top of stack' marker.

### DIFF
--- a/finagle-core/src/main/scala/com/twitter/finagle/tracing/TraceInitializerFilter.scala
+++ b/finagle-core/src/main/scala/com/twitter/finagle/tracing/TraceInitializerFilter.scala
@@ -258,11 +258,11 @@ private[finagle] object WireTracingFilter {
       def make(next: ServiceFactory[Req, Rep]) = filter andThen next
 
       object filter extends Filter[Req, Rep, Req, Rep] {
+        private[this] def recv(): Unit = Trace.record(Annotation.WireRecv)
+        private[this] def send(rsp: Rep): Unit = Trace.record(Annotation.WireSend)
         def apply(req: Req, service: Service[Req, Rep]) = {
-          Trace.record(Annotation.WireRecv)
-          service(req).onSuccess { _ =>
-            Trace.record(Annotation.WireSend)
-          }
+          recv()
+          service(req).onSuccess(send)
         }
       }
     }
--- a/finagle-core/src/main/scala/com/twitter/finagle/tracing/TraceInitializerFilter.scala
+++ b/finagle-core/src/main/scala/com/twitter/finagle/tracing/TraceInitializerFilter.scala
@@ -255,13 +255,12 @@ private[finagle] object WireTracingFilter {
     new Stack.Module0[ServiceFactory[Req, Rep]] {
       val role = WireTracingFilter.role
       val description = "Report wire recv/send events"
+      def make(next: ServiceFactory[Req, Rep]) = filter andThen next
       val filter = Filter.mk[Req, Rep, Req, Rep] { (req, svc) =>
         Trace.record(Annotation.WireRecv)
         svc(req) onSuccess { _ =>
-          Trace.record(Annotation.WireRecv)
+          Trace.record(Annotation.WireSend)
         }
       }
-      def make(next: ServiceFactory[Req, Rep]) =
-        filter andThen next
     }
 }
--- a/finagle-core/src/main/scala/com/twitter/finagle/tracing/TraceInitializerFilter.scala
+++ b/finagle-core/src/main/scala/com/twitter/finagle/tracing/TraceInitializerFilter.scala
@@ -241,7 +241,7 @@ private[finagle] object WireTracingFilter {
     finagleVersion,
     false)
 
-  def module[Req, Rep]: Stackable[ServiceFactory[Req, Rep]] =
+  def clientModule[Req, Rep]: Stackable[ServiceFactory[Req, Rep]] =
     new Stack.Module1[param.Label, ServiceFactory[Req, Rep]] {
       val role = WireTracingFilter.role
       val description = "Report finagle information and wire send/recv events"
--- a/finagle-core/src/main/scala/com/twitter/finagle/tracing/TraceInitializerFilter.scala
+++ b/finagle-core/src/main/scala/com/twitter/finagle/tracing/TraceInitializerFilter.scala
@@ -229,26 +229,26 @@ object ClientTracingFilter {
 private[finagle] object WireTracingFilter {
   val role = Stack.Role("WireTracingFilter")
 
-  case class TracingFilter[Req, Rep](
-    label: String,
-    finagleVersion: () => String = () => Init.finagleVersion
-  ) extends AnnotatingTracingFilter[Req, Rep](
-    label,
-    "clnt",
-    Annotation.WireSend,
-    Annotation.WireRecv,
-    Annotation.WireRecvError(_),
-    finagleVersion,
-    false)
-
   def clientModule[Req, Rep]: Stackable[ServiceFactory[Req, Rep]] =
     new Stack.Module1[param.Label, ServiceFactory[Req, Rep]] {
       val role = WireTracingFilter.role
       val description = "Report finagle information and wire send/recv events"
       def make(_label: param.Label, next: ServiceFactory[Req, Rep]) = {
         val param.Label(label) = _label
-        TracingFilter[Req, Rep](label) andThen next
+        Filter(label) andThen next
       }
+
+      case class Filter(
+        label: String,
+        finagleVersion: () => String = () => Init.finagleVersion
+      ) extends AnnotatingTracingFilter[Req, Rep](
+        label,
+        "clnt",
+        Annotation.WireSend,
+        Annotation.WireRecv,
+        Annotation.WireRecvError(_),
+        finagleVersion,
+        false)
     }
 
   def serverModule[Req, Rep]: Stackable[ServiceFactory[Req, Rep]] =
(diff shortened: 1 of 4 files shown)


olix0r:
disregard. i see this is actually in TraceInitializer.serverModule but not the http initializers.  will revisit.

olix0r:
Please excuse my flip-flopping on this (got confused by multiple branches ;/ ).  I think this change is desirable.  Note that I chose not to use `AnnotatingTracingFilter` since this is such a thin tracing layer.

finagle-core/src/main/scala/com/twitter/finagle/tracing/TraceInitializerFilter.scala
```diff
@@ -243,11 +243,24 @@ private[finagle] object WireTracingFilter {
 
   def module[Req, Rep]: Stackable[ServiceFactory[Req, Rep]] =
```
vkostyukov: Perhaps renaming this `module` to `clientModule` to make thing symmetric? Should be safe to do since it's package-private for `[finagle]`.

finagle-core/src/main/scala/com/twitter/finagle/tracing/TraceInitializerFilter.scala
```diff
@@ -243,11 +243,24 @@ private[finagle] object WireTracingFilter {
...
         TracingFilter[Req, Rep](label) andThen next
       }
     }
+
+  def serverModule[Req, Rep]: Stackable[ServiceFactory[Req, Rep]] =
+    new Stack.Module0[ServiceFactory[Req, Rep]] {
+      val role = WireTracingFilter.role
+      val description = "Report wire recv/send events"
+      def make(next: ServiceFactory[Req, Rep]) = filter andThen next
+      val filter = Filter.mk[Req, Rep, Req, Rep] { (req, svc) =>
```
atollena: Why not annotate on failures (with a new `WireSendError(error: String)`) here and reuse `AnnotatingTracingFilter`?

Otherwise LGTM.

olix0r: There is no such `Annotation` type as `WireSendError`, and I was trying to minimize the scope of this change--it would cause changes on the zipkin thrift struct, etc.  Furthermore, this wouldn't really be a "wire send error"--exceptions here would typically indicate some exception thrown somewhere on the server; and as it is, these exceptions are traced elsewhere in the stack.

olix0r: Bleh you're right I was getting confused with WireRecvError which does not
exist.  Still unsure if this makes sense semantically (it wouldn't be an
error encountered while trying to send on the wire)

On Friday, July 24, 2015, Antoine Tollenaere <notifications@github.com>
wrote:

> In
> finagle-core/src/main/scala/com/twitter/finagle/tracing/TraceInitializerFilter.scala
> <https://github.com/twitter/finagle/pull/401#discussion_r35445892>:
>
> >        val description = "Report finagle information and wire send/recv events"
> >        def make(_label: param.Label, next: ServiceFactory[Req, Rep]) = {
> >          val param.Label(label) = _label
> >          TracingFilter[Req, Rep](label) andThen next
> >        }
> >      }
> > +
> > +  def serverModule[Req, Rep]: Stackable[ServiceFactory[Req, Rep]] =
> > +    new Stack.Module0[ServiceFactory[Req, Rep]] {
> > +      val role = WireTracingFilter.role
> > +      val description = "Report wire recv/send events"
> > +      def make(next: ServiceFactory[Req, Rep]) = filter andThen next
> > +      val filter = Filter.mk[Req, Rep, Req, Rep] { (req, svc) =>
>
> Why not annotate on failures (with a new WireSendError(error: String))
> here and reuse AnnotatingTracingFilter?
>
> Otherwise LGTM.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/twitter/finagle/pull/401/files#r35445892>.
>


-- 
Oliver Gould <ver@buoyant.io>

vkostyukov:
@olix0r, do you think we should mention this in `CHNAGES`?

olix0r:
nudge. :wave:

atollena:
LGTM.

finagle-core/src/main/scala/com/twitter/finagle/tracing/TraceInitializerFilter.scala
```diff
@@ -229,25 +229,39 @@ object ClientTracingFilter {
...
+  def serverModule[Req, Rep]: Stackable[ServiceFactory[Req, Rep]] =
+    new Stack.Module0[ServiceFactory[Req, Rep]] {
+      val role = WireTracingFilter.role
+      val description = "Report wire recv/send events"
+      def make(next: ServiceFactory[Req, Rep]) = filter andThen next
+
+      val filter = Filter.mk[Req, Rep, Req, Rep] { (req, svc) =>
+        Trace.record(Annotation.WireRecv)
+        svc(req) onSuccess { _ =>
+          Trace.record(Annotation.WireSend)
```
kevinoliver: `.onSuccess` 

can you lift this function into a member to avoid an allocation on every `filter.apply`?

dschobel:
@olix0r do you still want this? if so, @tw-ngreen can you give it a review? I think you've probably got the most relevant experience from our side.

olix0r:
Sorry.  Yes, I still think this change is useful (we've just worked around it internally for the time being so i lost focus).  Let me rebase and address kevin's issue.

finagle-core/src/main/scala/com/twitter/finagle/tracing/TraceInitializerFilter.scala
```diff
@@ -229,25 +229,41 @@ object ClientTracingFilter {
...
+      val role = WireTracingFilter.role
+      val description = "Report wire recv/send events"
+      def make(next: ServiceFactory[Req, Rep]) = filter andThen next
+
+      object filter extends Filter[Req, Rep, Req, Rep] {
+        def apply(req: Req, service: Service[Req, Rep]) = {
+          Trace.record(Annotation.WireRecv)
+          service(req).onSuccess { _ =>
+            Trace.record(Annotation.WireSend)
+          }
```
dschobel: This filter still allocates on each request. Can you pull the onSuccess closure into a val outside the apply?

codeferret:
LGTM. Thanks!

finagle-core/src/main/scala/com/twitter/finagle/tracing/TraceInitializerFilter.scala
```diff
@@ -229,25 +229,41 @@ object ClientTracingFilter {
...
+      val role = WireTracingFilter.role
+      val description = "Report wire recv/send events"
+      def make(next: ServiceFactory[Req, Rep]) = filter andThen next
+
+      object filter extends Filter[Req, Rep, Req, Rep] {
+        private[this] def recv(): Unit = Trace.record(Annotation.WireRecv)
+        private[this] def send(rsp: Rep): Unit = Trace.record(Annotation.WireSend)
+        def apply(req: Req, service: Service[Req, Rep]) = {
+          recv()
+          service(req).onSuccess(send)
```
dschobel: so unfortunately this doesn't do the trick because the method will be eta-expanded into a function value on each request and that costs an allocation. anything but a function value burns us. I'll fix it up locally if that's cool with you.

dschobel:
posted internally, thanks for your patience @olix0r

mosesn:
@olix0r hey, sorry I didn't ask this before, but what's the backstory behind using a custom filter instead of AnnotatingTracingFilter?

olix0r:
@mosesn hey-- it's been a few months so i don't have 100% of my then-context in my now-brain, but I _believe_ that this is because we don't have an annotation that makes sense in the failure case...

mosesn:
Hmm, many of our protocols are still sending responses over the wire when we see a failure though, right?  Like we still say ServerSend on the server side.

olix0r:
To put this another way: what annotation should be emitted when the service fails?

mosesn:
Yeah, I'm not sure.  But if it sends back an http 5xx, or a mux TErr, it seems reasonable to still emit a wiresend.

olix0r:
I believe it's the case that if the server is actually going to respond with a message--for instance, a 5xx--then this service call actually succeeds with the result rather than an exception.  The exceptional case typically results in some connection loss or some such (perhaps mux handles this?  but this should be at the very top of the server stack so that there are no 'rescue' handlers installed above it.

In any case, I was trying to avoid introducing new annotation types.

mosesn:
Hmm, my impression was that this filter is on the server's service, ie the application-level behavior.  So the dispatcher will invoke the service (with this filter in it) and then will make a decision based on invoking this service.  Maybe this isn't really the right layer if we want to claim that we only write this trace after writing to the wire.

Dispatching looks something like:

```scala
read().flatMap { req: Req => svc(req) }.flatMap { rep: Rep => write(rep) }
```

so we would actually want to trace that we've written to the wire after the dispatcher writes successfully, but with this filter it would do it before we write it.  So maybe instead of adding this filter, we should have something that modifies dispatchers instead.

olix0r:
@mosesn yeah, I think I agree with you.  I can look into this approach.

mosesn:
Looks like https://github.com/twitter/finagle/pull/463 is probably the way forward here, so I'm closing this PR for now.

PR #403, opened 2015-07-28
Update README

lfcj:
Include directory to run command.
Also included `./sbt ...` as an option to be run, since some linux computers will only recognize it that way.

### DIFF
--- a/finagle-example/README
+++ b/finagle-example/README
@@ -1,12 +1,12 @@
 # How to run the examples:
 
-From the command line:
+From the command line in cloned directory `finangle`:
 
-% sbt "project finagle-example" run
+% sbt "project finagle-example" run || ./sbt "project finangle-example" run
 
 If the example requires a command line parameter, do this:
 
-% sbt "project finagle-example" "run p1 p2"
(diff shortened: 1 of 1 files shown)


finagle-example/README
```diff
@@ -1,12 +1,12 @@
 # How to run the examples:
 
-From the command line:
+From the command line in cloned directory `finangle`:
```
mosesn: git doesn't require that the cloned directory be named finagle, maybe we should say the root directory of the finagle repository?

mosesn:
I think the assumption with using raw sbt was that you have sbt installed on your machine.  Instead of using ||, it's probably fine to just make the examples use ./sbt by default.

mosesn:
Thanks for the contribution :]

finagle-example/README
spockz: Also there is a typo in `finangle`, it should be `finagle`.

dschobel:
@lfcj are you still interested in working on the open issues in this PR?

dschobel:
closing this until we hear back.

PR #405, opened 2015-08-10
build resource changing

tomzhang:
there is a issue with project/Build.sbt file

in line number: 17
val utilVersion = "6.26.0" + suffix

should remove "+ suffix" to enable utils modules to compile completely.


mosesn:
@tomzhang if you want to build against the develop branch, you should check out util and also do a publish-local against its develop branch.  We often publish changes in util snapshot and then consume them immediately in finagle snapshot, so we do this so that we don't break the build.

Instead of making this change, do you think you could improve the documentation to make this clear?  We mention it in the [CONTRIBUTING.md][https://github.com/twitter/finagle/blob/develop/CONTRIBUTING.md] file, but it sounds like it wasn't clear enough.

dschobel:
we can address docs in a separate PR.

PR #408, opened 2015-08-25
Rename finagle-memcachedx to finagle-memcached

luciferous:
This is an intermediate PR to serve as a reference for the changes needed for renaming the memcachedx module.


dschobel:
rename is finished, closing.

Issue #175, opened 2013-06-17
Allow retrieval of client certificate


bpfoster:
When using client auth with 2-way SSL, present the client certificate for custom authorization code to use.  The result should be something similar to the call of `SSLEngine.getSession().getPeerCertificateChain()`

Originally posted in the users group at https://groups.google.com/forum/?fromgroups#!topic/finaglers/j4xrFRF1MDI

danielpcox:
Does this commit on develop: https://github.com/twitter/finagle/commit/3d6f236f and the fact that Thrift RPC is now supported over TLS close this question?

mosesn:
I think so?  Have you done this by any chance :P

danielpcox:
Not yet, but it's on my list. :) I'll give it a shot next week and let you know.

danielpcox:
Make that the next three weeks... I'm pretty distracted at the moment. @mosesn would you be able to try it out? At the moment, I expect to be able to circle back around and confirm that this solves our problems by the end of next week, but anybody with some time to kill right now could pull develop and try extracting a client certificate over TLS via Thrift and HTTP.

I think you'd just have to write a little demo ping server that uses TLS and grabs some attribute out of the client cert, and then write a client to connect to it with any certificate.

mosesn:
End of next week is fine, I don't think anyone needs this more urgently than you do (and if they do, they can run their own tests :P)

danielpcox:
OK, I'm finally getting around to this now. I think I'm misunderstanding something fundamental, though. How do I actually get ahold of the Transport from inside a service's request handler? How can I do anything with the client's incoming certificate information so as to affect the response?

mosesn:
Looks like you should be able to call [`Transport.peerCertificate`][0] on the serverside.

I don't think it's officially deprecated yet, but we're [gently encouraging folks][1] to use the finagle-6 style APIs.

[0]: https://github.com/twitter/finagle/commit/3d6f236f#diff-d274cdcb169a7b7dbb5b2e81f9ea4959R93
[1]: http://twitter.github.io/finagle/guide/FAQ.html#how-do-i-configure-clients-and-servers-with-finagle-6-apis

danielpcox:
I'm sorry Moses, I still don't understand. How do I get a handle on the Transport object from within a server implementation? I've read all the documentation and I still only know how to do the most common things. I'm implementing a Thrift IDL-defined interface, and I don't see how the transport object could be injected at this point.

[...]

PR #412, opened 2015-09-03
fixing erroneous direct construction of Netty3Listener without params in finagle-thrift

danielpcox:
Greetings Finagle masters,

TL;DR In finagle-thrift (and none of the other protocol libraries), `Netty3Listener` was being constructed directly, without passing in the `params`.

I was originally trying to test the official solution to https://github.com/twitter/finagle/issues/175 and was able to get TLS over Thrift and extraction of the client certificate working using the old builder API, but _not with the new Finagle 6 API_. Eventually I discovered that `apply` on the `Netty3Listener` object was never being called, which led to this adjustment.

@mosesn and I discussed the removal of the hard-coded "thrift" label, and since it's only used as a backup scope for the stats receiver if the `ServerRegistry` doesn't have one for the listening address, and because it's manually configurable by the user, it seemed safe to remove it. I think it was probably only put there anyway because direct use of the constructor requires it.

Thanks to @mosesn @jeremyrsmith for helping me run this to ground.

### DIFF
--- a/finagle-thrift/src/main/scala/com/twitter/finagle/Thrift.scala
+++ b/finagle-thrift/src/main/scala/com/twitter/finagle/Thrift.scala
@@ -229,7 +229,7 @@ object Thrift extends Client[ThriftClientRequest, Array[Byte]] with ThriftRichCl
         if (framed) thrift.ThriftServerFramedPipelineFactory
         else thrift.ThriftServerBufferedPipelineFactory(protocolFactory)
 
-      Netty3Listener("thrift", pipeline)
+      Netty3Listener(pipeline, params)
     }
 
     protected def newDispatcher(


mosesn:
Have you tested this to make sure it solves the problem?

danielpcox:
I will be today. Just have to figure out how to build Finagle first. :)

danielpcox:
Yes. Hallelujah. I can confirm that this fix allows me to use the new Finagle 6 APIs to

a) do Thrift RPC over TLS and
b) extract the client certificate from the Transport to allow for PKI-based client authentication.

Examples of these miracles are here: https://github.com/DecipherNow/finagle-tls/tree/new-api
Obviously at the moment they require a manual publishM2 of this finagle-thrift fork to use.

mosesn:
:+1: shipit

nshkrob:
Makes sense. Can you add a simple unit test, e.g. by setting the Stats param and checking for stats?

danielpcox:
Sure, though it's crunch-time on another project so it'll have to wait until Monday.

suls:
I verified this with https://groups.google.com/forum/#!topic/finaglers/g_4i1xpfAJo ..

Would adding a thrift/TLS test case be overkill?

mosesn:
Nah, seems reasonable.  Do you want to make a PR against Daniel's branch?  The one caveat is that because we're going to squash this into one commit, only he would get the credit for the change.

nshkrob:
It would be definitely useful as a reference and as a test for this, thanks! @mosesn we can pull the test separately I think.

mosesn:
:+1:

PR #418, opened 2015-09-14
Make ServiceFactoryCache publicly accessible.

olix0r:
Problem

The logic encapsulated by `ServiceFactoryCache` is generally useful to
applications that build clients in dynamic ways.

Solution

Drop the 'private' classifier from the `ServiceFactoryCache` definition.

### DIFF
--- a/finagle-core/src/main/scala/com/twitter/finagle/factory/ServiceFactoryCache.scala
+++ b/finagle-core/src/main/scala/com/twitter/finagle/factory/ServiceFactoryCache.scala
@@ -58,7 +58,7 @@ private class IdlingFactory[Req, Rep](self: ServiceFactory[Req, Rep])
  * performance: one-shots could be created constantly for a hot cache
  * key, but should work well when there are a few hot keys.
  */
-private class ServiceFactoryCache[Key, Req, Rep](
+class ServiceFactoryCache[Key, Req, Rep](
     newFactory: Key => ServiceFactory[Req, Rep],
     statsReceiver: StatsReceiver = NullStatsReceiver,
     maxCacheSize: Int = 8)


mosesn:
LGTM

kevinoliver:
@olix0r its not obvious that this is a generally useful construct for end-users. Could you elaborate more on the use case?

A middle ground might be to mark if `private[finagle]` which would allow users to use it at their own risk by putting their code into `com.twitter.finagle`.

olix0r:
So, I'm okay with this being `private[finagle]`, but I also want some agreement that the API won't break or that this won't just disappear one release.  Making this public gives me a slightly warmer feeling in this regard.

Basically, I'm doing some Naming/BindingFactory-type things outside of finagle clients (for a very good reason, I assure you); and I need this functionality to avoid needless work.  For the time being I've just whole-hog copy/pasted `ServiceFactoryCache` (and `DynNameFactory` and `NameTreeFactory`) into my project.  This is okay, I suppose, but I'd also like the benefit of any bugfixes/improvements twitter makes in this space.

All in all, I'm fine with whatever you folks feel most comfortable supporting.

kevinoliver:
There wasn't general consensus that this should be public, and I generally prefer less public API support.

Lets do this as `private[finagle]` and we can try to remember to not destroy/mutilate it.

atollena:
@olix0r If we do decide that this isn't useful for BindingFactory anymore or that the interface must change, there will be a note in the change log (and you'll find out soon enough). At that point you can just copy the code to your project.

So LGTM.

olix0r:
I mean... isn't that just an argument against API stability in general?  "We may break this at any time and you'll have to figure out how not to be broken?"  If you want the ability to break this API in a minor release, I'll just copy the code now.

kevinoliver:
@olix0r a project can choose to keep all APIs frozen and rarely if ever modify them and of course that has downsides as well. The JDK is a good example of this and for the most part do a great job on it. But it leads to much slow release cycles and keeping broken APIs for a long time (forever so far). I highly doubt you are proposing that.

We try our best to keep public API stable and document it when we do break them. For things where we are less certain about their APIs we want to have more flexibility in modifying them. This is when we use the package private visibility modifiers and put APIs into the exp package. 

Its a balance or course and I hope we are doing a decent job on it.

olix0r:
@kevinoliver I understand. I suppose I'm just trying to register my interest in/dependency on this part of finagle.  If there are changes to these modules, I'd love to be involved in the discussion or at least aware of the changes before they land. To be clear: I'm absolutely not against API changes, though I am trying to avoid surprises (especially ones that are so deep/fundamental), and because finagle isn't developed in the open, it's hard to know what to expect.

So this PR is really trying to tease out:
- do you foresee these APIs changing anytime soon?
- if they do change, how can we do this in a non-surprising manner?  "`develop` will just change" is a bit scary to me.  Because we are feeding changes back into finagle, I suspect we'll be building against `develop` in the very near future...

kevinoliver:
I think @atollena and @jaked would best to answer your first question.

And regarding the latter we don't have any great mechanism for it today :/ Do you or anyone else have any ideas given how development works on this project?

olix0r:
@kevinoliver perhaps a finagle-dev list?  The finaglers list is fairly user-focused; but it might be nice to have a forum for discussing changes to internals. This may be too heavyweight for now considering that the list of people who care about this nitty-gritty level of detail is pretttty small.  It's not urgent, just something to keep an eye on if it actually causes problems.  (I'm of course open to other suggestions for lightweight synchronization on code changes).

Thanks for humoring me!

kevinoliver:
I like the idea of a finagle-dev@ list. (+@travisbrown)

dschobel:
merged in 96e91a9b629edb408052cdfa5323ae6507e2a257. thanks @olix0r.

PR #419, opened 2015-09-20
Fixes bad sbt download link

adriancole:
Probably should create a travis file so that instructions don't rust.


adriancole:
NM.. this was against the master branch

PR #428, opened 2015-10-17
finagle-thrit: add per RPC request stats

zfy0701:
an alternative is to do those stats in generated code
but use filter seems more clear to me and it doesn't rely on a specific generator. let me know what's your thought.

### DIFF
--- a/finagle-thrift/src/main/scala/com/twitter/finagle/thrift/TTwitterServerFilter.scala
+++ b/finagle-thrift/src/main/scala/com/twitter/finagle/thrift/TTwitterServerFilter.scala
@@ -1,9 +1,10 @@
 package com.twitter.finagle.thrift
 
 import com.twitter.finagle.context.Contexts
+import com.twitter.finagle.stats.ServerStatsReceiver
 import com.twitter.finagle.tracing.Trace
 import com.twitter.finagle.util.ByteArrays
-import com.twitter.finagle.{Service, SimpleFilter, Dtab}
+import com.twitter.finagle.{Dtab, Service, SimpleFilter}
 import com.twitter.io.Buf
 import com.twitter.util.Future
 import org.apache.thrift.protocol.{TMessage, TMessageType, TProtocolFactory}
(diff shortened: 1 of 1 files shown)


finagle-thrift/src/main/scala/com/twitter/finagle/thrift/TTwitterServerFilter.scala
```diff
@@ -59,11 +60,10 @@ private[finagle] class TTwitterServerFilter(
...
           }
         }
-
-        Trace.recordRpc({
-          val msg = new InputBuffer(request_, protocolFactory)().readMessageBegin()
-          msg.name
-        })
+        val msg = new InputBuffer(request_, protocolFactory)().readMessageBegin()
+        Trace.recordRpc(msg.name)
+        val rpcStatsReciver = ServerStatsReceiver.scope(serviceName).scope(msg.name)
```
mosesn: for what it's worth, scoping new stats receivers, and creating new counters should never be done on the hot path unless it's impossible to know the name of the counter until runtime.  since we know the names in advance, we can do better.

mosesn:
@zfy0701 I think we already have this in the [generated code][0].  Could you take a look and see if your code was generated with them?

[0]: https://github.com/twitter/scrooge/blob/develop/scrooge-generator/src/main/resources/scalagen/finagleClientFunction.scala

zfy0701:
the generated code you mentioned is just for client code

the server side code is at here:
https://github.com/twitter/scrooge/blob/develop/scrooge-generator/src/main/resources/scalagen/finagleServiceFunction.scala#L1

let me know if you prefer we add that in generated code.

mosesn:
Oh, I understand now.  Happy to accept this here if you can avoid reallocating new stats and counters on every request, but if you can't do it with this API, let's put it in the scrooge generated code, like with the client.

zfy0701:
To avoid create stats receiver every time I could add a member variable and let it get initialized the first time it is used.
However I feel unsafe because I didn't feel there is a hard constraint that one filter can be only used for one rpc. 

So I opened another pull request for scrooge
https://github.com/twitter/scrooge/pull/211

thanks for reviewing my change!

PR #429, opened 2015-10-18
[docs] Add missing return type

robsonpeixoto:


doc/src/sphinx/Futures.rst
```diff
@@ -168,7 +168,7 @@ The following retries a request infinitely should it fail with a
 
 	def fetchUrl(url: String): Future[HttpResponse]
```
mosesn: actually, since we no longer expose netty types in the public finagle API, how about we change this to http.Response?

robsonpeixoto: I'm a finagle noob. I just get this recursion without the return.

mosesn: sorry, I should have been more clear.  could we change the return type to http.Response?  HttpResponse is a netty3 type, and finagle no longer exposes netty3 types in the public API, since we're trying to switch to netty4.  So the change here will look like:

```scala
def fetchUrl(url: String): Future[http.Response]
def fetchUrlWithRetry(url: String): Future[http.Response] =...
```

Does that make more sense?

robsonpeixoto: Yes @mosesn. And will use the same type of the quickstart.

mosesn:
LGTM

adleong:

robsonpeixoto:
Thanks

PR #431, opened 2015-10-20
finagle-thrift: Update serverFromIface to try new style constructor.

zfy0701:
Problem:
In https://github.com/twitter/scrooge/pull/211, an new constructor with service name as
paramter is added.

Solution:
Try to find the new style constructor but fallback to old one if not found.

### DIFF
--- a/finagle-thrift/src/main/scala/com/twitter/finagle/rich.scala
+++ b/finagle-thrift/src/main/scala/com/twitter/finagle/rich.scala
@@ -3,7 +3,6 @@ package com.twitter.finagle
 import com.twitter.finagle.stats._
 import com.twitter.finagle.thrift._
 import com.twitter.finagle.util.Showable
-import com.twitter.scrooge.{ThriftMethod, ThriftStruct}
 import com.twitter.util.NonFatal
 import java.lang.reflect.{Constructor, Method}
 import java.net.SocketAddress
@@ -145,7 +144,8 @@ private[twitter] object ThriftUtil {
     impl: AnyRef,
     protocolFactory: TProtocolFactory,
     stats: StatsReceiver,
-    maxThriftBufferSize: Int
+    maxThriftBufferSize: Int,
+    label: String
   ): BinaryService = {
     def tryThriftFinagleService(iface: Class[_]): Option[BinaryService] =
       for {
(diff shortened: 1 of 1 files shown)


finagle-thrift/src/main/scala/com/twitter/finagle/rich.scala
```diff
@@ -145,7 +146,8 @@ private[twitter] object ThriftUtil {
     impl: AnyRef,
     protocolFactory: TProtocolFactory,
     stats: StatsReceiver,
-    maxThriftBufferSize: Int
+    maxThriftBufferSize: Int,
+    label: String
```
mosesn: hmm, we don't want to break binary backcompat.  can you overload with the old method too?

finagle-thrift/src/main/scala/com/twitter/finagle/rich.scala
```diff
@@ -155,14 +157,25 @@ private[twitter] object ThriftUtil {
...
-      for {
+      (for {
         baseName   <- findRootWithSuffix(iface.getName, "$FutureIface") orElse
           Some(iface.getName)
         serviceCls <- findClass[BinaryService](baseName + "$FinagleService") orElse
           findClass[BinaryService](baseName + "$FinagledService")
         baseClass  <- findClass1(baseName)
-        cons       <- findConstructor(serviceCls, baseClass, classOf[TProtocolFactory], classOf[StatsReceiver], Integer.TYPE)
-      } yield cons.newInstance(impl, protocolFactory, stats, Int.box(maxThriftBufferSize))
+      } yield {
```
mosesn: oh god.  thanks for doing this, this is crazy.  could you add a comment as to why we need this?

mosesn:
LGTM except for comment, will try to get more eyes on this.

blackicewei:
LGTM.  @zfy0701, can you resolve Moses's comment? I will pull it in afterwards.

zfy0701:
Added comments. Thank you

vkostyukov:
Thanks @zfy0701! I'm pulling this internally.

mosesn:
:+1:

vkostyukov:
Hey @zfy0701!

Any chance we can test this? I run this PR throght the sandbox and I can see a couple of failed tests internally. Basically, it crashes on `cons => cons.newInstance(newArgs: _*)` with "Wrong number of arguments". 

Do you think we can test this for both cases: w/o new constructor and w/ new constructor?

zfy0701:
@vkostyukov sorry I just realized I did something wrong (I should cast the newArgs as a Seq[Object]) and I probably didn't run the right set of test to check it.

I could do manual check on both case, but having automatic test on both version might be hard (I'm really really new to  finagle/scrooge repository).

Also it would be great if you could let me know which test is failing.

zfy0701:
@vkostyukov I found the cause is that I indeed pass one more extra argument for oldArgs, and I'm able to reproduce the problem by running finagle tests, after my this fix, the problem should be no more, could you try pull it again?

vkostyukov:
Thanks @zfy0701! I'm happy to pull this again! Will update this PR once it passes or not passes the internal tests.

mosesn:
This was merged internally, thanks for your patience!  I'll update this ticket again when we publish it to the `develop` branch.

vkostyukov:
This made it to the develop branch: https://github.com/twitter/finagle/commit/d2453314d0a9411a20c82d83cd9eebbfc7fd3288

PR #433, opened 2015-10-23
FIX: Race condition in IdleConnectionFilter

gpoulin:
If a connection is closed by the client during a request, `onClose`
could be filled (and the connection "removed" from the queue), before
the connection is added back to queue.

### DIFF
--- a/finagle-core/src/main/scala/com/twitter/finagle/netty3/channel/IdleConnectionFilter.scala
+++ b/finagle-core/src/main/scala/com/twitter/finagle/netty3/channel/IdleConnectionFilter.scala
@@ -108,7 +108,13 @@ class IdleConnectionFilter[Req, Rep](
     def apply(request: Req, service: Service[Req, Rep]) = {
       queue.remove(c)
       service(request) ensure {
-        queue.touch(c)
+        // The connection might have been closed by the client during the request
+        if (c.onClose.isDefined) {
+          queue.touch(c)
+          if (c.onClose.isDefined) {
+            queue.remove(c)
+          }
+        }
       }
     }
   }
--- a/CHANGES
+++ b/CHANGES
@@ -38,6 +38,13 @@ Breaking API Changes
   * finagle-core: `BackupRequestLost` is no longer itself an `Exception`. Use
     `BackupRequestLost.Exeption` in its place. ``RB_ID=758056``
 
+Bug Fixes
+~~~~~~~~~
+
+  * finagle-core: Fix race condition in `c.t.f.netty3.channel.IdleConnectionFilter` when
+    a connection is closed by the client.
+
+
 6.30.0
 ~~~~~~~
 


mosesn:
could you follow the instructions from the CONTRIBUTING guide?  It could also be useful to update the CHANGES file to note the bugfix.

So this is a little tricky because it's concurrent.  This means that we could get weird orderings, and makes it difficult to be efficient.

In particular, what we're concerned about is:

t1: queue.remove()
t2: queue.remove()
t1: queue.touch()

your proposed solution fixes it so that instead it looks like:

t1: queue.remove()
t2: queue.remove()
t1: queue.touch(); queue.remove()

this can also manifest as:

t1: queue.remove(); queue.touch()
t2: queue.remove()
t1: queue.remove()

we can do somewhat better, by guarding on if the connection has been closed already.

```scala
if (!c.onClose.isDefined) {
  queue.touch(c)
  if (c.onClose.isDefined) {
    queue.remove(c)
  }
}
```

so that in this case, the common case interleaving should be:

t1: queue.remove()
t2: queue.remove()

and the much less common case should be:

t1: queue.remove()
t2: queue.remove()
t1: queue.touch(); queue.remove()

We must allow this race if we don't want to add additional synchronization, because onClose is racy–we could read that onClose is not defined, but then onClose could fire and remove it.  But it means it only happens in that extreme race condition.

gpoulin:
@mosesn Sorry for the delay. I added comment in the changelog and added our extra verification to make the implementation more performant.

I read back more in detail the CONTRIBUTING.md and fix my PR accordinly. However, I'm not sure of which part you are referring to. Let me know if there is still issues with this PR.

finagle-core/src/main/scala/com/twitter/finagle/netty3/channel/IdleConnectionFilter.scala
```diff
@@ -108,7 +108,13 @@ class IdleConnectionFilter[Req, Rep](
     def apply(request: Req, service: Service[Req, Rep]) = {
       queue.remove(c)
       service(request) ensure {
-        queue.touch(c)
+        // The connection might have been closed by the client during the request
+        if (c.onClose.isDefined) {
```
mosesn: I think this should be `if (!c.onClose.isDefined)`

mosesn:
CHANGES and commit message look great, thanks!  Just one last fix.

mosesn:
Actually, it occurs to me that it would be nice to add a test for this case.

mosesn:
@gpoulin do you think you'll have an opportunity to take a look sometime?  no rush :smile_cat:

mosesn:
@gpoulin just wanted to check in if you think you'll have any more time to work on finishing this?

mosesn:
@gpoulin this is 99% of the way there, so we're going to pull it in and try to add a test on our end.  thanks so much!

mosesn:
So it looks like IdleConnectionFilter is no longer used in finagle, so I'm going to close this PR and delete the class.  Please let me know if you still need it for some reason!

PR #441, opened 2015-11-28
Make RegistryEntryLifecycle.role static

olix0r:
Problem

RegistryEntryLifecycle's role may not be accessed without instantiating a new
module.  This is cumbersome when referencing this module via e.g.
Stack.insertAfter.

Solution

Make RegistryEntryLifecycle.role static.


mosesn:
LGTM

mosesn:
Actually, would you mind changing your commit message to be prefixed with `finagle-core: `

dschobel:
lgtm

dschobel:
thanks for the patch, it just merged internally, expect it in develop on Monday.

PR #443, opened 2015-12-03
Redis sentinel support

thirstycrow:


.gitignore
```diff
@@ -20,3 +20,6 @@ out/
 sbt-launch-0.7.5.jar
 sbt-launch.jar
 *.pyc
+.project
```
mosesn: please don't add these (I think these are eclipse files?) can you instead add these to your global gitignore?

finagle-redis/src/main/scala/com/twitter/finagle/redis/Client.scala
```diff
@@ -82,19 +66,46 @@ class BaseClient(service: Service[Command, Reply]) {
     }
 
   /**
-   * Closes connection to Redis instance
+   * Select DB with specified zero-based index
+   * @param index
    */
-  def quit(): Future[Unit] =
-    doRequest(Quit) {
+  def select(index: Int): Future[Unit] =
```
mosesn: can you add these methods to `BaseClient` instead of `Client`, to be consistent with the regular style here?

finagle-redis/src/main/scala/com/twitter/finagle/redis/Client.scala
```diff
@@ -82,19 +66,46 @@ class BaseClient(service: Service[Command, Reply]) {
...
+ * @param service: Finagle service object built with the Redis codec
+ */
+class BaseClient(service: Service[Command, Reply]) {
+
+  /**
+   * Returns information and statistics about the server
+   * @param section Optional parameter can be used to select a specific section of information
+   * @return ChannelBuffer with collection of \r\n terminated lines if server has info on section
+   */
+  def info(section: ChannelBuffer = ChannelBuffers.EMPTY_BUFFER): Future[Option[ChannelBuffer]] =
```
mosesn: if this is optional, let's make this `Option[ChannelBuffer]` instead?

finagle-redis/src/main/scala/com/twitter/finagle/redis/Client.scala
```diff
@@ -82,19 +66,46 @@ class BaseClient(service: Service[Command, Reply]) {
...
+
+/**
+ * Connects to a single Redis host
+ * @param service: Finagle service object built with the Redis codec
+ */
+class BaseClient(service: Service[Command, Reply]) {
+
+  /**
+   * Returns information and statistics about the server
+   * @param section Optional parameter can be used to select a specific section of information
```
mosesn: could you mention the values that the parameters can be?  if you don't think it's appropriate to paste it, could you link to the relevant documentation on the redis website?

finagle-redis/src/main/scala/com/twitter/finagle/redis/Client.scala
```diff
@@ -82,19 +66,46 @@ class BaseClient(service: Service[Command, Reply]) {
...
+   * @return ChannelBuffer with collection of \r\n terminated lines if server has info on section
+   */
+  def info(section: ChannelBuffer = ChannelBuffers.EMPTY_BUFFER): Future[Option[ChannelBuffer]] =
+    doRequest(Info(section)) {
+      case BulkReply(message) => Future.value(Some(message))
+      case EmptyBulkReply() => Future.value(None)
```
mosesn: in what cases does it return an EmptyBulkReply?  can we just return `Future[ChannelBuffer]`?

finagle-redis/src/main/scala/com/twitter/finagle/redis/Client.scala
```diff
@@ -120,6 +131,15 @@ class BaseClient(service: Service[Command, Reply]) {
     messages.grouped(2).toSeq flatMap { case Seq(a, b) => Some(a, b); case _ => None }
   }
 
+  /**
+   * Helper function to convert a Redis multi-bulk reply into a Map[String, String]
+   */
+  private[redis] def returnMap(messages: Seq[ChannelBuffer]) = {
```
mosesn: could you add a return type annotation here?  I can guess, but it's nice to have it spelled out.

finagle-redis/src/main/scala/com/twitter/finagle/redis/SentinelClient.scala
```diff
@@ -0,0 +1,85 @@
...
+import org.jboss.netty.buffer.{ ChannelBuffers, ChannelBuffer }
+
+object SentinelClient {
+
+  /**
+   * Construct a sentinel client from a single host.
+   * @param host a String of host:port combination.
+   */
+  def apply(host: String): SentinelClient = SentinelClient(
+    ClientBuilder()
```
mosesn: could you instead use the finagle 6 API to construct this?  that way we get to use pipelining, etc.

finagle-redis/src/main/scala/com/twitter/finagle/redis/SentinelClient.scala
```diff
@@ -0,0 +1,85 @@
...
+      .build())
+
+  /**
+   * Construct a sentinel client from a single Service.
+   */
+  def apply(raw: Service[Command, Reply]): SentinelClient =
+    new SentinelClient(raw)
+  
+  object Node {
+    def apply(props: Map[String, String]) = {
```
mosesn: scaladoc? in particular, props must have a role-reported key, or this will throw.

mosesn: feel free to ignore: may call this fromMap or fromProperties?

finagle-redis/src/main/scala/com/twitter/finagle/redis/SentinelClient.scala
```diff
@@ -0,0 +1,85 @@
...
+        case "master" => new MasterNode(props)
+        case "slave" => new SlaveNode(props)
+        case "sentinel" => new SentinelNode(props)
+      }
+    }
+  }
+  
+  sealed trait Node {
+    val props: Map[String, String]
+    lazy val name: String = props("name")
```
mosesn: why make these all lazy?  these are all cheap operations, let's strict 'em.

finagle-redis/src/main/scala/com/twitter/finagle/redis/SentinelCommands.scala
```diff
@@ -0,0 +1,141 @@
+package com.twitter.finagle.redis
+
+import _root_.java.lang.{ Long => JLong, Boolean => JBoolean }
```
mosesn: I don't think this needs to be fully qualified.

finagle-redis/src/main/scala/com/twitter/finagle/redis/SentinelCommands.scala
```diff
@@ -0,0 +1,141 @@
...
+import scala.collection.immutable.{ Set => ImmutableSet }
+import com.twitter.finagle.redis.protocol._
+import com.twitter.finagle.redis.protocol.commands._
+import com.twitter.finagle.redis.util._
+import com.twitter.util.Future
+import org.jboss.netty.buffer.ChannelBuffer
+import java.net.InetAddress
+import java.net.InetSocketAddress
```
mosesn: sort imports, please

finagle-redis/src/main/scala/com/twitter/finagle/redis/SentinelCommands.scala
```diff
@@ -0,0 +1,141 @@
...
+
+trait SentinelCommands { self: BaseClient =>
+
+  import SentinelClient._
+
+  /**
+   * Show a list of monitored masters and their state.
+   */
+  def masters(): Future[List[MasterNode]] =
```
mosesn: let's prefer `Seq` over `List`

finagle-redis/src/main/scala/com/twitter/finagle/redis/SentinelCommands.scala
```diff
@@ -0,0 +1,141 @@
...
+    doRequest(SentinelMasters()) {
+      case MBulkReply(masters) => Future.value(
+        masters map {
```
mosesn: let's prefer `masters.map {` here

finagle-redis/src/main/scala/com/twitter/finagle/redis/Client.scala
mosesn: since this needs to be Strings eventually, can we take Seq[String] as an argument and use ReplyFormat.toString instead of ReplyFormat.toChannelBuffer?

finagle-redis/src/main/scala/com/twitter/finagle/redis/SentinelCommands.scala
```diff
@@ -0,0 +1,141 @@
...
+      case EmptyMBulkReply() => Future.value(Nil)
+    }
+
+  /**
+   * Show a list of sentinel instances for this master, and their state.
+   */
+  def sentinels(name: ChannelBuffer): Future[List[SentinelNode]] =
+    doRequest(SentinelSentinels(name)) {
+      case MBulkReply(names) => Future.value(
+        names map {
```
mosesn: `names.map {`

finagle-redis/src/main/scala/com/twitter/finagle/redis/SentinelCommands.scala
```diff
@@ -0,0 +1,141 @@
...
+        })
+      case EmptyMBulkReply() => Future.value(Nil)
+    }
+
+  /**
+   * Show the state and info of the specified master.
+   */
+  def master(name: ChannelBuffer): Future[MasterNode] =
+    doRequest(SentinelMaster(name)) {
+      case MBulkReply(messages) => Future.value(
```
mosesn: is it possible that there's no master, for example if the master just died and the dead master has been detected, but leader election hasn't finished yet?

finagle-redis/src/main/scala/com/twitter/finagle/redis/SentinelCommands.scala
```diff
@@ -0,0 +1,141 @@
...
+    doRequest(SentinelFailover(name)) {
+      case StatusReply(msg) => Future.Unit
+    }
+
+  /**
+   * Check if the current Sentinel configuration is able to reach the quorum
+   * needed to failover a master, and the majority needed to authorize the
+   * failover.
+   */
+  def ckquorum(name: ChannelBuffer): Future[String] =
```
mosesn: s/ckquorum/ckQuorum

finagle-redis/src/main/scala/com/twitter/finagle/redis/SentinelCommands.scala
```diff
@@ -0,0 +1,141 @@
...
+  def flushConfig(): Future[Unit] =
+    doRequest(SentinelFlushConfig()) {
+      case StatusReply(msg) => Future.Unit
+    }
+
+  /**
+   * Tells the Sentinel to start monitoring a new master with the
+   * specified name, ip, port, and quorum.
+   */
+  def monitor(name: ChannelBuffer, addr: InetSocketAddress, quorum: Int): Future[Unit] =
```
mosesn: feel free to ignore: unless you use `InetSocketAddress.createUnresolved` this will entail a DNS resolution.  Maybe it would be simpler to require that users specify a host and port separately?

finagle-redis/src/main/scala/com/twitter/finagle/redis/protocol/commands/Sentinels.scala
```diff
@@ -0,0 +1,181 @@
...
+import java.net.InetSocketAddress
+
+import org.jboss.netty.buffer.ChannelBuffer
+import org.jboss.netty.buffer.ChannelBuffers
+
+case class SentinelMaster(name: ChannelBuffer)
+  extends Sentinel(SentinelMaster.channelBuffer, Seq(name))
+
+object SentinelMaster extends SentinelHelper {
+  val command = "MASTER"
```
mosesn: could you follow the style here and and put these commands in Command.scala?

finagle-redis/src/main/scala/com/twitter/finagle/redis/protocol/commands/Sentinels.scala
```diff
@@ -0,0 +1,181 @@
+package com.twitter.finagle.redis.protocol.commands
+
+import com.twitter.finagle.redis.protocol._
+import com.twitter.finagle.redis.protocol.Commands._
+import com.twitter.finagle.redis.util._
+import com.twitter.finagle.redis.ClientError
+
+import java.net.InetSocketAddress
+
+import org.jboss.netty.buffer.ChannelBuffer
```
mosesn: merge me

finagle-redis/src/main/scala/com/twitter/finagle/redis/protocol/commands/Sentinels.scala
```diff
@@ -0,0 +1,181 @@
...
+import org.jboss.netty.buffer.ChannelBuffers
+
+case class SentinelMaster(name: ChannelBuffer)
+  extends Sentinel(SentinelMaster.channelBuffer, Seq(name))
+
+object SentinelMaster extends SentinelHelper {
+  val command = "MASTER"
+  def apply(args: Seq[Array[Byte]]): SentinelMaster = {
+    val name = trimList(args, 2, "SENTINEL MASTER")(0)
```
mosesn: why doesn't this match up with the command?

finagle-redis/src/main/scala/com/twitter/finagle/redis/protocol/commands/Sentinels.scala
```diff
@@ -0,0 +1,181 @@
...
+    new SentinelMaster(ChannelBuffers.wrappedBuffer(name))
+  }
+}
+
+case class SentinelMasters() extends Sentinel(SentinelMasters.channelBuffer, Nil)
```
mosesn: case object?

finagle-redis/src/main/scala/com/twitter/finagle/redis/protocol/commands/Sentinels.scala
```diff
@@ -0,0 +1,181 @@
...
+
+object Sentinel {
+
+  val subCommands: Seq[SentinelHelper] = Seq(
+    SentinelMaster, SentinelMasters, SentinelMonitor, SentinelRemove, SentinelSet)
+
+  def apply(args: Seq[Array[Byte]]): Sentinel = {
+    val subCommandString = new String(trimList(args.headOption.toList, 1, "SENTINEL")(0)).toUpperCase
+    val subCommand = subCommands.find { _.command == subCommandString }
+      .getOrElse(throw ClientError("Invalid Sentinel command " + subCommandString))
```
mosesn: string interpolate?

finagle-redis/src/test/scala/com/twitter/finagle/redis/commands/sentinel/SentinelClientIntegrationSuite.scala
```diff
@@ -0,0 +1,225 @@
...
+  val noSuchMaster = masterName(999)
+
+  override def beforeAll(): Unit = {
+    super.beforeAll()
+    for {
+      i <- 0 until masterCount
+      j <- 1 to slavesPerMaster
+    } withRedisClient(i + masterCount * j) { client =>
+      val (host, port) = hostAndPort(redisAddress(i))
+      Await.ready(client.slaveOf("127.0.0.1", port))
```
mosesn: could you add a timeout to the Await.ready?  it's a pain when something hangs and we don't have a timeout on it.

finagle-redis/src/test/scala/com/twitter/finagle/redis/commands/sentinel/SentinelClientIntegrationSuite.scala
```diff
@@ -0,0 +1,225 @@
...
+    }
+  }
+
+  private def masterName(i: Int) = "master" + i
+
+  private def waitUntil(msg: String)(check: => Boolean) = {
+    println(msg)
```
mosesn: rm println

finagle-redis/src/test/scala/com/twitter/finagle/redis/commands/sentinel/SentinelClientIntegrationSuite.scala
```diff
@@ -0,0 +1,225 @@
...
+    val startTime = Time.now
+    val until = startTime + 20.seconds
+    def checkLater(): Future[Boolean] = {
+      if (Time.now > until) Future.value(false)
+      else DefaultTimer.twitter.doLater(1.second) {
+        println(Time.now - startTime)
+        if (check) Future.value(true)
+        else checkLater
```
mosesn: use parens–this isn't an accessor by any means

finagle-redis/src/test/scala/com/twitter/finagle/redis/commands/sentinel/SentinelClientIntegrationSuite.scala
```diff
@@ -0,0 +1,225 @@
...
+      j <- 1 to slavesPerMaster
+    } withRedisClient(i + masterCount * j) { client =>
+      val (host, port) = hostAndPort(redisAddress(i))
+      Await.ready(client.slaveOf("127.0.0.1", port))
+    }
+  }
+
+  private def masterName(i: Int) = "master" + i
+
+  private def waitUntil(msg: String)(check: => Boolean) = {
```
mosesn: github shows that this line is gone, but I'm not sure how that could be possible . . . weird.

finagle-redis/src/test/scala/com/twitter/finagle/redis/commands/sentinel/SentinelClientIntegrationSuite.scala
```diff
@@ -0,0 +1,225 @@
...
+    val until = startTime + 20.seconds
+    def checkLater(): Future[Boolean] = {
+      if (Time.now > until) Future.value(false)
+      else DefaultTimer.twitter.doLater(1.second) {
+        println(Time.now - startTime)
+        if (check) Future.value(true)
+        else checkLater
+      }.flatten
+    }
+    assert(Await.result(checkLater))
```
mosesn: timeout

finagle-redis/src/test/scala/com/twitter/finagle/redis/commands/sentinel/SentinelClientIntegrationSuite.scala
mosesn: actually, I think I was just selecting it and it seemed reddish.  nvm.  seems fine now.

finagle-redis/src/test/scala/com/twitter/finagle/redis/commands/sentinel/SentinelClientIntegrationSuite.scala
```diff
@@ -0,0 +1,225 @@
...
+  }
+
+  test("Correctly perform the MONITOR and MASTER/MASTERS command", RedisTest, ClientTest) {
+    withSentinelClient(0) { client =>
+      // No masters at the beginning
+      assert(Await.result(client.masters()) === Nil)
```
mosesn: prefer == since scalatest 2.x the assert macro is dope

finagle-redis/src/test/scala/com/twitter/finagle/redis/commands/sentinel/SentinelClientIntegrationSuite.scala
```diff
@@ -0,0 +1,225 @@
...
+
+      val expected = (0 until masterCount).map { i =>
+        val name = masterName(i)
+        val address = redisAddress(i)
+        Await.ready(client.monitor(name, address, sentinelCount + i))
```
mosesn: timeout (and for all other Awaits in this file)

finagle-redis/src/test/scala/com/twitter/finagle/redis/commands/sentinel/SentinelClientIntegrationSuite.scala
```diff
@@ -0,0 +1,225 @@
...
+        (name -> (address.getHostString, address.getPort, sentinelCount + i))
+      }.toMap
+
+      val masters = Await.result(client.masters())
+        .map(m => m.name -> (m.ip, m.port, m.quorum))
+        .toMap
+      assert(masters === expected)
```
mosesn: == (here and elsewhere)

mosesn:
Wow, this review is awesome!  :+1: I'll try to get someone from the cache team over on our side to take a look at it too.

One thing I'm wondering about, should we provide sentinel-specific tooling?  For example, should we provide a read-only sentinel `Service` abstraction, where you can read from anybody, but only write to the master?

Also, is probably out of scope, but the way that the `AUTH` command is implemented is that it only checks for one client, which was fine because we previously only wanted to connect to one redis instance, but if a few redises can be treated as the same, it might make sense to treat them as a regular cluster–if that's the case, authentication should be treated specially, like it is in mysql, so that it happens as part of a handshake, instead of requiring that users do it themselves.

finagle-redis/src/main/scala/com/twitter/finagle/redis/Client.scala
thirstycrow: They were in BaseClient. I moved them to Client, because they are not available in sentinel and/or cluster mode.

finagle-redis/src/main/scala/com/twitter/finagle/redis/Client.scala
thirstycrow: EmptyBulkReply is returned when the section is unknown to the redis server.

finagle-redis/src/main/scala/com/twitter/finagle/redis/SentinelCommands.scala
thirstycrow: It always returns something about the master, including its status, unless it is removed. In that case, an exception will be thrown.

finagle-redis/src/main/scala/com/twitter/finagle/redis/protocol/commands/Sentinels.scala
thirstycrow: "MASTER" is a subcommand of "SENTINEL"

finagle-redis/src/test/scala/com/twitter/finagle/redis/commands/sentinel/SentinelClientIntegrationSuite.scala
thirstycrow: It is to avoid long running tests without any feedback. Shall I replace it with logging?

finagle-redis/src/main/scala/com/twitter/finagle/redis/protocol/commands/Sentinels.scala
thirstycrow: I think it's better not put them together.

    case class SentinelMasters() extends Sentinel(SentinelMasters.channelBuffer, Nil)

    object SentinelMasters extends SentinelHelper {
      val command = "MASTERS"
      def apply(args: Seq[Array[Byte]]): SentinelMasters = {
        new SentinelMasters()
      }
    }

mosesn: case classes with no parameters were deprecated, so we should find some kind of solution to this–if it really bothers you to put them under the same object (I think it's fine) we can rename the object in this case.

finagle-redis/src/test/scala/com/twitter/finagle/redis/commands/sentinel/SentinelClientIntegrationSuite.scala
mosesn: yes, that's a good idea

finagle-redis/src/main/scala/com/twitter/finagle/redis/protocol/commands/Sentinels.scala
thirstycrow: These are subcommands of "SENTINEL", which is added in Command.scala. I followed the "CONFIG" command's style here.

finagle-redis/src/main/scala/com/twitter/finagle/redis/protocol/commands/Sentinels.scala
mosesn: So the main places where this command are used seem to be in stats and tracing.  Do you want them to appear as "SENTINEL" or "MASTER" or "SENTINEL MASTER"?  In Config, it seems that we chose "CONFIG".

finagle-redis/src/main/scala/com/twitter/finagle/redis/SentinelClient.scala
thirstycrow: Is it ok with the following piece of code?

    def apply(host: String): SentinelClient = {
      val client = com.twitter.finagle.Redis.client
      SentinelClient(
        client
          .configured(client.params[DefaultPool.Param].copy(high = 1))
          // Daemonize is private to the builder package.
          // Is there another way to set this? Or can we just ignore it?
          // .configured(client.params[Daemonize](true))
          .newService(host))
    }

mosesn: Daemonize is already true by default in the finagle 6 API.

there's  no need to configure DefaultPool in the finagle 6 API because it uses the pipelining pool, so there will only ever be one outstanding connection anyway.

LGTM otherwise!

finagle-redis/src/main/scala/com/twitter/finagle/redis/protocol/commands/Sentinels.scala
thirstycrow: abstract class Config(sub: ChannelBuffer, args: Seq[ChannelBuffer]) extends Command {
      // I think you're mentioning this one.
      def command = Commands.CONFIG
      def toChannelBuffer = RedisCodec.toUnifiedFormat(Seq(CommandBytes.CONFIG, sub) ++ args)
    }

    case class ConfigResetStat() extends Config(sub = ConfigResetStat.channelBuffer, args = Seq())
    object ConfigResetStat extends ConfigHelper {
      // rather than this one.
      val command = "RESETSTAT" // <-- Where is it used?
      def apply(args: Seq[Array[Byte]]): ConfigResetStat = new ConfigResetStat()
    }

finagle-redis/src/main/scala/com/twitter/finagle/redis/protocol/commands/Sentinels.scala
thirstycrow: They both have a "command" property, with different values. Instead of using a different name, can we just use a dummy param?

    case class SentinelMasters(dummy: Int = 0) extends Sentinel(SentinelMasters.channelBuffer, Nil)

    object SentinelMasters extends SentinelHelper {
      val command = "MASTERS"
      def apply(args: Seq[Array[Byte]]): SentinelMasters = {
        new SentinelMasters()
      }
    }

finagle-redis/src/main/scala/com/twitter/finagle/redis/protocol/commands/Sentinels.scala
mosesn: yep, you're right, I was confused.

finagle-redis/src/main/scala/com/twitter/finagle/redis/protocol/commands/Sentinels.scala
mosesn: that seems even sillier than case class with no constructor args.  how about you leave it this way and we'll change it when we change the other case classes with no constructor args.

finagle-redis/src/main/scala/com/twitter/finagle/redis/SentinelClient.scala
```diff
@@ -0,0 +1,71 @@
...
+    val port: Int = props("port").toInt
+    val runid: String = props("runid")
+    val flags: Seq[String] = props("flags").split(",")
+    val pendingCommands: Int = props("pending-commands").toInt
+    val lastPingSent: Int = props("last-ping-sent").toInt
+    val lastPingReply: Int = props("last-ping-reply").toInt
+    val downAfterMilliseconds: Int = props("down-after-milliseconds").toInt
+  }
+  
+  trait DataNode extends Node {
```
taylorleese: This should probably be sealed as well.

finagle-redis/src/main/scala/com/twitter/finagle/redis/protocol/commands/Sentinels.scala
```diff
@@ -0,0 +1,169 @@
...
+  val command = "SET"
+  def apply(args: Seq[String]): SentinelSet = {
+    trimList(args, 2, "SENTINEL SET") match {
+      case name :: option :: value :: Nil =>
+        new SentinelSet(name, option, value)
+    }
+  }
+}
+
+trait SentinelHelper {
```
taylorleese: Probably should add sealed here as well.

finagle-redis/src/main/scala/com/twitter/finagle/redis/protocol/commands/Sentinels.scala
```diff
@@ -0,0 +1,169 @@
...
+        args.map(StringToChannelBuffer(_)))
+}
+
+object Sentinel {
+
+  val subCommands: Seq[SentinelHelper] = Seq(
+    SentinelMaster, SentinelMasters, SentinelMonitor, SentinelRemove, SentinelSet)
+
+  def fromBytes(args: Seq[Array[Byte]]): Sentinel = {
+    apply(args.map(new String(_)))
```
taylorleese: This is going to use the platform's default charset when constructing the string. I'm not sure if this is an actual problem in this scenario or not, but it would likely be best to make it explicit (e.g. UTF-8).

mahakp:
lgtm

mosesn:
Thanks @mahakp!  Let's get this out the door, we can do more work with adding native clustering support after we merge this in.  Thanks for your patience @thirstycrow!

mosesn:
This was merged internally.  Thanks for your patience!

mosesn:
Again, really appreciate your patience with us @thirstycrow.  Now in develop: https://github.com/twitter/finagle/commit/238f98dfb126700d71c221186a71a4a026b4dd45

PR #446, opened 2015-12-04
finagle-core: implementing RequestMeterFilter based on new AsyncMeter…

amartinsn:
Problem

We wanted to be able to control access to rate limited resources, by slowing down requests to fit on the request rate accepted by these resources.

Solution

We ended up implementing a filter named `RequestMeterFilter` using Twitter util's `AsyncMeter` to achieve that. Feedback on naming are really welcomed.

Result

As a result, I've added this `Filter` so other users with the same needs can use this solution.

### DIFF
--- a/finagle-core/src/main/scala/com/twitter/finagle/filter/RequestMeterFilter.scala
+++ b/finagle-core/src/main/scala/com/twitter/finagle/filter/RequestMeterFilter.scala
@@ -1,15 +1,15 @@
 package com.twitter.finagle.filter
 
-import java.util.concurrent.RejectedExecutionException
-
 import com.twitter.concurrent.AsyncMeter
 import com.twitter.finagle.{Failure, Service, SimpleFilter}
 import com.twitter.util.{Future, Throw}
 
+import java.util.concurrent.RejectedExecutionException
+
 /**
   * A [[com.twitter.finagle.Filter]] that rate limits requests to a fixed rate over time by
   * using the [[com.twitter.concurrent.AsyncMeter]] implementation. It can be used for slowing
-  * down access to throttled resources. Requests that are unable to acquire a permit are failed
+  * down access to throttled resources. Requests that cannot be enqueued to await a permit are failed
   * immediately with a [[com.twitter.finagle.Failure]] that signals that the work was never done,
   * so it's safe to reenqueue.
   *
--- a/finagle-core/src/main/scala/com/twitter/finagle/filter/RequestMeterFilter.scala
+++ b/finagle-core/src/main/scala/com/twitter/finagle/filter/RequestMeterFilter.scala
@@ -1,5 +1,7 @@
 package com.twitter.finagle.filter
 
+import java.util.concurrent.RejectedExecutionException
+
 import com.twitter.concurrent.AsyncMeter
 import com.twitter.finagle.{Failure, Service, SimpleFilter}
 import com.twitter.util.{Future, Throw}
@@ -8,13 +10,23 @@ import com.twitter.util.{Future, Throw}
   * A [[com.twitter.finagle.Filter]] that rate limits requests to a fixed rate over time by
   * using the [[com.twitter.concurrent.AsyncMeter]] implementation. It can be used for slowing
   * down access to throttled resources. Requests that are unable to acquire a permit are failed
-  * immediately with a [[com.twitter.finagle.Failure]] that signals a restartable or
-  * idempotent process.
+  * immediately with a [[com.twitter.finagle.Failure]] that signals that the work was never done,
+  * so it's safe to reenqueue.
+  *
+  * NOTE: If you're just trying not to be overwhelmed, you almost certainly want to use
+  * [[com.twitter.finagle.filter.RequestSemaphoreFilter]] instead, because RequestMeterFilter
+  * doesn't work well with "real" resources that are sometimes faster or slower (like a service
+  * that you're depending on that sometimes slows when it takes bursty traffic). This is better for
+  * resources that are artificially bounded, like a rate-limited API.
   */
-class RequestMeterFilter[Req, Rep](meter: AsyncMeter, permits: Int = 1) extends SimpleFilter[Req, Rep] {
+class RequestMeterFilter[Req, Rep](meter: AsyncMeter) extends SimpleFilter[Req, Rep] {
   def apply(request: Req, service: Service[Req, Rep]) = {
-    meter.await(permits).transform {
-      case Throw(noPermit) => Future.exception(Failure.rejected(noPermit))
+    meter.await(1).transform {
+      case Throw(noPermit) => noPermit match {
+        case e: RejectedExecutionException =>
+          Future.exception(Failure.rejected(noPermit))
+        case e => Future.exception(e)
+      }
       case _ => service(request)
     }
   }
--- a/finagle-core/src/main/scala/com/twitter/finagle/filter/RequestMeterFilter.scala
+++ b/finagle-core/src/main/scala/com/twitter/finagle/filter/RequestMeterFilter.scala
@@ -1,0 +1,21 @@
+package com.twitter.finagle.filter
+
+import com.twitter.concurrent.AsyncMeter
+import com.twitter.finagle.{Failure, Service, SimpleFilter}
+import com.twitter.util.{Future, Throw}
+
+/**
+  * A [[com.twitter.finagle.Filter]] that rate limits requests to a fixed rate over time by
+  * using the [[com.twitter.concurrent.AsyncMeter]] implementation. It can be used for slowing
+  * down access to throttled resources. Requests that are unable to acquire a permit are failed
+  * immediately with a [[com.twitter.finagle.Failure]] that signals a restartable or
+  * idempotent process.
+  */
+class RequestMeterFilter[Req, Rep](meter: AsyncMeter, permits: Int = 1) extends SimpleFilter[Req, Rep] {
+  def apply(request: Req, service: Service[Req, Rep]) = {
+    meter.await(permits).transform {
+      case Throw(noPermit) => Future.exception(Failure.rejected(noPermit))
+      case _ => service(request)
+    }
+  }
+}
--- a/finagle-core/src/test/scala/com/twitter/finagle/filter/RequestMeterFilterTest.scala
+++ b/finagle-core/src/test/scala/com/twitter/finagle/filter/RequestMeterFilterTest.scala
@@ -1,16 +1,17 @@
 package com.twitter.finagle.filter
 
-import java.util.concurrent.RejectedExecutionException
-
 import com.twitter.concurrent.AsyncMeter
 import com.twitter.conversions.time._
 import com.twitter.finagle.{Failure, Service}
 import com.twitter.util._
+
+import java.util.concurrent.RejectedExecutionException
+
 import org.junit.runner.RunWith
+import org.mockito.Mockito.when
 import org.scalatest.FunSuite
 import org.scalatest.junit.JUnitRunner
 import org.scalatest.mock.MockitoSugar
-import org.mockito.Mockito.when
 
 @RunWith(classOf[JUnitRunner])
 class RequestMeterFilterTest extends FunSuite with MockitoSugar {
@@ -22,8 +23,8 @@ class RequestMeterFilterTest extends FunSuite with MockitoSugar {
   test("return service execution after getting a permit") {
     val timer = new MockTimer
     Time.withCurrentTimeFrozen { ctl =>
-      val meter = AsyncMeter.newMeter(1, 1 second, 1)(timer)
-      val svc = new RequestMeterFilter(meter) andThen echoSvc
+      val meter = AsyncMeter.perSecond(1, 1)(timer)
+      val svc = new RequestMeterFilter(meter).andThen(echoSvc)
 
       assert(Await.result(svc(1)) == 1)
     }
@@ -32,8 +33,8 @@ class RequestMeterFilterTest extends FunSuite with MockitoSugar {
   test("mark dropped requests as failed") {
     val timer = new MockTimer
     Time.withCurrentTimeFrozen { ctl =>
-      val meter = AsyncMeter.newMeter(1, 1 second, 1)(timer)
-      val svc = new RequestMeterFilter(meter) andThen echoSvc
+      val meter = AsyncMeter.perSecond(1, 1)(timer)
+      val svc = new RequestMeterFilter(meter).andThen(echoSvc)
 
       val f1 = svc(1)
       assert(f1.isDefined)
@@ -44,7 +45,7 @@ class RequestMeterFilterTest extends FunSuite with MockitoSugar {
       val f3 = svc(3)
       assert(f3.isDefined)
       val failure = intercept[Failure] { Await.result(f3, 5.seconds) }
-      assert(failure.getCause.getClass == classOf[RejectedExecutionException])
+      intercept[RejectedExecutionException] { throw failure.getCause }
 
       ctl.advance(1.second)
       timer.tick()
@@ -58,7 +59,7 @@ class RequestMeterFilterTest extends FunSuite with MockitoSugar {
     when(meter.await(1)).thenReturn(Future.exception(new RuntimeException("Error!")))
 
     Time.withCurrentTimeFrozen { ctl =>
-      val svc = new RequestMeterFilter(meter) andThen echoSvc
+      val svc = new RequestMeterFilter(meter).andThen(echoSvc)
 
       val f1 = svc(3)
       assert(f1.isDefined)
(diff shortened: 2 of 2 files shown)


finagle-core/src/main/scala/com/twitter/finagle/filter/RequestMeterFilter.scala
```diff
@@ -0,0 +1,21 @@
...
+
+import com.twitter.concurrent.AsyncMeter
+import com.twitter.finagle.{Failure, Service, SimpleFilter}
+import com.twitter.util.{Future, Throw}
+
+/**
+  * A [[com.twitter.finagle.Filter]] that rate limits requests to a fixed rate over time by
+  * using the [[com.twitter.concurrent.AsyncMeter]] implementation. It can be used for slowing
+  * down access to throttled resources. Requests that are unable to acquire a permit are failed
+  * immediately with a [[com.twitter.finagle.Failure]] that signals a restartable or
```
mosesn: maybe, "signals that the work was never done, so it's safe to reenqueue"?

finagle-core/src/main/scala/com/twitter/finagle/filter/RequestMeterFilter.scala
```diff
@@ -0,0 +1,21 @@
...
+  * idempotent process.
```
mosesn: let's also add a strongly-worded warning that if you're just trying not to be overwhelmed, you almost certainly want to use RequestSemaphoreFilter instead, because RequestMeterFilter doesn't work well with "real" resources that are sometimes faster or slower (like a service that you're depending on that sometimes slows when it takes bursty traffic).  this is better for resources that are artificially bounded, like a rate-limited API.

finagle-core/src/test/scala/com/twitter/finagle/filter/RequestMeterFilterTest.scala
```diff
@@ -0,0 +1,44 @@
...
+import com.twitter.finagle.{Failure, Service}
+import com.twitter.util.{Await, Future}
+import org.junit.runner.RunWith
+import org.scalatest.FunSuite
+import org.scalatest.junit.JUnitRunner
+
+@RunWith(classOf[JUnitRunner])
+class RequestMeterFilterTest extends FunSuite {
+
+  val timer = DefaultTimer.twitter
```
mosesn: use mocktimer instead

finagle-core/src/test/scala/com/twitter/finagle/filter/RequestMeterFilterTest.scala
```diff
@@ -0,0 +1,44 @@
...
+  val meter = AsyncMeter.newMeter(1, 1 second, 1)(timer)
+
+  test("return service execution after getting a permit") {
+    val echoSvc = new Service[Int, Int] {
```
mosesn: Time.withCurrentTimeFrozen is useful for manipulating time here

finagle-core/src/test/scala/com/twitter/finagle/filter/RequestMeterFilterTest.scala
```diff
@@ -0,0 +1,44 @@
...
+      def apply(req: Int) = Future(req)
+    }
+    val svc = new RequestMeterFilter(meter) andThen echoSvc
+    assert(Await.result(svc(2)) == 2)
+  }
+
+  test("mark dropped requests as rejected") {
+    val neverSvc = new Service[Int, Int] {
```
mosesn: Time.withCurrentTimeFrozen

finagle-core/src/test/scala/com/twitter/finagle/filter/RequestMeterFilterTest.scala
```diff
@@ -0,0 +1,44 @@
...
+      def apply(req: Int) = Future.never
+    }
+    val svc = new RequestMeterFilter(meter, 2) andThen neverSvc
+    val f = intercept[Failure] { Await.result(svc(1)) }
+    assert(f.isFlagged(Failure.Restartable))
+  }
+
+  test("service failures are not wrapped as rejected") {
+    val exc = new Exception("app exc")
```
mosesn: Time.withCurrentTimeFrozen

finagle-core/src/main/scala/com/twitter/finagle/filter/RequestMeterFilter.scala
```diff
@@ -0,0 +1,21 @@
...
+  */
+class RequestMeterFilter[Req, Rep](meter: AsyncMeter, permits: Int = 1) extends SimpleFilter[Req, Rep] {
+  def apply(request: Req, service: Service[Req, Rep]) = {
+    meter.await(permits).transform {
```
mosesn: it might make sense not to mark it as rejected if you get an IllegalArgumentException, or a CancellationException?  Probably only RejectedExecutionException should be marked as rejected.

mosesn:
LGTM from a high level, let's hammer out the details!

finagle-core/src/main/scala/com/twitter/finagle/filter/RequestMeterFilter.scala
```diff
@@ -0,0 +1,21 @@
...
+import com.twitter.util.{Future, Throw}
+
+/**
+  * A [[com.twitter.finagle.Filter]] that rate limits requests to a fixed rate over time by
+  * using the [[com.twitter.concurrent.AsyncMeter]] implementation. It can be used for slowing
+  * down access to throttled resources. Requests that are unable to acquire a permit are failed
+  * immediately with a [[com.twitter.finagle.Failure]] that signals a restartable or
+  * idempotent process.
+  */
+class RequestMeterFilter[Req, Rep](meter: AsyncMeter, permits: Int = 1) extends SimpleFilter[Req, Rep] {
```
mosesn: now that I think about it, how about we don't make permits configurable?  the point of having a permits argument in await is for folks who want to sometimes await on a different number of permits–since we always await on the same number, it's fine to just pick 1, imo.

amartinsn: Yes, initially it was not configurable. Only did that for testing purposes. Now using MockTimer I might revert that. Ta.

finagle-core/src/test/scala/com/twitter/finagle/filter/RequestMeterFilterTest.scala
amartinsn: Not sure I got the idea of using `MockTimer`. It proceeds with `svc(1)` succesfully but when it calls `svc(2)` it gets stuck on the `meter.await(1)` part. It hangs forever...

```scala
  test("mark dropped requests as rejected") {
    val timer = new MockTimer
    val neverSvc = new Service[Int, Int] {
      def apply(req: Int) = Future.never
    }
    Time.withCurrentTimeFrozen { ctl =>
      val meter = AsyncMeter.newMeter(1, 1 second, 1)(timer)
      val svc = new RequestMeterFilter(meter) andThen neverSvc
      svc(1)

      val f = intercept[Failure] { Await.result(svc(2)) }
      assert(f.isFlagged(Failure.Restartable))

      ctl.advance(1 second)
      timer.tick()
    }
  }
```

finagle-core/src/main/scala/com/twitter/finagle/filter/RequestMeterFilter.scala
amartinsn: But we should wrap it in a Failure anyway right? So it can retried? If so, `Future.exception(Failure.wrap(noPermit))` would do the trick, right?

finagle-core/src/main/scala/com/twitter/finagle/filter/RequestMeterFilter.scala
amartinsn: Didn't get what you mean by "resources that are artificially bounded".

mosesn: suppose you're talking to the foursquare API.  they can handle more traffic than you're throwing at them, but as a policy, they have a ratelimit of N / second.  so you want to rate limit in order to protect them–if you were talking to a service that you owned, it could be that you want to send them traffic as fast as they can handle it, but within reason, ie not more than fifty concurrent requests.  in the foursquare example, they're artificially throttling–in the example of your service, it will be able to speed up or slow down as your service is fast or slow.

finagle-core/src/test/scala/com/twitter/finagle/filter/RequestMeterFilterTest.scala
mosesn: NB it hangs at Await.result(svc(2)) which is a blocking call.  You need to advance it before that.

finagle-core/src/main/scala/com/twitter/finagle/filter/RequestMeterFilter.scala
mosesn: Well, you wouldn't want to retry it if it had been cancelled or you had passed in bad arguments.  Failure is also mostly used for signaling inside of finagle, so not sure what wrapping in `Failure` buys you.

finagle-core/src/test/scala/com/twitter/finagle/filter/RequestMeterFilterTest.scala
amartinsn: The idea of this test is to make the meter throw an exception because I'm trying to acquire more than 1 permit within a second, and check the exception type, etc. I've changed it to advance after the first call to meter. I've tried advancing a couple hundred millisecs but it still hangs. Only if I advance 1 sec it stops hanging, but it doesn't behave the way I expect, as it doesn't throw any exception. What am I missing? Thanks... 

```scala
val timer = new MockTimer
Time.withCurrentTimeFrozen { ctl =>
  val meter = AsyncMeter.newMeter(1, 1 second, 1)(timer)
  meter.await(1)

  ctl.advance(1 second)
  timer.tick()

  Await.result(meter.await(1))
  println("done!")
}
```

finagle-core/src/main/scala/com/twitter/finagle/filter/RequestMeterFilter.scala
amartinsn: Yes you are right. The client is already controlling the request rate, so if it goes beyond that rate, it should throw an exception. Will change that.

finagle-core/src/test/scala/com/twitter/finagle/filter/RequestMeterFilterTest.scala
mosesn: So the way that AsyncMeter works is that it allows `burst` through at a time–if you examine the first future, it has already been satisfied.  So only the second one is waiting, which means that there is only one waiter.  It would be incorrect for it to throw.  If instead you do:

```scala
val timer = new MockTimer
Time.withCurrentTimeFrozen { ctl =>
  val meter = AsyncMeter.newMeter(1, 1 second, 1)(timer)
  val f1 = meter.await(1)
  assert(f1.isDone)
  val f2 = meter.await(1)
  assert(!f2.isDefined)
  val f3 = meter.await(1)
  assert(f3.isDefined)
  intercept[RejectedExecutionException] { Await.result(f3, 5.seconds) }
  ctl.advance(1.second)
  timer.tick()
  assert(f2.isDone)
}
```

In fact, we have tests like this for [AsyncMeter][0] already.

[0]: https://github.com/twitter/util/blob/develop/util-core/src/test/scala/com/twitter/concurrent/AsyncMeterTest.scala#L53

amartinsn: Pushed earlier today a new version.

finagle-core/src/main/scala/com/twitter/finagle/filter/RequestMeterFilter.scala
```diff
@@ -0,0 +1,33 @@
...
+import java.util.concurrent.RejectedExecutionException
+
+import com.twitter.concurrent.AsyncMeter
+import com.twitter.finagle.{Failure, Service, SimpleFilter}
+import com.twitter.util.{Future, Throw}
+
+/**
+  * A [[com.twitter.finagle.Filter]] that rate limits requests to a fixed rate over time by
+  * using the [[com.twitter.concurrent.AsyncMeter]] implementation. It can be used for slowing
+  * down access to throttled resources. Requests that are unable to acquire a permit are failed
```
mosesn: Let's be precise here: "Requests that cannot be enqueued to await a permit are failed immediately"

finagle-core/src/test/scala/com/twitter/finagle/filter/RequestMeterFilterTest.scala
```diff
@@ -0,0 +1,84 @@
...
+
+  val echoSvc = new Service[Int, Int] {
+    def apply(req: Int) = Future(req)
+  }
+
+  test("return service execution after getting a permit") {
+    val timer = new MockTimer
+    Time.withCurrentTimeFrozen { ctl =>
+      val meter = AsyncMeter.newMeter(1, 1 second, 1)(timer)
+      val svc = new RequestMeterFilter(meter) andThen echoSvc
```
mosesn: `.andThen(echoSvc)` here and elsewhere (infix style is only for symbolic operators)

finagle-core/src/test/scala/com/twitter/finagle/filter/RequestMeterFilterTest.scala
```diff
@@ -0,0 +1,84 @@
...
+class RequestMeterFilterTest extends FunSuite with MockitoSugar {
+
+  val echoSvc = new Service[Int, Int] {
+    def apply(req: Int) = Future(req)
+  }
+
+  test("return service execution after getting a permit") {
+    val timer = new MockTimer
+    Time.withCurrentTimeFrozen { ctl =>
+      val meter = AsyncMeter.newMeter(1, 1 second, 1)(timer)
```
mosesn: feel free to ignore: `AsyncMeter.perSecond(1, 1)` here and elsewhere

finagle-core/src/test/scala/com/twitter/finagle/filter/RequestMeterFilterTest.scala
```diff
@@ -0,0 +1,84 @@
...
+      val f1 = svc(1)
+      assert(f1.isDefined)
+
+      val f2 = svc(2)
+      assert(!f2.isDefined)
+
+      val f3 = svc(3)
+      assert(f3.isDefined)
+      val failure = intercept[Failure] { Await.result(f3, 5.seconds) }
+      assert(failure.getCause.getClass == classOf[RejectedExecutionException])
```
mosesn: ffti: maybe cleaner as:
```scala
intercept[RejectedExecutionException] {
  throw failure.getCause
}
```

finagle-core/src/main/scala/com/twitter/finagle/filter/RequestMeterFilter.scala
```diff
@@ -0,0 +1,33 @@
+package com.twitter.finagle.filter
+
+import java.util.concurrent.RejectedExecutionException
```
mosesn: this should go at the bottom of the imports, so that they're in sorted order

mosesn:
This is looking pretty good!

finagle-core/src/test/scala/com/twitter/finagle/filter/RequestMeterFilterTest.scala
amartinsn: I loved this syntax but it, actually I was looking for a nicer way of expressing this assertion, but it didn't work.

```scala
val failure = intercept[Failure] { Await.result(f3, 5.seconds) }
intercept[RejectedExecutionException] { failure.getCause }
```
Breaks the test with message:

`Expected exception java.util.concurrent.RejectedExecutionException to be thrown, but no exception was thrown.`

finagle-core/src/main/scala/com/twitter/finagle/filter/RequestMeterFilter.scala
amartinsn: Done

finagle-core/src/test/scala/com/twitter/finagle/filter/RequestMeterFilterTest.scala
amartinsn: Done

finagle-core/src/test/scala/com/twitter/finagle/filter/RequestMeterFilterTest.scala
amartinsn: Done

finagle-core/src/test/scala/com/twitter/finagle/filter/RequestMeterFilterTest.scala
amartinsn: Quite liked it! Was looking for a cleaner way of expressing this assertion! Thanks!

finagle-core/src/main/scala/com/twitter/finagle/filter/RequestMeterFilter.scala
amartinsn: Done

mosesn:
LGTM!  I'll try to get some more eyes on it too.

dschobel:
lgtm.

amartinsn:
Nice!!!!

adleong:

PR #447, opened 2015-12-08
pubsub support

thirstycrow:


finagle-redis/src/main/scala/com/twitter/finagle/redis/protocol/commands/PubSub.scala
```diff
@@ -0,0 +1,96 @@
...
+  def command: String
+  def listener: SubscribeListener
+}
+
+trait SubscribeListener {
+  def onException(ex: Throwable): Unit
+  def onMessage(msg: Reply): Unit
+}
+
+case class PSubscribe(patterns: Seq[ChannelBuffer], listener: SubscribeListener) extends SubscribeCommand {
```
thirstycrow: It looks strange to put the listener here in the command, but I don't know how to pass it to the dispatcher at construction time.

finagle-redis/src/main/scala/com/twitter/finagle/redis/PubSubCommands.scala
```diff
@@ -0,0 +1,40 @@
+package com.twitter.finagle.redis
+
+import com.twitter.finagle.redis.naggati.ProtocolError
```
mosesn: never used

finagle-redis/src/main/scala/com/twitter/finagle/redis/PubSubCommands.scala
```diff
@@ -0,0 +1,40 @@
...
+import com.twitter.finagle.redis.protocol._
+import com.twitter.finagle.redis.util.ReplyFormat
+import com.twitter.util.{ Future, Time }
```
mosesn: let's write this as {Future, Time}

finagle-redis/src/main/scala/com/twitter/finagle/redis/PubSubCommands.scala
```diff
@@ -0,0 +1,40 @@
...
+import org.jboss.netty.buffer.ChannelBuffer
```
mosesn: merge me!

finagle-redis/src/main/scala/com/twitter/finagle/redis/SubscribeClient.scala
```diff
@@ -0,0 +1,108 @@
+package com.twitter.finagle.redis
+
+import com.twitter.finagle.Service
```
mosesn: sort imports

finagle-redis/src/main/scala/com/twitter/finagle/redis/SubscribeClient.scala
```diff
@@ -0,0 +1,108 @@
...
+
+  /**
+   * Construct a sentinel client from a single host.
+   * @param host a String of host:port combination.
+   */
+  def apply(host: String): SubscribeClient = {
+    new SubscribeClient(com.twitter.finagle.Redis.Subscribe.newService(host))
+  }
+
+  object MessageBytes {
```
mosesn: ffti: no need to nest these, we can make this top-level

finagle-redis/src/main/scala/com/twitter/finagle/redis/SubscribeClient.scala
```diff
@@ -0,0 +1,108 @@
...
+    val SUBSCRIBE = StringToChannelBuffer("subscribe")
```
mosesn: I think the style is `Subscribe`, no?

finagle-redis/src/main/scala/com/twitter/finagle/redis/SubscribeClient.scala
```diff
@@ -0,0 +1,108 @@
...
+import com.twitter.util.TimerTask
+import com.twitter.finagle.ServiceClosedException
+
+object SubscribeClient {
+
+  type MessageHandler = (ChannelBuffer, ChannelBuffer) => Unit
+  type PMessageHandler = (ChannelBuffer, ChannelBuffer, ChannelBuffer) => Unit
+
+  /**
+   * Construct a sentinel client from a single host.
```
mosesn: should be Subscribe client

finagle-redis/src/main/scala/com/twitter/finagle/redis/SubscribeClient.scala
```diff
@@ -0,0 +1,108 @@
...
+  }
+}
+
+class SubscribeClient(service: Service[SubscribeCommand, Unit]) extends SubscribeListener {
+
+  import SubscribeClient._
+
+  def subscribe(channels: Seq[ChannelBuffer])(handler: MessageHandler): Future[Unit] =
+    doRequest(Subscribe(channels, this))
+      .onSuccess(_ => channels.foreach(msgHandlers.put(_, handler)))
```
mosesn: instead of `onSuccess`, should we use `ensure` here instead?

finagle-redis/src/main/scala/com/twitter/finagle/redis/SubscribeClient.scala
```diff
@@ -0,0 +1,108 @@
...
+      .onSuccess(_ => channels.foreach(msgHandlers.remove(_)))
+
+  def pUnsubscribe(patterns: Seq[ChannelBuffer]): Future[Unit] =
+    doRequest(PUnsubscribe(patterns, this))
+      .onSuccess(_ => patterns.foreach(pMsgHandlers.remove(_)))
+
+  /**
+   * Releases underlying service object
+   */
+  def release() = {
```
mosesn: let's use `close` and `Closable` here instead.

finagle-redis/src/main/scala/com/twitter/finagle/redis/SubscribeClient.scala
```diff
@@ -0,0 +1,108 @@
...
+    service.close()
+  }
+
+  def doRequest(cmd: SubscribeCommand) = {
```
mosesn: let's not expose this publicly: `private[this]`?

finagle-redis/src/main/scala/com/twitter/finagle/redis/SubscribeClient.scala
```diff
@@ -0,0 +1,108 @@
...
+    service(cmd)
+  }
+
+  import scala.collection.mutable.HashMap
```
mosesn: don't love imports in the middle of the file.  can we put these at the top?

finagle-redis/src/main/scala/com/twitter/finagle/redis/SubscribeClient.scala
```diff
@@ -0,0 +1,108 @@
...
+  import com.twitter.finagle.redis.protocol._
+  import com.twitter.finagle.redis.protocol.{ Subscribe => SubscribeCmd }
+
+  private val msgHandlers = HashMap[ChannelBuffer, MessageHandler]()
```
mosesn: private[this]?

finagle-redis/src/main/scala/com/twitter/finagle/redis/SubscribeClient.scala
```diff
@@ -0,0 +1,108 @@
...
+  private val pMsgHandlers = HashMap[ChannelBuffer, PMessageHandler]()
```
mosesn: ditto

finagle-redis/src/main/scala/com/twitter/finagle/redis/SubscribeClient.scala
```diff
@@ -0,0 +1,108 @@
...
+
+  def onException(e: Throwable) {
```
mosesn: be explicit here about the type you're returning: `: Unit = {`.  Are you sure you don't want to return `: Future[Unit]`?

finagle-redis/src/main/scala/com/twitter/finagle/redis/SubscribeClient.scala
```diff
@@ -0,0 +1,108 @@
...
+    val sub = if (msgHandlers.nonEmpty) doRequest(Subscribe(msgHandlers.keys.toSeq, this)) else Future.Done
+    val psub = if (pMsgHandlers.nonEmpty) doRequest(PSubscribe(pMsgHandlers.keys.toSeq, this)) else Future.Done
+    Futures.join(sub, psub).onFailure {
+      case sce: ServiceClosedException =>
+        // Service closed, do nothing
+      case ex =>
+        DefaultTimer.twitter.schedule(Time.now + 1.second)(onException(e))
```
mosesn: the timer should be configurable

finagle-redis/src/main/scala/com/twitter/finagle/redis/SubscribeClient.scala
```diff
@@ -0,0 +1,108 @@
...
+      case MBulkReply(BulkReply(MessageBytes.PMESSAGE) :: BulkReply(pattern) :: BulkReply(channel) :: BulkReply(message) :: Nil) =>
+        pMsgHandlers.get(pattern).map(_(pattern, channel, message))
+      case MBulkReply(BulkReply(tpe) :: BulkReply(channel) :: IntegerReply(count) :: Nil) =>
+        tpe match {
+          case MessageBytes.PSUBSCRIBE   =>
+          case MessageBytes.PUNSUBSCRIBE =>
+          case MessageBytes.SUBSCRIBE    =>
+          case MessageBytes.UNSUBSCRIBE  =>
+          case _ =>
+            throw new IllegalArgumentException()
```
mosesn: let's pass a useful string to IllegalArgumentException, here and elsewhere

finagle-redis/src/main/scala/com/twitter/finagle/redis/SubscribeClient.scala
```diff
@@ -0,0 +1,108 @@
...
+  }
+
+  def onMessage(message: Reply): Unit = {
+    message match {
+      case MBulkReply(BulkReply(MessageBytes.MESSAGE) :: BulkReply(channel) :: BulkReply(message) :: Nil) =>
+        msgHandlers.get(channel).map(_(channel, message))
+      case MBulkReply(BulkReply(MessageBytes.PMESSAGE) :: BulkReply(pattern) :: BulkReply(channel) :: BulkReply(message) :: Nil) =>
+        pMsgHandlers.get(pattern).map(_(pattern, channel, message))
+      case MBulkReply(BulkReply(tpe) :: BulkReply(channel) :: IntegerReply(count) :: Nil) =>
+        tpe match {
```
mosesn: why do we not want to do anything here?

finagle-redis/src/main/scala/com/twitter/finagle/redis/dispatch/SubscribeDispatcher.scala
```diff
@@ -0,0 +1,38 @@
...
+import com.twitter.finagle.dispatch.GenSerialClientDispatcher
+import com.twitter.finagle.redis.protocol._
+import com.twitter.finagle.transport.Transport
+import com.twitter.util.{Future, Promise}
+import java.util.concurrent.atomic.AtomicReference
+
+class SubscribeDispatcher(trans: Transport[SubscribeCommand, Reply])
+    extends GenSerialClientDispatcher[SubscribeCommand, Unit, SubscribeCommand, Reply](trans) {
+
+  private var listener = new AtomicReference[SubscribeListener]
```
mosesn: should be val

finagle-redis/src/main/scala/com/twitter/finagle/redis/protocol/commands/PubSub.scala
```diff
@@ -0,0 +1,96 @@
...
+import com.twitter.finagle.redis.ClientError
+import com.twitter.finagle.redis.protocol.Commands._
+import com.twitter.finagle.redis.util._
+
+abstract class SubscribeCommand extends RedisMessage {
+  def command: String
+  def listener: SubscribeListener
+}
+
+trait SubscribeListener {
```
mosesn: let's move this closer to the implementation–I had to do quite a bit of looking to find it.  also, let's not call it SubscribeListener, since we use Listeners for things that listen for connections and then respond to requests (ie Netty3Listener).  This is similar, but not exactly the same thing.  Maybe call it something different?  Handler?

finagle-redis/src/main/scala/com/twitter/finagle/Redis.scala
```diff
@@ -63,4 +66,48 @@ object Redis extends Client[Command, Reply] {
 
   def newService(dest: Name, label: String): Service[Command, Reply] =
     client.newService(dest, label)
+
+  object Subscribe extends com.twitter.finagle.Client[SubscribeCommand, Unit] {
```
mosesn: could you document why we need a special subscription client, here and elsewhere?  presumably it's because finagle doesn't have native support for object-streaming-style computation.  it's also worth noting that if instead of a Service[Req, Rep] we had a Client that could return a Closable AsyncStream, that would probably be better, and it's something we should probably target as we try to retool finagle to support more streaming work.

mosesn:
I don't *super* love this approach (I'd rather expose an actual streaming abstraction) but given finagle's poor streaming support right now, this is probably our best bet.

The only other way I can think of doing this would be to mark the redis command a sealed trait, and check whether it's a subscription-style or non-subscription-style command.

finagle-redis/src/main/scala/com/twitter/finagle/Redis.scala
```diff
@@ -63,4 +66,48 @@ object Redis extends Client[Command, Reply] {
...
+        param.ProtocolLibrary("redis.subscribe")
+
+      /**
+       * A default client stack which supports the pipelined redis client.
+       */
+      def newStack: Stack[ServiceFactory[SubscribeCommand, Unit]] = StackClient.newStack
+        .replace(DefaultPool.Role, SingletonPool.module[SubscribeCommand, Unit])
+    }
+
+    case class SubscribeClient(
```
mosesn: actually, do we need a separate client at this layer?  could we type it at the other client layer?

finagle-redis/src/main/scala/com/twitter/finagle/redis/SubscribeClient.scala
mosesn: these maps need to be synchronized, or else support concurrent access (à la ConcurrentHashMap)–as it is, it's not threadsafe.

finagle-redis/src/main/scala/com/twitter/finagle/redis/dispatch/SubscribeDispatcher.scala
```diff
@@ -0,0 +1,38 @@
...
+    }
+    .onFailure { ex =>
+      Option(listener.get()).map(_.onException(ex))
+    }
+
+  protected def dispatch(req: SubscribeCommand, p: Promise[Unit]): Future[Unit] =
+    trans.write(req).onSuccess { _ => p.setDone() }
+
+  override def apply(req: SubscribeCommand): Future[Unit] = {
+    listener.compareAndSet(null, req.listener)
```
mosesn: does this mean that dispatchers must be discarded after each use?

mosesn: can we add a test to check that we're behaving correctly here?  it would be a bummer to just silently keep using the old listener.

finagle-redis/src/main/scala/com/twitter/finagle/redis/protocol/Command.scala
```diff
@@ -422,8 +441,8 @@ class CommandCodec extends UnifiedProtocolCodec {
 
   val log = Logger(getClass)
 
-  val encode = new Encoder[Command] {
-    def encode(obj: Command) = Some(obj.toChannelBuffer)
+  val encode = new Encoder[RedisMessage] {
```
mosesn: could you mention in CHANGES that you've added support for pubsub, and also that you're changing these APIs?

finagle-redis/src/main/scala/com/twitter/finagle/redis/protocol/commands/PubSub.scala
mosesn: hmmm, would you be OK with doing it at construction-time?  or do you think it's important to be able to do it request-time?

finagle-redis/src/main/scala/com/twitter/finagle/redis/SubscribeClient.scala
thirstycrow: Upper case is used in CommandBytes, so I followed the same style here. MessageBytes is used only in SubscribeClient, so I made it private, and it would be safe to rename them later.

BTW: Sometimes, I find it hard to decide the case the name of a val should be in. What are the conventions are you follow in twitter?

finagle-redis/src/main/scala/com/twitter/finagle/redis/SubscribeClient.scala
thirstycrow: Why should we register a handler, if the subscription is not succeeded?

finagle-redis/src/main/scala/com/twitter/finagle/redis/SubscribeClient.scala
thirstycrow: I found during testing that the acknowledgement message may come after messages for the subscribed channel. So the handlers are registered when the subscription request is sent successfully, instead of waiting for the acknowledgement messages. I added some comments in the code to make it look better.

finagle-redis/src/main/scala/com/twitter/finagle/redis/dispatch/SubscribeDispatcher.scala
thirstycrow: All commands created by the same SubscribeClient instance carries the same listener, which is the SubscribeClient instance itself. It would be better if the SubscribeClient can be passed as a constructor argument for the dispatcher.

finagle-redis/src/main/scala/com/twitter/finagle/Redis.scala
thirstycrow: I don't quite understand what you mean.

finagle-redis/src/main/scala/com/twitter/finagle/redis/protocol/Command.scala
thirstycrow: Updated. I don't figure out what a 'RB_ID' is, so I did not add them, is that ok?

finagle-redis/src/main/scala/com/twitter/finagle/redis/SubscribeClient.scala
mosesn: finagle-redis was an external contribution, so it uses a slightly different style.  we usually try to adhere to the scala style guide, although we have a few different rules, which aren't all published externally unfortunately.  probably best to follow the style of finagle-redis here, so just ignore my comment.

finagle-redis/src/main/scala/com/twitter/finagle/redis/SubscribeClient.scala
mosesn: good point :+1:

finagle-redis/src/main/scala/com/twitter/finagle/Redis.scala
mosesn: my understanding is that we need a redis client when we create the rich subscribe client with all the fancy subscription methods, but I think we might be able to still use a regular redis client, and might not need the special subscribe redis client.

finagle-redis/src/main/scala/com/twitter/finagle/redis/dispatch/SubscribeDispatcher.scala
mosesn: ahh, I see. could you add a comment?

finagle-redis/src/main/scala/com/twitter/finagle/redis/protocol/Command.scala
mosesn: yes, RB_ID is only for internal changes.

finagle-redis/src/main/scala/com/twitter/finagle/Redis.scala
thirstycrow: It is necessary because when a client enters the pub/sub state, it can no longer issue commands other than the (un)subscribe ones, until it exits from the pub/sub state. We have to create two clients if we want to use pub/sub and still be able to access the data. I added some comments for this on c.t.f.redis.SubscribeClient.

finagle-redis/src/main/scala/com/twitter/finagle/Redis.scala
thirstycrow: Comments add on c.t.f.redis.SubscribeClient

finagle-redis/src/main/scala/com/twitter/finagle/Redis.scala
mosesn: could we still use clients of the same type, and just document this?

thirstycrow: In that case, we'll have to keep track of the client state, and prevent users from issuing ordinary commands to a pub/sub client by throwing IllegalStateException. I think it's better to keep them separate for type safety.

mosesn: Don't we already have that problem today in that someone could call two subscribe commands?  Seems like we need to document it anyway (shrug)

thirstycrow: That's not the same. We may subscribe multiple times, but it will not change the way how we communicate with the server. In the subscribe mode, there's no correspondence between the outgoing and incoming messages, while in the ordinary mode, the outgoing and incoming messages are in one-by-one correspondence (so pipelining works). I think it's more appropriate to consider them two different protocols, despite that they encode messages in the same manner. How about we pull up Redis.Subscribe, rename it to RedisSubscribe, add some methods to create rich subscribe clients, and document the designing considerations in its scaladoc? We could also add a README for finagle-redis to provide some get-started examples.

mosesn: I think the right way of doing this would be to keep using the same redis protocol, but ensure that if a connection is subscribed to, then it can't be used for pipelining, and vice versa.  This way we actually can share the client between subscription and normal operations.

This will also be generally useful because we need to support this behavior for redis transactions too.

thirstycrow: Could you give some details about the transaction affairs? I thought it had no relation with the subscribe commands, but I'm not pretty sure on that.

thirstycrow: Some more thoughts on adding subscribe support to the existing pipelining client:

1. It is hard to tell whether a client has quit from pub/sub mode successfully, because we are not replied when we issue an unsubscribe command. When we issued  an unsubscribe command to the last channel/pattern we are currently subscribed to, we have to wait until the acknowledgement message for this command arrived, before we can confidently say that it is safe to send normal commands. But during the waiting gap, it is possible that some other subscribe command are issued. So, the acknowledgement message carries some information about the connection status at the time it is sent out from the server, which may not still be true at the time it is received by the client. To solve this problem, we'll have to provide some special unsubscribe command, so that when it is issued, the client put itself into a temporally-unavailable state, and reject any more command, neither normal nor subscribe, until it receives the unsubscribe acknowledgement.

2. We cannot manage the client pub/sub status transparently with sharding. We'll have to provider sharders specifically designed for redis, to keep track all of the pub/sub status of the underlying shards.

thirstycrow:
@mosesn I made some modifications to make it work well with load balancing. Please take a look at the new implementation. And is there anything still need to be done, except merging the subscribe client and the normal client?

finagle-redis/src/main/scala/com/twitter/finagle/Redis.scala
mosesn: Yes, transactions are unrelated to subscription, except in that when you're in a transaction, you can't use that connection for anything else, the same way that you can't use a subscription connection for anything else.

If we changed the connection pool API slightly, so that it was something like:

```scala
trait ReservingPool[Req, Rep] {
  // nobody else will use this connection until it's marked as no longer needed (ie closed)
  def reserve(): Service[Req, Rep]

  // may be reused
  def get(): Service[Req, Rep]
}
```

Then we could use it with redis and be guaranteed of not violating the protocol.

1.  When you say we are not replied when we issue an unsubscribe command, but we will receive an acknowledgment, do you mean the TCP ack, or does redis also provide an acknowledgment?  This seems pretty doable–we can have something which doesn't mark the reserved connection as "closed" until after we've received the acknowledgment.

2.  Yeah, we have tools like this for memcached (see the sharded multiget) but I'm a little hesitant to start building sharding into our clients.  It might make sense to make sharding helpers at first so that it's easy for folks to make sharders that work for their use case.  In the long run, it would be nice to have a toolkit that we can hook into any protocol that we need to shard manually (redis, mysql, memcached) but for now it probably makes sense to just make little custom helpers.

Another thing to consider is that a lot of the tooling finagle has is around naming, and balancing between hosts.  This isn't something that we can do in something like redis, because we only rarely consider different hosts to be the same (ie in redis cluster mode).

It might make sense to see what we can do to simplify the redis client by stripping the parts we don't need.

we should be able to remove:
both of the FactoryToService
BindingFactory
AddrMetadataExtraction
Loadbalancing
FailFast
FailureAccrual
DtabStatsFilter

Anyway, most of these are out of scope.  We should just set them as TODOs for future work.

mosesn:
this looks fine to me, but I think we might want to put this in an experimental package (like com.twitter.finagle.redis.exp), because I think we should change the model away from the the handlers one when we make StackClient a little more flexible (in the next few months, hopefully).

I think a reasonable list of TODOs for redis is:

1.  Stop using the mutable subscription clients once we have better streaming support–hand back a Closable AsyncStream.
2.  Clarify connection pooling for "reserved" connections vs "shared" connections.
3.  Remove extra finagle client stuff we don't need anymore
4.  Provide tooling for sharding (eg a "subscribe" method that scatter-gathers automatically).
5.  Support transactions

mosesn:
Actually, now that I think about it, one way we could simplify this would be if we provided a method that lets you choose whether you want the pipelined connection pool or the non-pipelined one when constructing a client.  Then you could use the pipelined connection pool for shared work (like get, set), and the non-pipelined one for reserved work (like transactions, subscriptions).  It's somewhat manual, but  it would get the job done, as long as we make sure not to close the returned service until we unsubscribe.

finagle-redis/src/main/scala/com/twitter/finagle/Redis.scala
thirstycrow: Subscription is more complicated than transactions. Consider the following scenario:
   
    send: subscribe foo      <-- The client subscribes to a channel,

    recv: subscribe foo 1    <-- and receives an acknowledgement like this. The
                                 number at the end is the number of channels the
                                 client subscribed to. The client will exit from
                                 pub/sub mode, when it drops down to 0.

    send: unsubscribe foo    <-- Not interested in this channel any more.

    send: subscribe bar      <-- Another subscription, before the acknowledgement
                                 message arrives.

    recv: unsubscribe foo 0  <-- It's not safe to reuse this connection. :(

    recv: subscribe bar 1

mosesn: oh, I see now.  we would need two kinds of reusable connections, one for subscriptions, and one for regular (non-transaction) redis.  but this should still be doable.

finagle-redis/src/main/scala/com/twitter/finagle/redis/dispatch/SubscribeDispatcher.scala
```diff
@@ -0,0 +1,70 @@
...
+        throw new IllegalArgumentException(s"Unexpected reply type: ${message.getClass.getSimpleName}")
+    }
+  }
+
+  private[this] def loop(): Unit =
+    trans.read().onSuccess { reply =>
+      handleMessage(reply)
+      loop()
+    }
+    .onFailure { ex =>
```
mosesn: this should be on the previous line, and let's wrap ex => in NonFatal(ex) =>

finagle-redis/src/main/scala/com/twitter/finagle/redis/dispatch/SubscribeDispatcher.scala
```diff
@@ -0,0 +1,70 @@
...
+            // detect something unexpected.
+          case _ =>
+            throw new IllegalArgumentException(s"Unsupported message type: ${tpe.toString(Charsets.Utf8)}")
+        }
+      case _ =>
+        throw new IllegalArgumentException(s"Unexpected reply type: ${message.getClass.getSimpleName}")
+    }
+  }
+
+  private[this] def loop(): Unit =
```
mosesn: shouldn't we check at some point in time that this loop has finished?  let's return Future[Unit] here.

finagle-redis/src/main/scala/com/twitter/finagle/redis/dispatch/SubscribeDispatcher.scala
```diff
@@ -0,0 +1,70 @@
...
+    trans.read().onSuccess { reply =>
```
mosesn: let's flatMap here instead.

finagle-redis/src/main/scala/com/twitter/finagle/redis/dispatch/SubscribeDispatcher.scala
```diff
@@ -0,0 +1,70 @@
...
+      handleMessage(reply)
+      loop()
+    }
+    .onFailure { ex =>
+      Option(handler.get()).map(_.onException(this, ex))
```
mosesn: it's idiomatic to use foreach for side-effects.

finagle-redis/src/main/scala/com/twitter/finagle/redis/dispatch/SubscribeDispatcher.scala
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+  protected def dispatch(req: SubscribeCommand, p: Promise[Unit]): Future[Unit] =
+    trans.write(req).onSuccess { _ => p.setDone() }
```
mosesn: should we fail the promise if we fail the write?

finagle-redis/src/main/scala/com/twitter/finagle/Redis.scala
thirstycrow: I think it's ok to share the same connection among all subscribers, and share another for all regular uses. And perhaps we can also use the for-regular-use connection for transactions, if we make sure that all commands for the same transaction will be sent with a single write (by buffering the commands involved in a transaction in client-side local buffers, and sending them all together when a EXEC command is issued)

mosesn: hmm, interesting.  I guess it depends on the implementation–the downside is that if you send the parts of a transaction eagerly instead of lazily, then if some of the work is nontrivial, like sending a big request, you don't delay it until the last moment.  especially since finagle typically moves IO to a different thread, this would be a bummer.

thirstycrow: Yes, you're right. And another thing I failed to take into account is the WATCH command. It's not part of the transaction, but affects the transaction result. So, we're gonna need individual connections for each concurrent transactions.

finagle-redis/src/main/scala/com/twitter/finagle/redis/dispatch/SubscribeDispatcher.scala
thirstycrow: I think it's not necessary. We do the work in the onFailure callback.

mosesn: my concern is that we have no way of checking if it hangs forever and leaks memory.  admittedly it's probably not terribly likely.

thirstycrow: It may be better to check it at somewhere underlying, in the transport, perhaps.

mosesn: I think it's a valid use case to close but still be reading in the transport though.  Well, we can keep it this way for now, and fix it if it causes a problem.

finagle-redis/src/main/scala/com/twitter/finagle/Redis.scala
thirstycrow: It seems that the current TransactionalClient implementation is not thread safe for the same reason - WATCH's are no guaranteed to use the same connection that the transaction is going to use.

thirstycrow:
I moved the subscribe client things into c.t.f.redis.exp. And I hope we can make an agreement on the rich api first. The underlying implementation can be optimized later anyway.

Here's what I expect it should look like:

Creating a client.

```scala
val client: RedisRichClient = Redis.newRichClient(hostAndPort) // instead of Redis.client.newRichClient(hostAndPort)
```

Normal commands are kept as-is.

```scala
client.set(key, value)
client.get(key)
```

I hope we can subscribe/unsubscribe just like how we use normal commands, hiding the fact that a different connection is used.

```scala
client.subscribe(foo, bar)
client.unsubscribe(bar)
```

Transactions should be done isolated with each other, and with explicit boundaries.

```scala
client.transactional { tx =>
  tx.watch(Seq(foo))
  tx.multi()
  tx.set(bar, baz)
  if (someTest()) {
    tx.commit()
  } else {
    tx.discard()
  }
}
```

'transactional' is defined as following. Hope they are not going to add a TRANSACTIONAL command in the future :)

```scala
def transactional: Service[TransactionalClient => Future[T], T]
```

Sentinels use an almost completely different set of commands, I think we should type it as another class.

```scala
val sentinel: RedisSentinelRichClient = Redis.newSentinelClient(hostAndPort)

sentinel.masters()
sentinel.slaves(masterName)
```

I don't have ideas about how to do it with redis cluster yet.

mosesn:
We should keep the `Redis.client.newRichClient` style instead of `Redis.newRichClient` style, because that's the style throughout the rest of finagle.

What do you think about using the same transaction API as we have in [mysql][0]?

`  def transaction[T](f: Client => Future[T]): Future[T]`

Otherwise, sounds good.

[0]: https://github.com/twitter/finagle/blob/develop/finagle-mysql/src/main/scala/com/twitter/finagle/mysql/Client.scala#L95

thirstycrow:
@mosesn Sorry for late reply, I didn't get much spare time last week.

I finished the rich client part with the latest commits. Please give a review on it.

It seems it's finagle-redis that is using a different style than the rest. I checked finagle-http and finagle-mysql, and they are using the ```Protocol.newRichClient``` style. So I changed to this style in the new code.

finagle-redis/src/main/scala/com/twitter/finagle/redis/Client.scala
```diff
@@ -38,79 +34,66 @@ class Client(service: Service[Command, Reply])
...
-   * Closes connection to Redis instance
-   */
-  def quit(): Future[Unit] =
-    doRequest(Quit) {
-      case StatusReply(message) => Future.Unit
+  def transactionSupport[T](f: TransactionalClient => Future[T]): Future[T] = {
+    val singleton = singletonFactory
+    val client = new TransactionalClient(singleton)
+    f(client).ensure {
+      client.reset() ensure singleton.close()
```
taylorleese: ```client.reset().ensure(singleton.close())```

CHANGES
```diff
@@ -7,6 +7,17 @@ Note that ``RB_ID=#`` correspond to associated messages in commits.
 6.x
 -----
 
+New Features
+~~~~~~~~~~~~
+
+  * finagle-redis: Supports for Pub/Sub.
```
taylorleese: Supports => Support

finagle-redis/src/main/scala/com/twitter/finagle/redis/Client.scala
```diff
@@ -38,79 +34,66 @@ class Client(service: Service[Command, Reply])
...
+abstract class BaseClient(
+    protected val factory: ServiceFactory[Command, Reply])
+  extends Closable {
 
   /**
-   * Releases underlying service object
+   * Releases underlying service factory object
    */
-  def release() = service.close()
+  def close(deadline: Time) = factory.close(deadline)
```
taylorleese: Let's add an explicit return type here.

finagle-redis/src/main/scala/com/twitter/finagle/redis/Client.scala
```diff
@@ -38,79 +34,66 @@ class Client(service: Service[Command, Reply])
...
   /**
    * Helper function for passing a command to the service
    */
-  private[redis] def doRequest[T](cmd: Command)(handler: PartialFunction[Reply, Future[T]]) =
-    service(cmd) flatMap (handler orElse {
-      case ErrorReply(message)  => Future.exception(new ServerError(message))
-      case _                    => Future.exception(new IllegalStateException)
-    })
+  private[redis] def doRequest[T](cmd: Command)(handler: PartialFunction[Reply, Future[T]]) = {
+    factory.toService.apply(cmd) flatMap (handler orElse {
```
taylorleese: Use .flatMap.

finagle-redis/src/main/scala/com/twitter/finagle/redis/protocol/commands/PubSub.scala
```diff
@@ -0,0 +1,96 @@
...
+  def toChannelBuffer = RedisCodec.toUnifiedFormat(CommandBytes.SUBSCRIBE +: channels)
+}
+
+case class Unsubscribe(channels: Seq[ChannelBuffer], handler: SubscribeHandler) extends SubscribeCommand {
+  def command = Commands.UNSUBSCRIBE
+  def toChannelBuffer = RedisCodec.toUnifiedFormat(CommandBytes.UNSUBSCRIBE +: channels)
+}
+
+object Publish {
+  def apply(args: List[Array[Byte]]) = {
```
taylorleese: Add return type.

finagle-redis/src/main/scala/com/twitter/finagle/redis/util/TestServer.scala
```diff
@@ -21,20 +22,28 @@ object RedisCluster { self =>
...
-      instanceStack.push(instance)
+  def start(count: Int = 1): Seq[ExternalRedis] = {
+    0 until count map { i =>
+      start(new ExternalRedis())
     }
   }
-  def stop() {
-    instanceStack.pop().stop()
+
+  def start(instance: ExternalRedis) = {
```
taylorleese: Add return type.

finagle-redis/src/main/scala/com/twitter/finagle/redis/util/TestServer.scala
```diff
@@ -21,20 +22,28 @@ object RedisCluster { self =>
...
+    instance.start()
+    instanceStack.push(instance)
+    instance
+  }
+
+  def stop() = {
```
taylorleese: Add return type.

finagle-redis/src/main/scala/com/twitter/finagle/redis/util/TestServer.scala
```diff
@@ -106,6 +115,13 @@ class ExternalRedis() {
     start()
   }
 
+  def newClient() = Redis.newRichClient(s"127.0.0.1:${address.get.getPort}")
```
taylorleese: Add return type.

finagle-redis/src/test/scala/com/twitter/finagle/redis/RedisTest.scala
```diff
@@ -23,6 +24,26 @@ trait RedisTest extends FunSuite {
   protected val boo = StringToChannelBuffer("boo")
   protected val moo = StringToChannelBuffer("moo")
   protected val num = StringToChannelBuffer("num")
+
+  def result[T](awaitable: Awaitable[T], timeout: Duration = 1.second) =
```
taylorleese: Add return types here and below.

thirstycrow:
Fixed the coding style issues. And since SubscribeCommand now subclasses Command instead of RedisMessage, there is no need to use Encoder[RedisMessage], so I changed it back.

mahakp:
lgtm

finagle-redis/src/main/scala/com/twitter/finagle/redis/Client.scala
```diff
@@ -38,79 +34,66 @@ class Client(service: Service[Command, Reply])
...
 
-  /**
-   * Authorizes to db
-   * @param password
-   */
-  def auth(password: ChannelBuffer): Future[Unit] =
-    doRequest(Auth(password)) {
-      case StatusReply(message) => Future.Unit
-    }
+  private[this] def singletonFactory: ServiceFactory[Command, Reply] =
```
mosesn: let's add parens for this function so it's clear that it's a function, and not an accessor.

finagle-redis/src/main/scala/com/twitter/finagle/redis/exp/RedisPool.scala
```diff
@@ -0,0 +1,68 @@
...
+  private[this] val singletonPool =
+    new SingletonPool(underlying, statsReceiver.scope("singletonpool"))
+
+  private[this] val subscribePool =
+    new SingletonPool(underlying, statsReceiver.scope("subscribepool"))
+
+  final def apply(conn: ClientConnection): Future[Service[Command, Reply]] = {
+    RedisPool.useFor() match {
+      case Some(RedisPool.Transaction) =>
+        underlying(conn)
```
mosesn: my suspicion is that we can do a little bit better than this by using a conventional serial pool (like DefaultPool) but we can make this optimization in the future.

mosesn:
thanks @mahakp!  Let's get this out the door, thanks for being so patient @thirstycrow.  You don't have to make any changes, the notes I just made were so for myself.

mosesn:
Sorry, ended up in jury duty last week, trying again to merge this internally.

mosesn:
This was finally merged internally.  Thanks for your patience!

mosesn:
@thirstycrow thanks for being patient with us!  it's finally in develop.  https://github.com/twitter/finagle/commit/b6bc1601bdb6bbe8fda8c9c193b540166fa48fb4

PR #448, opened 2015-12-09
finagle-core: Add ServerDispatcher and AnnotatedServerDispatcher

penland365:
wrapper types.

Problem:

stack. There is no parallel for a Server. In addition, we would like
to have a simple type for transforming ServerDispatchers into
AnnotatedServerDispatchers.

Solution:

Introduce the ServerDispatcher type, which has two methods dispatch
and handle. Introduce the AnnotatedServerDispatcher type as well,
which can take any ServerDispatcher, and return that dispatcher
with WireRecv on receipt and WireSend upon a succesful write
by the transporter.

This is an attempt to find a solution to https://github.com/twitter/finagle/pull/401 .


finagle-core/src/main/scala/com/twitter/finagle/dispatch/ServerDispatcher.scala
```diff
@@ -103,8 +114,30 @@ class SerialServerDispatcher[Req, Rep](
...
+  * A wrapper type around a 
+  * [[com.twitter.finagle.dispatcher.ServerDispather ServerDispatcher]]
+  *
+  * This type adds a 
+  * [[com.twitter.finagle.tracing.Annotation.WireRecv WireRecv]]
+  * once a dispatcher has received a request, and a
+  * [[com.twitter.finagle.tracing.Annotation.WireSend WireSend]] on a succesful
+  * write by a transporter.
+ */
+final class AnnotatedServerDispatcher[Req, Rep, In, Out, A <: ServerDispatcher[Req, Rep, In, Out]]
```
mosesn: why parameterize A?

finagle-core/src/main/scala/com/twitter/finagle/dispatch/ServerDispatcher.scala
```diff
@@ -1,11 +1,36 @@
...
 import com.twitter.finagle.{Service, NoStacktrace, CancelledRequestException}
 import com.twitter.util._
 import java.util.concurrent.atomic.AtomicReference
 
+/**
+  * A base Server Dipatcher type that allows us to safely Annotate a
+  * dispatcher via
+  * [[com.twitter.finagle.dispatcher.ServerDispatherAnnotator ServerDispatcherAnnotator]]
+ */
+trait ServerDispatcher[Req, Rep, In, Out] {
```
mosesn: looks like `In` is never used?

finagle-core/src/main/scala/com/twitter/finagle/dispatch/ServerDispatcher.scala
```diff
@@ -103,8 +114,30 @@ class SerialServerDispatcher[Req, Rep](
...
+  * [[com.twitter.finagle.tracing.Annotation.WireRecv WireRecv]]
+  * once a dispatcher has received a request, and a
+  * [[com.twitter.finagle.tracing.Annotation.WireSend WireSend]] on a succesful
+  * write by a transporter.
+ */
+final class AnnotatedServerDispatcher[Req, Rep, In, Out, A <: ServerDispatcher[Req, Rep, In, Out]]
+  (dispatcher: A) extends ServerDispatcher[Req, Rep, In, Out] {
+
+  override def handle(rep: Rep): Future[Unit] =
+    dispatcher.handle(rep) onSuccess { _ => Trace.record(Annotation.WireSend) }
```
mosesn: `.onSuccess {`

finagle-core/src/main/scala/com/twitter/finagle/dispatch/ServerDispatcher.scala
```diff
@@ -103,8 +114,30 @@ class SerialServerDispatcher[Req, Rep](
...
+  override def dispatch(req: Req, eos: Promise[Unit]) =
     service(req) ensure eos.setDone()
 
-  protected def handle(rep: Rep) = trans.write(rep)
+  override def handle(rep: Rep) = trans.write(rep)
+}
+
+/**
+  * A wrapper type around a 
+  * [[com.twitter.finagle.dispatcher.ServerDispather ServerDispatcher]]
```
mosesn: s/ServerDispather/ServerDispatcher

finagle-core/src/main/scala/com/twitter/finagle/dispatch/ServerDispatcher.scala
```diff
@@ -103,8 +114,30 @@ class SerialServerDispatcher[Req, Rep](
...
-  protected def dispatch(req: Req, eos: Promise[Unit]) =
+  override def dispatch(req: Req, eos: Promise[Unit]) =
     service(req) ensure eos.setDone()
 
-  protected def handle(rep: Rep) = trans.write(rep)
+  override def handle(rep: Rep) = trans.write(rep)
+}
+
+/**
+  * A wrapper type around a
```
mosesn: "A wrapper type" isn't terribly descriptive.  How about something like:

"Wraps a ServerDispatcher, adding annotations when a request is read from the wire, and later is written to the wire."

finagle-core/src/main/scala/com/twitter/finagle/dispatch/ServerDispatcher.scala
mosesn: let's also be clear about what kinds of ServerDispatchers it can wrap.  this is only OK for RPC, right?

finagle-core/src/main/scala/com/twitter/finagle/dispatch/ServerDispatcher.scala
```diff
@@ -1,11 +1,36 @@
...
+   * argument `eos` (end-of-stream promise) must be fulfilled when the request
+   * is complete.
+   *
+   * For non-streaming requests, `eos.setDone()` should be called immediately,
+   * since the entire request is present. For streaming requests,
+   * `eos.setDone()` must be called at the end of stream (in HTTP, this is on
+   * receipt of last chunk). Refer to the implementation in
+   * [[com.twitter.finagle.http.codec.HttpServerDispatcher]].
+   */
+  private[dispatch] def dispatch(req: Out, eos: Promise[Unit]): Future[Rep]
```
mosesn: why not keep this protected?

mosesn:
LGTM in general (caveat lector: I consulted on the final design, so I'm biased :P)

Could you add a test?  Also, could you wrap the server dispatchers you care about in a few protocols to demonstrate?  I think the relevant ones will be http, thrift, at least.  mux if you can get away with it.

finagle-core/src/main/scala/com/twitter/finagle/dispatch/ServerDispatcher.scala
penland365: It shouldn't be - that was leftover from a functional attempt to return the same `type` passed in, and I had just `yy` the original method definition.

Fixed.

finagle-core/src/main/scala/com/twitter/finagle/dispatch/ServerDispatcher.scala
penland365: Done

finagle-core/src/main/scala/com/twitter/finagle/dispatch/ServerDispatcher.scala
penland365: My desire is to have this only wrap anything that is a sub-type of `ServerDispatcher`. That would mean mixing in the type on things like [finagle-http](https://github.com/twitter/finagle/blob/develop/finagle-http/src/main/scala/com/twitter/finagle/http/codec/HttpServerDispatcher.scala#L17-L20) and [finagle-mux](https://github.com/twitter/finagle/blob/develop/finagle-mux/src/main/scala/com/twitter/finagle/mux/Server.scala#L186-L192), but ( fingers crossed ) that shouldn't entail too much work.

finagle-core/src/main/scala/com/twitter/finagle/dispatch/ServerDispatcher.scala
penland365: Done.

finagle-core/src/main/scala/com/twitter/finagle/dispatch/ServerDispatcher.scala
penland365: If it's protected it's not visible to the `AnnotatedServerDispatcher`. I went back and forth on this - I'm open to other solutions.

finagle-core/src/main/scala/com/twitter/finagle/dispatch/ServerDispatcher.scala
penland365: Yeah, it's not. Same as above - I had `yy` that definition from another implementation that was an abstract class that need a Transporter.

Fixed.

finagle-core/src/main/scala/com/twitter/finagle/dispatch/ServerDispatcher.scala
mosesn: OK, let me be more precise.  There are at least three models of request / response interactions that we're considering in finagle right now.  We support two of them, byte streaming (over http) and object RPC today, and we're considering support for a third, which would be object streaming.

AnnotationServerDispatcher is appropriate for object RPC (we have everything we need as soon as we write the first byte to the wire), but it's probably not appropriate for either object streaming or byte streaming.

finagle-core/src/main/scala/com/twitter/finagle/dispatch/ServerDispatcher.scala
mosesn: We can't make it `private[dispatch]` because we have `ServerDispatchers` outside of the `dispatch` package.  Let's make it public and call it a day–I don't think scala has the ability to say, "let subclasses fiddle with this method on other instances of this type".

finagle-core/src/main/scala/com/twitter/finagle/dispatch/ServerDispatcher.scala
```diff
@@ -1,11 +1,35 @@
...
+import com.twitter.finagle.tracing.{Annotation, Trace}
 import com.twitter.finagle.transport.Transport
 import com.twitter.finagle.{Service, NoStacktrace, CancelledRequestException}
 import com.twitter.util._
 import java.util.concurrent.atomic.AtomicReference
 
+/**
+  * A base Server Dipatcher type that allows us to safely Annotate a
+  * dispatcher via
+  * [[com.twitter.finagle.dispatcher.ServerDispatherAnnotator ServerDispatcherAnnotator]]
```
kevinoliver: s/ServerDispatherAnnotator/ServerDispatcherAnnotator/

finagle-core/src/main/scala/com/twitter/finagle/dispatch/ServerDispatcher.scala
```diff
@@ -1,11 +1,35 @@
...
 
 import com.twitter.finagle.context.Contexts
+import com.twitter.finagle.tracing.{Annotation, Trace}
 import com.twitter.finagle.transport.Transport
 import com.twitter.finagle.{Service, NoStacktrace, CancelledRequestException}
 import com.twitter.util._
 import java.util.concurrent.atomic.AtomicReference
 
+/**
+  * A base Server Dipatcher type that allows us to safely Annotate a
```
kevinoliver: s/Dipatcher/Dispatcher/

finagle-core/src/main/scala/com/twitter/finagle/dispatch/ServerDispatcher.scala
```diff
@@ -103,8 +113,29 @@ class SerialServerDispatcher[Req, Rep](
     service.close()
   }
 
-  protected def dispatch(req: Req, eos: Promise[Unit]) =
+  override def dispatch(req: Req, eos: Promise[Unit]) =
```
kevinoliver: this and `handle` don't actually override anything. we only use override for overriding implementations, not when its only an interface.

finagle-core/src/main/scala/com/twitter/finagle/dispatch/ServerDispatcher.scala
```diff
@@ -103,8 +113,29 @@ class SerialServerDispatcher[Req, Rep](
...
+  * adding a
+  * [[com.twitter.finagle.tracing.Annotation.WireRecv WireRecv]]
+  * when a request is read from the wire, and later a
+  * [[com.twitter.finagle.tracing.Annotation.WireSend WireSend]]
+  * when it writes successfully to the wire.
+ */
+final class AnnotatedServerDispatcher[Req, Rep, Out]
+  (dispatcher: ServerDispatcher[Req, Rep, Out]) extends ServerDispatcher[Req, Rep, Out] {
+
+  override def handle(rep: Rep): Future[Unit] =
```
kevinoliver: same comment re overrides here

finagle-core/src/main/scala/com/twitter/finagle/dispatch/ServerDispatcher.scala
```diff
@@ -103,8 +113,29 @@ class SerialServerDispatcher[Req, Rep](
...
+    dispatcher.handle(rep).onSuccess { _ => Trace.record(Annotation.WireSend) }
```
kevinoliver: in order to avoid the closure allocation on every call, we can store it in a member variable.

```
private[this] val RecordWireSend: Unit => Unit = 
  _ => Trace.record(Annotation.WireSend) 

...
dispatcher.handle(rep).onSuccess(RecordWireSed)

File: finagle-core/src/main/scala/com/twitter/finagle/dispatch/ServerDispatcher.scala [hunk 12]
```diff
@@ -103,8 +113,29 @@ class SerialServerDispatcher[Req, Rep](
...
+/**
+  * Wraps a
+  * [[com.twitter.finagle.dispatcher.ServerDispatcher ServerDispatcher]]
+  * adding a
+  * [[com.twitter.finagle.tracing.Annotation.WireRecv WireRecv]]
+  * when a request is read from the wire, and later a
+  * [[com.twitter.finagle.tracing.Annotation.WireSend WireSend]]
+  * when it writes successfully to the wire.
+ */
+final class AnnotatedServerDispatcher[Req, Rep, Out]
```
kevinoliver: minor style notes:
* -1 indent for the scaladoc block please.
* pull the `(dispatcher: ServerDispatcher[Req, Rep, Out])` up to the prev line.

penland365:
This implementation isn't going to work as is => we're wrapping a type that never gets called externally ( just noticed this on test ). I didn't realize the state was managed internally. We're calling `handle` and `dispatcher` internally from [here on GenSerialServerDispatcher](https://github.com/twitter/finagle/blob/develop/finagle-core/src/main/scala/com/twitter/finagle/dispatch/ServerDispatcher.scala#L51-L53) and [here on Mux](https://github.com/twitter/finagle/blob/develop/finagle-mux/src/main/scala/com/twitter/finagle/mux/Server.scala#L206).

penland365:
This yak has turned into something much larger than I first though. I've pushed a change that does annotate anything that subclass `GenSerialServerDispatcher`, and should be able to be mix-in to any other Dispatcher class that extends `ServerDispatcher` ( like hopefully mux ).

However, while trying to test this I kept failing to get a Trace to record. That would be because of this [commit](https://github.com/twitter/finagle/commit/946a755d126640bbadbfe4f3703c096c907d38cd) where we are specifically clearing local before entering the Dispatch loop. Since we don't know the `TraceId` right now at the Dispatcher level, this is going to need some additional work before it's ready to go.

penland365:
Superseded by #463

PR #449, opened 2015-12-16
P2CBalancer: Reduce the magnitude of the Penalty

stevegury:
**Problem**
The PeakEWMA load balancer applies a penalty to the load of a server when it 
doesn't have any historical latency. That penalty is initialized at 
`Double.MaxValue/2`, and we add the number of outstanding messages to compute 
the load of a server. The idea behind that, is that the load balancer behaves 
like `LeastLoaded` at start-up before having any historical data.

The problem is that the Penalty is a very big double, and adding a small 
integer will not change its value because of the lack of double precision in 
that range.
e.g.:
```
scala> val Penalty: Double = Double.MaxValue/2
Penalty: Double = 8.988465674311579E307

scala> val b = Penalty + 1
b: Double = 8.988465674311579E307

scala> Penalty < b
res4: Boolean = false

scala> Penalty == b
res5: Boolean = true
```

Actually, the number of outstanding requests needed to increase the load 
properly is humongous (bigger than Int.MaxValue).
```
scala> Math.nextUp(Penalty) - Penalty
res0: Double = 9.979201547673599E291

scala> res0 > Int.MaxValue
res1: Boolean = true
```

**Solution**
Reduce the Penalty to a value (`Long.MaxValue >> 16`) where the double precision 
is good enough to add small integers. This penalty is about equal to 140737 seconds or 39 hours, which is enough to deal with extremely latent systems (meaning that a host without historical latencies will not be chosen over one with history unless its ewma latency is above 39 hours).

### DIFF
--- a/finagle-core/src/main/scala/com/twitter/finagle/loadbalancer/P2CBalancer.scala
+++ b/finagle-core/src/main/scala/com/twitter/finagle/loadbalancer/P2CBalancer.scala
@@ -96,7 +96,7 @@ private trait PeakEwma[Req, Rep] { self: Balancer[Req, Rep] =>
 
   protected class Metric(sr: StatsReceiver, name: String) {
     private[this] val epoch = nanoTime()
-    private[this] val Penalty: Double = Double.MaxValue/2
+    private[this] val Penalty: Double = Int.MaxValue/2
     // The mean lifetime of `cost`, it reaches its half-life after Tau*ln(2).
     private[this] val Tau: Double = decayTime.inNanoseconds.toDouble
     require(Tau > 0)
--- a/finagle-core/src/main/scala/com/twitter/finagle/loadbalancer/P2CBalancer.scala
+++ b/finagle-core/src/main/scala/com/twitter/finagle/loadbalancer/P2CBalancer.scala
@@ -96,7 +96,7 @@ private trait PeakEwma[Req, Rep] { self: Balancer[Req, Rep] =>
 
   protected class Metric(sr: StatsReceiver, name: String) {
     private[this] val epoch = nanoTime()
-    private[this] val Penalty: Double = Int.MaxValue/2
+    private[this] val Penalty: Double = Long.MaxValue >> 16
     // The mean lifetime of `cost`, it reaches its half-life after Tau*ln(2).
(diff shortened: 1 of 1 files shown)


mosesn:
Steve!  How's it going?  Are you using finagle at netflix?

mosesn:
LGTM

kevinoliver:
:ship:

davebarr:
<3

stevegury:
Hi guys!
No we're not using Finagle, however I needed a load balancing algorithm and reused this one for another project. Then I found this (minor) bug.

I should swing by TwitterHQ sometime next year to say hi.

mosesn:
ahh, nice!  looking forward to your visit :+1:

stevegury:
After a night of reflexion I realized that `Int.MaxValue/2` is not enough.
Because we measure the latency in nanoseconds, it only represents a penalty of `(Int.MaxValue / 2) / 1e9` ~= 1.07 second.

Choosing a bigger Penalty (e.g. `Long.MaxValue >> 16` ~= 140737 seconds or 39 hours) fixes the problem and also works well with potentially extremely latent systems.

mosesn:
Would it be possible to use longs here instead of doubles?  I haven't looked at this part too closely, so I'm not sure what we use the fractional bits for.

stevegury:
Unfortunately, I don't think so. 
The EWMA algorithm uses an exponential to adjust the fact that the time-serie is non-evenly distributed.
Making `cost` a long (or an integer) would just move the problem of precision in the computation of cost, leading to rounding errors.

roanta:
Thanks for the PR Steve! I'll pull this in internally.

roanta:
How did you test this btw? Do you think it's feasible to add a unit test?

stevegury:
I didn't go the extra mile with writing a unit-test, but one possible solution would be to load balance requests across two latent servers and to check that the load is perfectly distributed (adding a counter on the Service or something).
Indeed with only two hosts, the P2C should behave like LeastLoaded (or JSQ) and the penalty shouldn't perturb the balancing.

mosesn:
This was merged in here: https://github.com/twitter/finagle/commit/9a5d4af4ed5d36f314962797636d8bfc9381b95a thanks!

PR #451, opened 2015-12-22
Control Thrift transport framing with a Stack.Param

olix0r:
Problem:

It is cumbersome to control whether Thrift clients and servers are framed
because there is no Stack.Param that influences this behavior.

Solution:

Introduce a Stack.Param, Thrift.param.Framed, and remove the Client/Server
boolean attribute.

### DIFF
--- a/finagle-thrift/src/main/scala/com/twitter/finagle/Thrift.scala
+++ b/finagle-thrift/src/main/scala/com/twitter/finagle/Thrift.scala
@@ -73,6 +73,11 @@ object Thrift extends Client[ThriftClientRequest, Array[Byte]] with ThriftRichCl
       val default = ProtocolFactory(Protocols.binaryFactory())
     }
 
+    case class Framed(enabled: Boolean)
+    implicit object Framed extends Stack.Param[Framed] {
+      val default = Framed(true)
+    }
+
     /**
      * A `Param` to set the max size of a reusable buffer for the thrift response.
      * If the buffer size exceeds the specified value, the buffer is not reused,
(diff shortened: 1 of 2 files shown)


finagle-thrift/src/main/scala/com/twitter/finagle/Thrift.scala
```diff
@@ -73,6 +73,11 @@ object Thrift extends Client[ThriftClientRequest, Array[Byte]] with ThriftRichCl
       val default = ProtocolFactory(Protocols.binaryFactory())
     }
 
+    case class Framed(enabled: Boolean)
```
kevinoliver: please add scaladoc for what this is for and that it defaults to on.

finagle-thrift/src/main/scala/com/twitter/finagle/Thrift.scala
```diff
@@ -128,8 +133,7 @@ object Thrift extends Client[ThriftClientRequest, Array[Byte]] with ThriftRichCl
 
   case class Client(
     stack: Stack[ServiceFactory[ThriftClientRequest, Array[Byte]]] = Client.stack,
-    params: Stack.Params = StackClient.defaultParams + ProtocolLibrary("thrift"),
-    framed: Boolean = true
+    params: Stack.Params = StackClient.defaultParams + ProtocolLibrary("thrift")
```
kevinoliver: this is a breaking API change so it should be listed in CHANGES.

kevinoliver:
@olix0r just two small things.

kevinoliver:
lgtm, i'll try to get this pulled in.

kevinoliver:
Thanks for your patience with the holidays. This patch got merged in late today and should show up on the develop branch next Monday.

olix0r:
thanks! 🐋

PR #457, opened 2016-02-02
Create a simple example of how to use zookeeper with finagle

sunnykaka:
Problem

Lack of examples and documents about how to using zookeeper with finagle.

Solution

Create a simple example of how to use zookeeper with finagle.

Result

It's a example, it does not affect other modules.

### DIFF
--- a/finagle-example/src/main/scala/com/twitter/finagle/example/zookeeper/ServiceConsumer.scala
+++ b/finagle-example/src/main/scala/com/twitter/finagle/example/zookeeper/ServiceConsumer.scala
@@ -3,12 +3,11 @@ package com.twitter.finagle.example.zookeeper
 import java.io.ByteArrayInputStream
 import java.nio.charset.StandardCharsets
 
-import com.twitter.finagle.Service
-import com.twitter.finagle.builder.ClientBuilder
+import com.twitter.finagle.Http
 import com.twitter.finagle.example.zookeeper.ServiceProvider.EchoServer
-import com.twitter.finagle.http._
+import com.twitter.finagle.http.{Http => _, _}
 import com.twitter.io.Reader
-import com.twitter.util.Future
+import com.twitter.util.Await
 
 /**
   * This demo shows how to consume the service in zookeeper.
@@ -21,22 +20,21 @@ object ServiceConsumer {
   def main(args: Array[String]): Unit = {
 
     //use zookeeper to discover service
-    val client: Service[Request, Response] = ClientBuilder()
-      .codec(Http())
-      .hostConnectionLimit(10)
-      .dest(ServiceProvider.buildConsumerPath(EchoServer.servicePath))
-      .build()
+    val client = Http.client.
+      withSessionPool.maxSize(10).
+      newService(ServiceProvider.buildConsumerPath(EchoServer.servicePath), "echo-client")
 
     //create a "Greetings!" request.
     val data = Reader.fromStream(new ByteArrayInputStream("Greetings!".getBytes(StandardCharsets.UTF_8)))
     val request = Request(Version.Http11, Method.Post, "/", data)
 
-    client(request) onSuccess { response =>
+    Await.ready(client(request)) onSuccess { response =>
       println(s"response status: ${response.status}, response string: ${response.contentString} ")
     } onFailure { e =>
       println(s"error: $e")
     } ensure {
       client.close()
     }
+
   }
 }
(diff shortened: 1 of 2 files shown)


finagle-example/src/main/scala/com/twitter/finagle/example/zookeeper/ServiceConsumer.scala
```diff
@@ -0,0 +1,42 @@
...
+  *
+  * Run ServiceProvider before ServiceConsumer
+  *
+  */
+object ServiceConsumer {
+
+  def main(args: Array[String]): Unit = {
+
+    //use zookeeper to discover service
+    val client: Service[Request, Response] = ClientBuilder()
```
mosesn: would you mind using a finagle 6-style client?  This instead becomes:

```scala
Http
  .client
  .withSessionPool.maxSize(10)
  .newService(ServiceProvider.buildConsumerPath(EchoServer.servicePath), "http-client")
```

The new version of finagle that supports the with-style parameters should be released later this week.

finagle-example/src/main/scala/com/twitter/finagle/example/zookeeper/ServiceProvider.scala
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  object EchoServer {
+    val servicePath = "/services/echo"
+  }
+
+  def main(args: Array[String]): Unit = {
+
+    //start your service on port 8080
+    val server: Server = ServerBuilder()
```
mosesn: Ditto here:

```scala
Http.server.withLabel("echoserver").serve(":8080", new EchoServer)
```

mosesn:
LGTM, but it would be nice to use zk2 instead of zk, which is what we use internally, and we think it's more reliable.  I'll get some folks to come take a look at your usage :+1:

chchen:
zk2 only works for the resolution path, right? For announcing, zk is the only supported scheme I think.

sunnykaka:
Thanks @mosesn , I will use the finagle 6-style api.

sunnykaka:
Hello @chchen , I didn't use zk2 in my project yet.
I'll try it.

finagle-example/src/main/scala/com/twitter/finagle/example/zookeeper/ServiceConsumer.scala
```diff
@@ -0,0 +1,40 @@
...
+  * Run ServiceProvider before ServiceConsumer
+  *
+  */
+object ServiceConsumer {
+
+  def main(args: Array[String]): Unit = {
+
+    //use zookeeper to discover service
+    val client = Http.client.
+      withSessionPool.maxSize(10).
```
mosesn: can we move the . to be in front?  that's the typical style we use inf inagle

finagle-example/src/main/scala/com/twitter/finagle/example/zookeeper/ServiceProvider.scala
```diff
@@ -0,0 +1,65 @@
+package com.twitter.finagle.example.zookeeper
+
+import java.net.InetSocketAddress
```
mosesn: do you mind sorting your imports?

finagle-example/src/main/scala/com/twitter/finagle/example/zookeeper/ServiceProvider.scala
```diff
@@ -0,0 +1,65 @@
...
+    //syntax:          schema!host!path!shardId
+    //schema:          for server, use zk, for client, use zk2
+    //host:            zookeeper connection string
+    //path:            service registration path in zookeeper
+    //shardId:         it's used internally by Twitter, can be set to 0 in most cases
+    val echoServicePath = buildProviderPath(EchoServer.servicePath)
+
+    //start your service on port 8080 and register to zookeeper
+    Await.ready(
+      Http.server.
```
mosesn: dots in front :+1:

mosesn:
LGTM except for a couple nits.  Sorry for the delay, github doesn't tell us when you merge in a new commit.

sunnykaka:
Hi @mosesn , I have fix it.

mosesn:
LGTM, thanks!

mosesn:
@sunnykaka this was merged internally, thanks for the contribution!  :fire:

mosesn:
This made it to develop here: https://github.com/twitter/finagle/commit/766f38974663b80509f7d19ae9263c659b11d8a6.  Really appreciate the PR!  :fire:

PR #459, opened 2016-02-04
Add extractors for Status of HTTP Response

spockz:
> When writing response classifiers, and other code that deals with responses,
> in a generic manner, it is currently inconvenient to match against specific
> ranges of errors such as Server Errors or Client errors.
> 
> Solution
> 
> By adding extractors for the different categories (Unknown, Informational,
> Successful, Redirection, Client Error, and Server Error) it becomes possible
> to easily determine in which category a response falls.
> 
> Result
> 
> Determining whether a response was succesful is as easy as matching
> on Successful(_).

It can be debated whether we should also provide an extractor that works on the `Response` directly, and if so, where the code should be placed.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -9,7 +9,7 @@ Note that ``RB_ID=#`` correspond to associated messages in commits.
 
 New Features
 ~~~~~~~~~~~~
-  * finagle-http: Introduce convience extractors to pattern match `c.t.f.http.Response.status`
+  * finagle-http: Introduce convenience extractors to pattern match `c.t.f.http.Response.status`
     against the different categories.
 
 6.33.0
--- a/CHANGES
+++ b/CHANGES
@@ -7,8 +7,16 @@ Note that ``RB_ID=#`` correspond to associated messages in commits.
 6.x
 -----
 
+6.34.0
+~~~~~~
+
+New Features
+~~~~~~~~~~~~
+  * finagle-http: Introduce convience extractors to pattern match `c.t.f.http.Response.status`
+    against the different categories.
+
 6.33.0
-~~~~~~~
+~~~~~~
 
 New Features
 ~~~~~~~~~~~~
--- a/CHANGES
+++ b/CHANGES
@@ -7,9 +7,6 @@ Note that ``RB_ID=#`` correspond to associated messages in commits.
 6.x
 -----
 
-6.34.0
-~~~~~~
-
 New Features
 ~~~~~~~~~~~~
   * finagle-http: Introduce convience extractors to pattern match `c.t.f.http.Response.status`
--- a/finagle-http/src/main/scala/com/twitter/finagle/http/Status.scala
+++ b/finagle-http/src/main/scala/com/twitter/finagle/http/Status.scala
@@ -33,50 +33,39 @@ case class Status(code: Int) {
 object Status {
 
   /**
-   * Matches when the status code isn't in range of any of the
-   * categories: {{Informational}}, {{Successful}}, {{Redirection}},
-   * {{ClientError}}, {{ServerError}}.
+   * Matches when the status code isn't in range of any of the  categories: {{Informational}},
+   * {{Successful}}, {{Redirection}}, {{ClientError}}, {{ServerError}}.
    */
   object UnknownStatus {
     def unapply(status: Status): Option[Status] =
       Option(status).filter(s => s.code < 100 || s.code >= 600)
   }
 
-  /**
-   * Matches when the status code is between 100 and 200.
-   */
+  /** Matches when the status code is between 100 and 200. */
   object Informational {
     def unapply(status: Status): Option[Status] =
       inRange(100, 200, status)
   }
 
-  /**
-   * Matches when the status code is between 200 and 300.
-   */
+  /** Matches when the status code is between 200 and 300. */
   object Successful {
     def unapply(status: Status): Option[Status] =
       inRange(200, 300, status)
   }
 
-  /**
-   * Matches when the status code is between 300 and 400.
-   */
+  /** Matches when the status code is between 300 and 400. */
   object Redirection {
     def unapply(status: Status): Option[Status] =
       inRange(300, 400, status)
   }
 
-  /**
-   * Matches when the status code is between 400 and 500.
-   */
+  /** Matches when the status code is between 400 and 500. */
   object ClientError {
     def unapply(status: Status): Option[Status] =
       inRange(400, 500, status)
   }
 
-  /**
-   * Matches when the status code is between 500 and 600.
-   */
+  /** Matches when the status code is between 500 and 600. */
   object ServerError {
     def unapply(status: Status): Option[Status] =
       inRange(500, 600, status)
--- a/finagle-http/src/main/scala/com/twitter/finagle/http/Status.scala
+++ b/finagle-http/src/main/scala/com/twitter/finagle/http/Status.scala
@@ -74,7 +74,7 @@ object Status {
   @inline
   private[finagle] def inRange(lower: Int, upper: Int,
                                status: Status): Option[Status] =
-    Option(status).filter(s => s.code >= lower && s.code < upper)
+    Some(status).filter(s => s.code >= lower && s.code < upper)
 
 
   val Continue = Status(100)
--- a/finagle-http/src/main/scala/com/twitter/finagle/http/Status.scala
+++ b/finagle-http/src/main/scala/com/twitter/finagle/http/Status.scala
@@ -36,9 +36,9 @@ object Status {
    * Matches when the status code isn't in range of any of the  categories: {{Informational}},
    * {{Successful}}, {{Redirection}}, {{ClientError}}, {{ServerError}}.
    */
-  object UnknownStatus {
+  object Unknown {
     def unapply(status: Status): Option[Status] =
-      Option(status).filter(s => s.code < 100 || s.code >= 600)
+      Some(status).filter(s => s.code < 100 || s.code >= 600)
   }
 
   /** Matches when the status code is between 100 and 200. */
--- a/finagle-http/src/main/scala/com/twitter/finagle/http/Status.scala
+++ b/finagle-http/src/main/scala/com/twitter/finagle/http/Status.scala
@@ -29,7 +29,47 @@ case class Status(code: Int) {
     }
 }
 
+
+
+
+
 object Status {
+
+  @inline
+  private[finagle] def inRange(lower: Int, upper: Int)
+                              (status: Status): Option[Status] =
+    Some(status).filter(s => s.code >= lower && s.code < upper)
+
+  object UnknownStatus {
+    def unapply(status: Status): Option[Status] =
+      Some(status).filter(s => s.code < 100 || s.code >= 600)
+  }
+
+  object Informational {
+    def unapply(status: Status): Option[Status] =
+      inRange(100, 200)(status)
+  }
+
+  object Successful{
+    def unapply(status: Status): Option[Status] =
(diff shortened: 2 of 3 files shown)


CHANGES
```diff
@@ -7,8 +7,16 @@ Note that ``RB_ID=#`` correspond to associated messages in commits.
 6.x
 -----
 
+6.34.0
```
kevinoliver: leave this next version out please. our release tooling automatically bumps the 6.x to the next version when its release time.

finagle-http/src/main/scala/com/twitter/finagle/http/Status.scala
```diff
@@ -29,7 +29,47 @@ case class Status(code: Int) {
...
+    def unapply(status: Status): Option[Status] =
+      Some(status).filter(s => s.code < 100 || s.code >= 600)
+  }
+
+  object Informational {
+    def unapply(status: Status): Option[Status] =
+      inRange(100, 200)(status)
+  }
+
+  object Successful{
```
kevinoliver: nit: space before the {

finagle-http/src/main/scala/com/twitter/finagle/http/Status.scala
```diff
@@ -29,7 +29,47 @@ case class Status(code: Int) {
...
+  private[finagle] def inRange(lower: Int, upper: Int)
+                              (status: Status): Option[Status] =
+    Some(status).filter(s => s.code >= lower && s.code < upper)
+
+  object UnknownStatus {
+    def unapply(status: Status): Option[Status] =
+      Some(status).filter(s => s.code < 100 || s.code >= 600)
+  }
+
+  object Informational {
```
kevinoliver: mind adding trivial scaladoc to these public objects? can be as simple as:

```
/** Matches when the status code is between 100–199. */
```

kevinoliver:
this will be a nice addition. thanks.

mosesn:
hey, would you mind adding a commit where the first line of the commit message is the title of this RB, prefixed with `finagle-http: `?

finagle-http/src/main/scala/com/twitter/finagle/http/Status.scala
```diff
@@ -29,7 +29,47 @@ case class Status(code: Int) {
...
 
+
+
+
+
 object Status {
+
+  @inline
+  private[finagle] def inRange(lower: Int, upper: Int)
+                              (status: Status): Option[Status] =
```
taylorleese: Do we really need this to be curried?

finagle-http/src/main/scala/com/twitter/finagle/http/Status.scala
```diff
@@ -29,7 +29,47 @@ case class Status(code: Int) {
...
+    Some(status).filter(s => s.code >= lower && s.code < upper)
```
taylorleese: On the off chance somebody passes in a null Status it would be nice if this were ```Option(status)``` here rather than ```Some(status)```. Same comment applies to line 45.

spockz:
@mosesn: Added the commit per request

finagle-http/src/main/scala/com/twitter/finagle/http/Status.scala
spockz: fixed

finagle-http/src/main/scala/com/twitter/finagle/http/Status.scala
spockz: Done

CHANGES
spockz: Done

finagle-http/src/main/scala/com/twitter/finagle/http/Status.scala
spockz: No, not really. It was like that because I wanted to write the unapply function like this:
```scala
def unapply: Status => Option[Status] =
  inRange(100,200)
```
But it appears that leads to several compile errors.

finagle-http/src/main/scala/com/twitter/finagle/http/Status.scala
spockz: Good point, fixed.

finagle-http/src/main/scala/com/twitter/finagle/http/Status.scala
```diff
@@ -1,20 +1,20 @@
...
- * - it simplifies construction (users only supply the code)
- * - it avoids the need to validate user-defined reason phrases
- * - it omits the possibility of statuses with duplicate reason phrases
- *
- * The only downside is that we lose the ability to create custom statuses with
- * "vanity" reason phrases, but this should be tolerable.
- *
- * For Java-friendly enums, see [[com.twitter.finagle.http.Statuses]].
- */
+  * Represents an HTTP status code.
```
kevinoliver: sorry to nitpick: but our style used for scaladoc is 1 less indent on *s.

```
/**
 * ...
 * ...
 */
```

kevinoliver:
lgtm @spockz once you cleanup the scaladoc style.

spockz:
Should we/I add an extra extractor to match on a `Response` as well? Maybe in a different PR?

finagle-http/src/main/scala/com/twitter/finagle/http/Status.scala
spockz: Is there something that automatically formats this way? My auto-layout destroys it back to the layout in the PR. Was this also described somewhere?

mosesn: Yep, we adhere to the comment style here: https://twitter.github.io/effectivescala/#Formatting-Comments

finagle-http/src/main/scala/com/twitter/finagle/http/Status.scala
```diff
@@ -29,7 +29,65 @@ case class Status(code: Int) {
...
+  /**
+   * Matches when the status code is between 500 and 600.
+   */
+  object ServerError {
+    def unapply(status: Status): Option[Status] =
+      inRange(500, 600, status)
+  }
+
+  @inline
+  private[finagle] def inRange(lower: Int, upper: Int,
```
mosesn: can you just put all of this on one line? we typically don't wrap until 100 columns.

finagle-http/src/test/scala/com/twitter/finagle/http/StatusTest.scala
```diff
@@ -15,4 +16,45 @@ class StatusTest extends FunSuite {
     assert(Status(599).reason == "Server Error")
     assert(Status(601).reason == "Unknown Status")
   }
+
+  test("Pattern match usability check") {
```
mosesn: can you break this up into separate tests for each thing you're testing?  ie, "matches unknown" "matches informational" etc.

finagle-http/src/test/scala/com/twitter/finagle/http/StatusTest.scala
```diff
@@ -15,4 +16,45 @@ class StatusTest extends FunSuite {
...
+    null match {
```
mosesn: why test this?  null isn't a valid input, is it?  if it is, we should instead accept an Option[Status], imo.

CHANGES
```diff
@@ -7,8 +7,13 @@ Note that ``RB_ID=#`` correspond to associated messages in commits.
 6.x
 -----
 
+New Features
+~~~~~~~~~~~~
+  * finagle-http: Introduce convience extractors to pattern match `c.t.f.http.Response.status`
```
mosesn: s/convience/convenience

mosesn:
LGTM except for a few nits.

finagle-http/src/test/scala/com/twitter/finagle/http/StatusTest.scala
spockz: `null` isn't a status, therefore, it shouldn't be matched by any of the category matchers.

mosesn: Right, but my impression is that in scala we typically assume that null is not a valid argument to pass.  If an instance might be null, it's up to the user to use it safely.  If that weren't the case, we would need to null-check every argument ever.

We could make the argument that there's no harm in adding a test for null, but since folks often look at tests as example code, I'd be worried that this will end up being cargo-culted.

finagle-http/src/main/scala/com/twitter/finagle/http/Status.scala
```diff
@@ -29,7 +29,54 @@ case class Status(code: Int) {
...
+  /** Matches when the status code is between 500 and 600. */
+  object ServerError {
+    def unapply(status: Status): Option[Status] =
+      inRange(500, 600, status)
+  }
+
+  @inline
+  private[finagle] def inRange(lower: Int, upper: Int,
+                               status: Status): Option[Status] =
+    Option(status).filter(s => s.code >= lower && s.code < upper)
```
mosesn: this should be `Some(status)`, I think.

spockz: Fine by me. :)

taylorleese: I had commented that it should be ```Option(status)``` in case it was null by some chance. That's what I usually do stylistically out of paranoia, but not a huge deal either way.

finagle-http/src/main/scala/com/twitter/finagle/http/Status.scala
```diff
@@ -29,7 +29,54 @@ case class Status(code: Int) {
...
+
 object Status {
+
+  /**
+   * Matches when the status code isn't in range of any of the  categories: {{Informational}},
+   * {{Successful}}, {{Redirection}}, {{ClientError}}, {{ServerError}}.
+   */
+  object UnknownStatus {
+    def unapply(status: Status): Option[Status] =
+      Option(status).filter(s => s.code < 100 || s.code >= 600)
```
taylorleese: This should match line 77 to be consistent. I prefer ```Option```, but I may be the only one so consistency is more important.

finagle-http/src/main/scala/com/twitter/finagle/http/Status.scala
```diff
@@ -29,7 +29,54 @@ case class Status(code: Int) {
...
 }
 
+
 object Status {
+
+  /**
+   * Matches when the status code isn't in range of any of the  categories: {{Informational}},
+   * {{Successful}}, {{Redirection}}, {{ClientError}}, {{ServerError}}.
+   */
+  object UnknownStatus {
```
vkostyukov: How about to rename it just `Unknown`? We already have prefix `Status` so `Status.UnknownStats` feels redundant.

finagle-http/src/main/scala/com/twitter/finagle/http/Status.scala
mosesn: Ah, we typically only use `Option` in finagle when we're using an API that can definitely return `null` sometimes.  We treat `Option` like the Nullable annotation in findbugs–if it's not marked Option, it definitely exists.

finagle-http/src/main/scala/com/twitter/finagle/http/Status.scala
spockz: Good point, fixed.

mosesn:
LGTM

vkostyukov:
I'm pulling this internally. Will close this PR once it's merged.

vkostyukov:
This has been merged internally. Thanks @spockz!

PR #463, opened 2016-02-09
Introduce ServerDispatcher and ServerDispatcherInitializer

penland365:
Motivation:

stack. There is no parallel for a Server.

Solution:

Introduce the ServerDispatcher type, which has two methods dispatch
and handle. Introduce the ServerDispatcherInitializer as well,
which inits Tracing in a Server Dispatcher outside of the stack.

Result:

All dispatcher utilizing the SerialServerDispatcher now record
WireRecv and WireSend. Other dispatchers may utilize the
ServerDispatcherInitializer to record as appropriate.

This is an attempt to find a solution to https://github.com/twitter/finagle/pull/401 , and supersedes my previous attempt at https://github.com/twitter/finagle/pull/448 .


penland365:
This is an ongoing attempted solution to #401 - @mosesn has been very kind in helping me along w/ this. Any issues / faults w/ this PR are definitely mine. 

If this solution looks workable, there are additional changes that need to be made to `finagle-stream` `finagle-memcache` `finagle-thrift`. The hope with this PR is that those changes are pretty painless.

I wanted additional feedback before attempting finagle-mux.

penland365:
I went ahead and updated protocol except for Mux - memcache / stream / thrift / SPDY were pretty easy to change given the current extension point.

penland365:
*nudge*

:wave:

kevinoliver:
@penland365 i'll try to take a look today

kevinoliver:
@penland365 it doesn't apply cleanly to `develop` anymore.

kevinoliver:
Ok, I've skimmed through it and I think the concept seems reasonable.

My main concern which can be easily fixed — there is a good amount of API breaks. I think we can avoid that by having forwarding methods and constructors.

There is are a bunch of style nitpicks that we can visit later.

And don't sell yourself short — this is worthy of a new feature in `finagle/CHANGES`.

penland365:
Thanks @kevinoliver ! I merged in the most recent `develop` pushes - I sorted too many Scala Imports. I'm currently going through the process of updating / building / publishing Util and Scrooge to make sure the merge worked fully.

Loading this back into my memory banks, the other issue to me was around the `TraceInitializerFilter` - for servers, this becomes a bit superfluous since Tracing is initialized at the Dispatcher level instead. I wasn't sure what the best way to address this was.

finagle-core/src/main/scala/com/twitter/finagle/dispatch/ServerDispatcher.scala
```diff
@@ -1,11 +1,65 @@
...
+  *
+  * @tparam Req the Request type of the [[ServerDispatcher]]
+  * @tparam Rep the Response type of the [[ServerDispatcher]]
+  * @param tracer the Tracer to be used
+  * @param fReq a function from `Request` type to 
+  *  [[com.twitter.finagle.tracing.TraceId TraceId]]
+  * @param fRep a function from `Response` type to
+  *  [[com.twitter.finagle.tracing.TraceId TraceId]]
+  */
+//case class ServerDispatcherInitializer[Req, Rep](tracer: Tracer,
```
mosesn: rm this line

finagle-core/src/main/scala/com/twitter/finagle/dispatch/ServerDispatcher.scala
```diff
@@ -1,11 +1,65 @@
...
+  * `WireSend` by the 
+  * [[com.twitter.finagle.transport.Transport transporter]]. Clients should
+  * provided custom behavior via the handle and dispatch methods
+  * @tparam Req the type of Request
+  * @tparam Rep the type of Response
+  * @tparam Out the type of message coming "out" of a
+  *  [[com.twitter.finagle.transport.Transport Transport]]
+  * @param init a [[ServerDispatcherInitializer]]
+  */
+abstract class ServerDispatcher[Req, Rep, Out](init: ServerDispatcherInitializer)
```
mosesn: this doesn't use ServerDispatcherInitializer, as far as I can tell.  let's not make it a constructor parameter until the layer where we need it.

finagle-core/src/main/scala/com/twitter/finagle/dispatch/ServerDispatcher.scala
```diff
@@ -93,22 +133,37 @@ abstract class GenSerialServerDispatcher[Req, Rep, In, Out](trans: Transport[In,
...
   }
 
-  protected def dispatch(req: Req, eos: Promise[Unit]) =
-    service(req) ensure eos.setDone()
+  protected def dispatch(req: Req, eos: Promise[Unit]) = init.fReq(req) match {
+    case Some(traceId) => Trace.letTracerAndId(init.tracer, traceId) {
+      Trace.record(Annotation.WireRecv)
+      service(req) ensure eos.setDone()
+    }
+    case None          => service(req) ensure eos.setDone()
```
mosesn: let's not do this kind of special spacing

finagle-core/src/main/scala/com/twitter/finagle/dispatch/ServerDispatcher.scala
```diff
@@ -93,22 +133,37 @@ abstract class GenSerialServerDispatcher[Req, Rep, In, Out](trans: Transport[In,
...
+  }
 
-  protected def handle(rep: Rep) = trans.write(rep)
+  protected def handle(rep: Rep) = init.fRep(rep) match {
+    case Some(traceId) => Trace.letTracerAndId(init.tracer, traceId) {
+      trans.write(rep).onSuccess(RecordWireSend)
+    }
+    case None          => trans.write(rep)
```
mosesn: ditto

finagle-core/src/main/scala/com/twitter/finagle/dispatch/ServerDispatcher.scala
```diff
@@ -93,22 +133,37 @@ abstract class GenSerialServerDispatcher[Req, Rep, In, Out](trans: Transport[In,
...
+  protected def dispatch(req: Req, eos: Promise[Unit]) = init.fReq(req) match {
+    case Some(traceId) => Trace.letTracerAndId(init.tracer, traceId) {
+      Trace.record(Annotation.WireRecv)
+      service(req) ensure eos.setDone()
+    }
+    case None          => service(req) ensure eos.setDone()
+  }
 
-  protected def handle(rep: Rep) = trans.write(rep)
+  protected def handle(rep: Rep) = init.fRep(rep) match {
```
mosesn: do responses often have the trace id embedded in them?

mosesn:
Sorry for the delay in reviewing, it fell off my radar!  Does this need to be configurable?  Or can we make it private[finagle]?

Thinking about it more, @kevinoliver do you think it would make sense to have a `TransactionContext` object that lives for the space of a request, that can be updated at any time during that request?  We already have the TLS peer certificate, which should be available during this entire time too (although really it's `SessionContext`, not `TransactionContext`).

This looks good, I wonder if we can replace the layers of our protocol implementations where we get the trace id out of the request and call `Trace.letId` with this param.

finagle-core/src/main/scala/com/twitter/finagle/dispatch/ServerDispatcher.scala
penland365: Done

finagle-core/src/main/scala/com/twitter/finagle/dispatch/ServerDispatcher.scala
penland365: I agree - at the time, I thought it would be beneficial for all clients to know that a `ServerDispatcherInitializer` was required to properly construct a SeverDispatcher. Coming back to it a month later though, it just feels like extra noise that should be handled in documentation.

Removed :+1:

finagle-core/src/main/scala/com/twitter/finagle/dispatch/ServerDispatcher.scala
penland365: Removed, sorry, old habits and what not.

finagle-core/src/main/scala/com/twitter/finagle/dispatch/ServerDispatcher.scala
penland365: Done

finagle-core/src/main/scala/com/twitter/finagle/dispatch/ServerDispatcher.scala
penland365: Honestly? ¯\_(ツ)_/¯

I don't have strong opinions on this at all - we have use cases where we like the Trace id embedded in a response, but I can absolutely see this as being something handled on a per-protocol basis.

penland365:
Had a bad merge in [http/Codec.scala](https://github.com/penland365/finagle/blob/server_dispatcher/finagle-http/src/main/scala/com/twitter/finagle/http/Codec.scala) that's been fixed, and I made the changes @mosesn requested. 

No worries at all on the delay!

finagle-core/src/main/scala/com/twitter/finagle/dispatch/ServerDispatcher.scala
mosesn: Could you elaborate on your use case?  Will it ever be a different trace id than the one in the request?

penland365: Yeah, I wasn't thinking clearly here - we use some http services internally that we would like to have the ID embedded, but those would

- Never have a different ID that than the request
- Wouldn't be affected by the SerialServerDispatcher anyways

Thinking back to a month ago, I originally envisioned the `ServerDispatcherInitializer` as a way to tell the Dispatcher whether you wanted any traces returned by making `fRep: (a: Any) => TraceId` optional. That's probably too cute by half.

mosesn:
@penland365 let us know if you need any help on this!

penland365:
screen shot 2016-04-06 at 12 41 17 pm

So, @mosesn @kevinoliver , what do ya'll need from me next on this? I haven't made it `private[finagle]` yet ( wasn't sure if that was a request or not, but it seems reasonable to me ). I have updated the style nits. Given that there has been a new push on develop, I'll get those new changes merged in this afternoon.

penland365:
Just pushed a fresh merge 👍 . It should be up to date w/ latest `develop`.

mosesn:
@penland365 I thought we had agreed that you were going to remove the responseToTraceId bit?

mosesn:
@penland365 just wanted to check in.  how's it going?

fsareshwala:
@penland365 This is a great change. I see it needs a little love so I am going to try and take it over since we're looking for something like this internally as well. I'll fix up the code a bit and submit it -- you'll be the author and I'll be the signer. Thank you for working on this thus far! Great work!

bryce-anderson:
This has gotten pretty stale and there has been no activity for quite a while, so I'm going to close it. If there is revived interest in updating this and getting it merged, feel free to reopen.

PR #467, opened 2016-02-13
Allow control over thrift protocol upgrade.

liamstewart:
Problem

Some thrift server implementations react badly to unknown methods. In
particular, the go implementation throws up its hands and closes the
connection. This makes finagle thrift clients unusable with go servers
as finagle's attempt to upgrade the protocol results in a closed
connection.

Solution

Introduce the ability to control protocol upgrading. This is done
through a new Param (for Stack clients) and a new parameter to
ThriftClient(Buffered|Framed)Codec(Factory)? (for older clients). In
both cases, the preparer simply returns the underlying ServiceFactory
when the parameter is set to false. The parameter default is true to
by default to retain backwards compatibility.

Result

Existing code will continue to run as-is - protocol upgrading will be
attempted. New code gains the ability to turn off upgrading. When
upgrading is off, the upgrade message will not be sent and clients will
work with funky servers.

---

We have an internal hack workaround using a custom codec factory with an `apply` method that returns a new framed codec but with `prepareConnFactory` overridden to simply return `underlying`. This doesn't work with Stack clients so pushing a broader fix back.

I am looking for guidance on two things:

1. Testing. Where / how would be the right way to test this inside finagle?
2. Implementation of avoiding the protocol upgrade in the `Thrift.Client` object. Right now, I switch in the `preparer` method - stats on prepare latency disappear (which I think is right). It seems like it would be nice to have the `stack` method just not do any replacement, but it doesn't have access to params.

### DIFF
--- a/finagle-thrift/src/main/scala/com/twitter/finagle/Thrift.scala
+++ b/finagle-thrift/src/main/scala/com/twitter/finagle/Thrift.scala
@@ -100,6 +100,15 @@ object Thrift extends Client[ThriftClientRequest, Array[Byte]] with ThriftRichCl
     implicit object MaxReusableBufferSize extends Stack.Param[MaxReusableBufferSize] {
       val default = MaxReusableBufferSize(maxThriftBufferSize)
     }
+
+    /**
+     * A `Param` to control upgrading the thrift protocol to TTwitter.
+     * @see The [[https://twitter.github.io/finagle/guide/Protocols.html?highlight=Twitter-upgraded#thrift user guide]] for details on Twitter-upgrade Thrift.
+     */
+    case class AttemptProtocolUpgrade(upgrade: Boolean)
+    implicit object AttemptProtocolUpgrade extends Stack.Param[AttemptProtocolUpgrade] {
+      val default = AttemptProtocolUpgrade(true)
+    }
   }
 
   object Client {
@@ -112,20 +121,25 @@ object Thrift extends Client[ThriftClientRequest, Array[Byte]] with ThriftRichCl
           params: Stack.Params,
           next: ServiceFactory[ThriftClientRequest, Array[Byte]]
         ) = {
-          val Label(label) = params[Label]
-          val param.ClientId(clientId) = params[param.ClientId]
-          val param.ProtocolFactory(pf) = params[param.ProtocolFactory]
-          val Stats(stats) = params[Stats]
-          val preparer = new ThriftClientPreparer(pf, label, clientId)
-          val underlying = preparer.prepare(next, params)
-          new ServiceFactoryProxy(underlying) {
-            val stat = stats.stat("codec_connection_preparation_latency_ms")
-            override def apply(conn: ClientConnection) = {
-              val elapsed = Stopwatch.start()
-              super.apply(conn).ensure {
-                stat.add(elapsed().inMilliseconds)
+          val param.AttemptProtocolUpgrade(upgrade) = params[param.AttemptProtocolUpgrade]
+          if (upgrade) {
+            val Label(label) = params[Label]
+            val param.ClientId(clientId) = params[param.ClientId]
+            val param.ProtocolFactory(pf) = params[param.ProtocolFactory]
+            val Stats(stats) = params[Stats]
+            val preparer = new ThriftClientPreparer(pf, label, clientId)
+            val underlying = preparer.prepare(next, params)
+            new ServiceFactoryProxy(underlying) {
+              val stat = stats.stat("codec_connection_preparation_latency_ms")
+              override def apply(conn: ClientConnection) = {
+                val elapsed = Stopwatch.start()
+                super.apply(conn).ensure {
+                  stat.add(elapsed().inMilliseconds)
+                }
               }
             }
+          } else {
+            next
           }
         }
       }
@@ -177,6 +191,9 @@ object Thrift extends Client[ThriftClientRequest, Array[Byte]] with ThriftRichCl
 
     def withClientId(clientId: thrift.ClientId): Client =
       configured(param.ClientId(Some(clientId)))
+
+    def withAttemptProtocolUpgrade(attemptUpgrade: Boolean): Client =
+      configured(param.AttemptProtocolUpgrade(attemptUpgrade))
 
     def clientId: Option[thrift.ClientId] = params[param.ClientId].clientId
 
--- a/finagle-thrift/src/main/scala/com/twitter/finagle/Thrift.scala
+++ b/finagle-thrift/src/main/scala/com/twitter/finagle/Thrift.scala
@@ -124,8 +124,7 @@ object Thrift extends Client[ThriftClientRequest, Array[Byte]] with ThriftRichCl
           val Label(label) = params[Label]
           val param.ClientId(clientId) = params[param.ClientId]
           val param.ProtocolFactory(pf) = params[param.ProtocolFactory]
-          val param.AttemptProtocolUpgrade(upgrade) = params[param.AttemptProtocolUpgrade]
-          val preparer = new ThriftClientPreparer(pf, label, clientId, false, upgrade)
+          val preparer = new ThriftClientPreparer(pf, label, clientId, false)
           preparer.prepare(next, params)
         }
       }
@@ -178,8 +177,11 @@ object Thrift extends Client[ThriftClientRequest, Array[Byte]] with ThriftRichCl
     def withClientId(clientId: thrift.ClientId): Client =
       configured(param.ClientId(Some(clientId)))
 
-    def withAttemptProtocolUpgrade(attemptUpgrade: Boolean): Client =
-      configured(param.AttemptProtocolUpgrade(attemptUpgrade))
+    def withProtocolUpgrade: Client =
+      configured(param.AttemptProtocolUpgrade(true))
+
+    def withNoProtocolUpgrade: Client =
+      configured(param.AttemptProtocolUpgrade(false))
 
     def clientId: Option[thrift.ClientId] = params[param.ClientId].clientId
 
--- a/finagle-thrift/src/main/scala/com/twitter/finagle/Thrift.scala
+++ b/finagle-thrift/src/main/scala/com/twitter/finagle/Thrift.scala
@@ -124,7 +124,7 @@ object Thrift extends Client[ThriftClientRequest, Array[Byte]] with ThriftRichCl
           val Label(label) = params[Label]
           val param.ClientId(clientId) = params[param.ClientId]
           val param.ProtocolFactory(pf) = params[param.ProtocolFactory]
-          val preparer = new ThriftClientPreparer(pf, label, clientId, false)
+          val preparer = new ThriftClientPreparer(pf, label, clientId)
           preparer.prepare(next, params)
         }
       }
@@ -177,10 +177,10 @@ object Thrift extends Client[ThriftClientRequest, Array[Byte]] with ThriftRichCl
     def withClientId(clientId: thrift.ClientId): Client =
       configured(param.ClientId(Some(clientId)))
 
-    def withProtocolUpgrade: Client =
+    def withAttemptProtocolUpgrade: Client =
       configured(param.AttemptProtocolUpgrade(true))
 
-    def withNoProtocolUpgrade: Client =
+    def withNoAttemptProtocolUpgrade: Client =
       configured(param.AttemptProtocolUpgrade(false))
 
     def clientId: Option[thrift.ClientId] = params[param.ClientId].clientId
--- a/finagle-thrift/src/main/scala/com/twitter/finagle/Thrift.scala
+++ b/finagle-thrift/src/main/scala/com/twitter/finagle/Thrift.scala
@@ -11,7 +11,7 @@ import com.twitter.finagle.thrift.service.ThriftResponseClassifier
 import com.twitter.finagle.thrift.{ClientId => _, _}
 import com.twitter.finagle.tracing.Tracer
 import com.twitter.finagle.transport.Transport
-import com.twitter.util.{Duration, Stopwatch, Monitor}
+import com.twitter.util.{Duration, Monitor}
 import java.net.SocketAddress
 import org.apache.thrift.protocol.TProtocolFactory
 
@@ -121,26 +121,12 @@ object Thrift extends Client[ThriftClientRequest, Array[Byte]] with ThriftRichCl
           params: Stack.Params,
           next: ServiceFactory[ThriftClientRequest, Array[Byte]]
         ) = {
+          val Label(label) = params[Label]
+          val param.ClientId(clientId) = params[param.ClientId]
+          val param.ProtocolFactory(pf) = params[param.ProtocolFactory]
           val param.AttemptProtocolUpgrade(upgrade) = params[param.AttemptProtocolUpgrade]
-          if (upgrade) {
-            val Label(label) = params[Label]
-            val param.ClientId(clientId) = params[param.ClientId]
-            val param.ProtocolFactory(pf) = params[param.ProtocolFactory]
-            val Stats(stats) = params[Stats]
-            val preparer = new ThriftClientPreparer(pf, label, clientId)
-            val underlying = preparer.prepare(next, params)
-            new ServiceFactoryProxy(underlying) {
-              val stat = stats.stat("codec_connection_preparation_latency_ms")
-              override def apply(conn: ClientConnection) = {
-                val elapsed = Stopwatch.start()
-                super.apply(conn).ensure {
-                  stat.add(elapsed().inMilliseconds)
-                }
-              }
-            }
-          } else {
-            next
-          }
+          val preparer = new ThriftClientPreparer(pf, label, clientId, false, upgrade)
+          preparer.prepare(next, params)
         }
       }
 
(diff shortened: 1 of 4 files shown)


kevinoliver:
Thanks for the patch, @liamstewart. I like the idea of it. I'll try to take a look in the next few days.

finagle-thrift/src/main/scala/com/twitter/finagle/Thrift.scala
```diff
@@ -100,15 +100,24 @@ object Thrift extends Client[ThriftClientRequest, Array[Byte]] with ThriftRichCl
     implicit object MaxReusableBufferSize extends Stack.Param[MaxReusableBufferSize] {
       val default = MaxReusableBufferSize(maxThriftBufferSize)
     }
+
+    /**
+     * A `Param` to control protocol upgrading.
```
kevinoliver: mind linking to the docs since this is a bit opaque. something like:

```
@see The [[https://twitter.github.io/finagle/guide/Protocols.html?highlight=Twitter-upgraded#thrift user guide]] for details on Twitter-upgrade Thrift.
```

finagle-thrift/src/main/scala/com/twitter/finagle/thrift/ThriftClientBufferedCodec.scala
```diff
@@ -40,13 +40,15 @@ object ThriftClientBufferedCodec {
     apply()
   }
 
-  def apply(protocolFactory: TProtocolFactory) =
-    new ThriftClientBufferedCodecFactory(protocolFactory)
+  def apply(protocolFactory: TProtocolFactory, _attemptProtocolUpgrade: Boolean = true) =
```
kevinoliver: for backwards compatibility with java users (who don't get the benefits of default arguments) can you add it as forwarding method instead:

```
def apply(protocolFactory: TProtocolFactory) =
  apply(protocolFactory, true)

def apply(protocolFactory: TProtocolFactory, _attemptProtocolUpgrade: Boolean) =
  ...
```

also, since you are editing, can you add return types?

finagle-thrift/src/main/scala/com/twitter/finagle/thrift/ThriftClientBufferedCodec.scala
```diff
@@ -40,13 +40,15 @@ object ThriftClientBufferedCodec {
...
+    new ThriftClientBufferedCodecFactory(protocolFactory, _attemptProtocolUpgrade)
 
 }
 
-class ThriftClientBufferedCodecFactory(protocolFactory: TProtocolFactory) extends
-  CodecFactory[ThriftClientRequest, Array[Byte]]#Client
+class ThriftClientBufferedCodecFactory(
+    protocolFactory: TProtocolFactory,
+    _attemptProtocolUpgrade: Boolean = true)
```
kevinoliver: similar deal here. 

add:

```
def this(protocolFactory: TProtocolFactory) = 
  this(protocolFactory, true)
```

finagle-thrift/src/main/scala/com/twitter/finagle/thrift/ThriftClientBufferedCodec.scala
```diff
@@ -54,11 +56,14 @@ class ThriftClientBufferedCodecFactory(protocolFactory: TProtocolFactory) extend
...
+    new ThriftClientBufferedCodec(protocolFactory, config, _attemptProtocolUpgrade)
   }
 }
 
-class ThriftClientBufferedCodec(protocolFactory: TProtocolFactory, config: ClientCodecConfig)
-    extends ThriftClientFramedCodec(protocolFactory, config) {
+class ThriftClientBufferedCodec(
+    protocolFactory: TProtocolFactory,
+    config: ClientCodecConfig,
+    attemptProtocolUpgrade: Boolean = true)
```
kevinoliver: and here too.

finagle-thrift/src/main/scala/com/twitter/finagle/thrift/ThriftClientFramedCodec.scala
```diff
@@ -27,36 +27,45 @@ object ThriftClientFramedCodec {
 class ThriftClientFramedCodecFactory(
     clientId: Option[ClientId],
     _useCallerSeqIds: Boolean,
-    _protocolFactory: TProtocolFactory)
+    _protocolFactory: TProtocolFactory,
+    _attemptProtocolUpgrade: Boolean)
```
kevinoliver: and here.

finagle-thrift/src/main/scala/com/twitter/finagle/thrift/ThriftClientFramedCodec.scala
```diff
@@ -27,36 +27,45 @@ object ThriftClientFramedCodec {
...
+   * Use the given protocolFactory instead of the default `TBinaryProtocol.Factory`
    */
   def protocolFactory(pf: TProtocolFactory) =
-    new ThriftClientFramedCodecFactory(clientId, _useCallerSeqIds, pf)
+    new ThriftClientFramedCodecFactory(clientId, _useCallerSeqIds, pf, _attemptProtocolUpgrade)
+
+  /**
+   * Enable or disable protocol upgrading.
+   */
+  def attemptProtocolUpgrade(x: Boolean): ThriftClientFramedCodecFactory =
```
kevinoliver: rename `x` to `enabled`.

finagle-thrift/src/main/scala/com/twitter/finagle/thrift/ThriftClientFramedCodec.scala
```diff
@@ -27,36 +27,45 @@ object ThriftClientFramedCodec {
...
+      _protocolFactory, config, clientId, _useCallerSeqIds, _attemptProtocolUpgrade)
 }
 
 class ThriftClientFramedCodec(
   protocolFactory: TProtocolFactory,
   config: ClientCodecConfig,
   clientId: Option[ClientId] = None,
-  useCallerSeqIds: Boolean = false
+  useCallerSeqIds: Boolean = false,
+  attemptProtocolUpgrade: Boolean = true
```
kevinoliver: and here.

kevinoliver:
> Testing. Where / how would be the right way to test this inside finagle?

Typically we'd test this sort of thing in `c.t.f.thrift.EndToEndTest`. I think you can write a test that verifies a few things:
* client has no conn prep time stats
* server doesn't see a ClientId

> Implementation of avoiding the protocol upgrade in the Thrift.Client object. Right now, I switch in the preparer method - stats on prepare latency disappear (which I think is right). It seems like it would be nice to have the stack method just not do any replacement, but it doesn't have access to params.

Not sure I understand your question or what you are looking for here.

liamstewart:
Thanks for the feedback. I'll make changes and look at and end to end test and then poke this PR.

liamstewart:
PTAL

Also added a `withAttemptProtocolUpgrade` method to the `Thrift.Client` class.

kevinoliver:
@liamstewart i won't have time to look at this for a little while, hopefully someone else will get to it.

vkostyukov:
@liamstewart thanks for the PR!

Yesterday, I merged a patch that adds `Stack.Params` argument to the `Codec.prepareConnFactory` so you won't need to propagate your boolean flag all the way down. That patch will be synced on Monday. Let's wait for it and than rebase this PR - it should be simplified a lot.

liamstewart:
Thanks, @vkostyukov. I've taken a look at the new develop branch and rebased my PR. I've made some changes to the way the protocol upgrade attempt is handled to better work with the preparer (so that actions in the preparer that don't involve upgrading the connection, such as your new payload size stats filter and the connection validation filter, are not skipped). This simplifies the code that uses the preparer quite a bit.

finagle-thrift/src/main/scala/com/twitter/finagle/thrift/ThriftClientBufferedCodec.scala
```diff
@@ -26,7 +26,15 @@ object ThriftClientBufferedCodec {
...
    * Create a [[com.twitter.finagle.thrift.ThriftClientBufferedCodecFactory]]
    */
-  def apply() = new ThriftClientBufferedCodecFactory
+  def apply(): ThriftClientBufferedCodecFactory =
+    apply(Protocols.binaryFactory(), true)
+
+  def apply(protocolFactory: TProtocolFactory): ThriftClientBufferedCodecFactory =
+    apply(protocolFactory, true)
+
+  def apply(protocolFactory: TProtocolFactory,
```
vkostyukov: Why we still need this?

finagle-thrift/src/main/scala/com/twitter/finagle/thrift/ThriftClientBufferedCodec.scala
```diff
@@ -39,26 +47,34 @@ object ThriftClientBufferedCodec {
...
-  def apply(protocolFactory: TProtocolFactory) =
-    new ThriftClientBufferedCodecFactory(protocolFactory)
-
 }
 
-class ThriftClientBufferedCodecFactory(protocolFactory: TProtocolFactory) extends
-  CodecFactory[ThriftClientRequest, Array[Byte]]#Client
+class ThriftClientBufferedCodecFactory(
+    protocolFactory: TProtocolFactory,
+    _attemptProtocolUpgrade: Boolean)
```
vkostyukov: I don't think we need ` _attemptProtocolUpgrade` anymore.

finagle-thrift/src/main/scala/com/twitter/finagle/Thrift.scala
```diff
@@ -115,18 +124,9 @@ object Thrift extends Client[ThriftClientRequest, Array[Byte]] with ThriftRichCl
...
-          new ServiceFactoryProxy(underlying) {
-            val stat = stats.stat("codec_connection_preparation_latency_ms")
-            override def apply(conn: ClientConnection) = {
-              val elapsed = Stopwatch.start()
-              super.apply(conn).ensure {
-                stat.add(elapsed().inMilliseconds)
-              }
-            }
-          }
+          val param.AttemptProtocolUpgrade(upgrade) = params[param.AttemptProtocolUpgrade]
```
vkostyukov: Why we need to extract `AttemptProtocolUpgrade` from params and pass it separately? Preparer already takes `params` so you can extract them later.

finagle-thrift/src/main/scala/com/twitter/finagle/thrift/ThriftClientBufferedCodec.scala
```diff
@@ -39,26 +47,34 @@ object ThriftClientBufferedCodec {
...
+    new ThriftClientBufferedCodec(protocolFactory, config, _attemptProtocolUpgrade)
   }
 }
 
-class ThriftClientBufferedCodec(protocolFactory: TProtocolFactory, config: ClientCodecConfig)
-    extends ThriftClientFramedCodec(protocolFactory, config) {
+class ThriftClientBufferedCodec(
+    protocolFactory: TProtocolFactory,
+    config: ClientCodecConfig,
+    attemptProtocolUpgrade: Boolean)
```
vkostyukov: We don't need it here - preparer has it already through `params`.

finagle-thrift/src/main/scala/com/twitter/finagle/thrift/ThriftClientFramedCodec.scala
```diff
@@ -108,19 +127,62 @@ private[thrift] class ThriftClientChannelBufferEncoder
...
+      }
+
+    upgradedService.map { upgraded =>
+      new ValidateThriftService(upgraded, protocolFactory)
+    }
+  }
+
+  def prepare(
+    underlying: ServiceFactory[ThriftClientRequest, Array[Byte]],
+    params: Stack.Params
```
vkostyukov: Just extract your new param here.

vkostyukov:
Thanks @liamstewart! It looks better, but I think there is still a room for improvement. Let's not propagate our new param all over the stack: codec constructor -> prepare connection.Thrift client preparer has a function `prepare` that takes params so you can extract your new param from there.

finagle-thrift/src/main/scala/com/twitter/finagle/thrift/ThriftClientBufferedCodec.scala
liamstewart: My intent with the changes to `ThriftClient(Buffered|Framed)Codec(Factory)` was to provide a way for client code that was still using the builder interface with `.codec(...)` to make use of the new option. I'm more than happy to remove all of them if you guys are ok limiting control of the new feature to `with`-methods / Stack API - that seems to be the case - can you confirm?

vkostyukov: Oh, you're right. I was thinking we can add method to a client builder `attemptProtocolUpgrade`, but realized that it's Thrift-specific. We're planing to remove all the codecs from Finagle in the nearest future (we even have a ticket for that and it's planed to a next tech-debt sprint) so I'm not 100% sure we need to add new functionality there. Let me chat with people to see what they think. I'll get back to you. Thanks, this is a solid work!

vkostyukov: Okay. I chatted with a people at the team. Let's drop the codec support for now and only add stack param available via the `with`-method.

finagle-thrift/src/main/scala/com/twitter/finagle/thrift/ThriftClientFramedCodec.scala
```diff
@@ -28,41 +28,60 @@ object ThriftClientFramedCodec {
...
+   * Use the given protocolFactory instead of the default `TBinaryProtocol.Factory`
    */
   def protocolFactory(pf: TProtocolFactory) =
-    new ThriftClientFramedCodecFactory(clientId, _useCallerSeqIds, pf)
+    new ThriftClientFramedCodecFactory(clientId, _useCallerSeqIds, pf, _attemptProtocolUpgrade)
+
+  /**
+   * Enable or disable protocol upgrading.
+   */
+  def attemptProtocolUpgrade(enabled: Boolean): ThriftClientFramedCodecFactory =
```
vkostyukov: Let's name that `withNoProtocolUpgrade` as per [item 2 in design principles](http://twitter.github.io/finagle/guide/Configuration.html#design-principles). But please, leave the stack param as is.

liamstewart:
Updated, though I left the new `with` methods named `with(No)?AttemptProtocolUpgrade` as that keeps them consistent with their corresponding stack parameter.

vkostyukov:
Awesome! :+1: 

Thanks a ton @liamstewart!

finagle-thrift/src/test/scala/com/twitter/finagle/thrift/EndToEndTest.scala
```diff
@@ -519,6 +520,32 @@ class EndToEndTest extends FunSuite with ThriftTest with BeforeAndAfter {
...
+
+    val sr = new InMemoryStatsReceiver()
+    val client = Thrift.client
+      .configured(Stats(sr))
+      .withProtocolFactory(pf)
+      .withClientId(ClientId("aClient"))
+      .withNoAttemptProtocolUpgrade
+      .newIface[B.ServiceIface](server)
+
+    assert(Await.result(client.someway()) == null)
```
dschobel: please add a short timeout to this await to pre-empt indefinitely blocking tests.

finagle-thrift/src/main/scala/com/twitter/finagle/Thrift.scala
```diff
@@ -100,6 +100,15 @@ object Thrift extends Client[ThriftClientRequest, Array[Byte]] with ThriftRichCl
     implicit object MaxReusableBufferSize extends Stack.Param[MaxReusableBufferSize] {
       val default = MaxReusableBufferSize(maxThriftBufferSize)
     }
+
+    /**
+     * A `Param` to control upgrading the thrift protocol to TTwitter.
+     * @see The [[https://twitter.github.io/finagle/guide/Protocols.html?highlight=Twitter-upgraded#thrift user guide]] for details on Twitter-upgrade Thrift.
+     */
+    case class AttemptProtocolUpgrade(upgrade: Boolean)
```
dschobel: bikeshed: `AttemptTTwitterUpgrade` is more informative imo

finagle-thrift/src/test/scala/com/twitter/finagle/thrift/EndToEndTest.scala
```diff
@@ -519,6 +520,32 @@ class EndToEndTest extends FunSuite with ThriftTest with BeforeAndAfter {
 
     Await.ready(ss.close())
   }
+
+  test("clientId is not sent and prep stats are not recorded when TTwitter upgrading is disabled") {
```
dschobel: this is sufficient I think but if you're feeling motivated, the upgrade request will show up in tracing as well so we could verify that disabling upgrades yields 1 fewer request trace.

mosesn:
LGTM

roanta:
LGTM

adleong:
I have taken over this pull request here: https://github.com/twitter/finagle/pull/474

cacoco:
@adleong we've merged this internally on Friday (with the changes Dan suggested) and will be sync'd on the next push to Github (Monday). Thanks.

mosesn:
merged here: https://github.com/twitter/finagle/commit/563264de172b3443a75b3edd2fb1b06b07e34e22  thanks again!

PR #472, opened 2016-03-03
Support wildcards in Dtab prefixes

olix0r:
Problem

In some cases, it is convenient to simply ignore parts of a path when
processing dtabs.  This is currently quite awkward to accomplish with custom
namers.

Consider the case when a name like the following is processed:

    /http/1.1/GET/host/resource

Solution

Alter the Dtab syntax to support wildcards in prefix paths.  This enables dtabs in the form:

    /http/1.1/* => /h

This would allow the above name to easily be processed to /h/host/resource.

### DIFF
--- a/finagle-core/src/main/scala/com/twitter/finagle/Dtab.scala
+++ b/finagle-core/src/main/scala/com/twitter/finagle/Dtab.scala
@@ -145,7 +145,7 @@ object Dentry {
    * dentry     ::= prefix '=>' tree
    * }}}
    *
-   * where the production `prefix` is from the grammer documented in
+   * where the production `prefix` is from the grammar documented in
    * [[Prefix.read]] and the production `tree` is from the grammar
    * documented in [[com.twitter.finagle.NameTree.read
    * NameTree.read]].
--- a/finagle-core/src/main/scala/com/twitter/finagle/Dtab.scala
+++ b/finagle-core/src/main/scala/com/twitter/finagle/Dtab.scala
@@ -1,6 +1,7 @@
 package com.twitter.finagle
 
 import com.twitter.app.Flaggable
+import com.twitter.io.Buf
 import com.twitter.util.Local
 import java.io.PrintWriter
 import scala.collection.generic.CanBuildFrom
@@ -32,7 +33,7 @@ case class Dtab(dentries0: IndexedSeq[Dentry])
    */
   def lookup(path: Path): NameTree[Name.Path] = {
     val matches = dentries.collect {
-      case Dentry(prefix, dst) if path.startsWith(prefix) =>
+      case Dentry(prefix, dst) if prefix.matches(path) =>
         val suff = path.drop(prefix.size)
         dst.map { pfx => Name.Path(pfx ++ suff) }
     }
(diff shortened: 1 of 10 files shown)


olix0r:
There is more polishing to be done on this branch before it should be merged, but I wanted to socialize an early sketch of this idea before investing time in polishing it.

mosesn:
NB that we can already do `/http/1.1/* => /h` with `/http/1.1 => /h`.  LGTM–my main concern was that it would complicate precedence, but it looks like it won't because we removed the "longest delegation wins" rule.

olix0r:
> NB that we can already do `/http/1.1/* => /h` with `/http/1.1 => /h`

I don't think that's right.  Given a name `/http/1.1/GET/host/resource`:

- `/http/1.1 => /h` produces `/h/GET/host/resource`
- `/http/1.1/* => /h` produces `/h/host/resource`

mosesn:
Ah, I see, you're right of course.

atollena:
lgtm. I think this is a good addition, but how and when to use wildcard vs leave something as a residual isn't that obvious (moses getting confused backs this up). It would be nice to have an example in [the docs](https://github.com/twitter/finagle/blob/develop/doc/src/sphinx/Names.rst#interpreting-paths-with-delegation-tables). There is quite some duplication between `Path` and `Prefix` but I guess you found no easy way to unify them.

olix0r:
@atollena i can probably reuse some code between the two implementations -- i opted to copy/paste to prove it, but I'll spend a little time polishing it and adding docs later today.

finagle-core/src/main/scala/com/twitter/finagle/NameTreeParsers.scala
```diff
@@ -105,6 +106,19 @@ private class NameTreeParsers private (str: String) {
     Buf.ByteArray.Owned(baos.getBuf, 0, baos.size)
   }
 
+  private[this] def isDentryPrefixElemChar(c: Char) = isLabelChar(c) || c == '*'
+
+  private[this] def parseDentryPrefixElem(): Dentry.Prefix.Elem = {
+    peek match {
```
mosesn: ffti: if statement may be more clear here.

finagle-http/src/main/scala/com/twitter/finagle/http/codec/HttpDtab.scala
```diff
@@ -44,17 +44,30 @@ object HttpDtab {
   private def pathFailure(path: String, cause: IllegalArgumentException) =
     Failure("Invalid path: "+path, cause)
 
+  private def prefixFailure(prefix: String, cause: IllegalArgumentException) =
+    Failure("Invalid prefix: "+prefix, cause)
+
   private def nameFailure(name: String, cause: IllegalArgumentException) =
     Failure("Invalid name: "+name, cause)
 
   private def decodePath(b64path: String): Try[Path] =
```
mosesn: unused, we can rm it

finagle-core/src/main/scala/com/twitter/finagle/Dtab.scala
```diff
@@ -154,6 +162,106 @@ object Dentry {
...
+     */
+    def read(s: String): Prefix = NameTreeParsers.parseDentryPrefix(s)
+
+    val empty = new Prefix()
+
+    sealed trait Elem {
+      def show: String
+    }
+
+    object Any extends Elem {
```
mosesn: Let's avoid the `scala.Any` name conflict if possible

finagle-core/src/main/scala/com/twitter/finagle/Path.scala
```diff
@@ -103,6 +84,27 @@ object Path {
     true
   }
 
+  // We're extra careful with allocation here because any time
+  // there are nonbase delegations, we need to serialize the paths
+  // to strings
+  private[finagle] def showElem(buf: Buf): String = {
```
mosesn: can we make this a val'd up function?  as it is, we're entailing an allocation when we placeholder syntax the method right now.

finagle-core/src/main/scala/com/twitter/finagle/Dtab.scala
```diff
@@ -154,6 +162,106 @@ object Dentry {
...
+      else Prefix((elems ++ that.elems):_*)
+    def size: Int = elems.size
+    def isEmpty: Boolean = elems.isEmpty
+
+    def ++(path: Path): Prefix =
+      if (path.isEmpty) this
+      else this ++ Prefix(path)
+
+    lazy val showElems = elems.map(_.show)
+    lazy val show = s"""/${showElems mkString "/"}"""
```
mosesn: How about `showElems.mkString("/")` here?

finagle-core/src/main/scala/com/twitter/finagle/Dtab.scala
```diff
@@ -154,6 +162,106 @@ object Dentry {
...
+    override def toString = s"""Prefix(${showElems mkString ","})"""
```
mosesn: How about `showElems.mkString(",")` here?

finagle-core/src/main/scala/com/twitter/finagle/Path.scala
olix0r: When you say "val'd up function" you mean:
```scala
val showElem: Buf => String = ...
```
?

mosesn: yes

finagle-http/src/main/scala/com/twitter/finagle/http/codec/HttpDtab.scala
olix0r: used on L69

finagle-core/src/main/scala/com/twitter/finagle/Dtab.scala
olix0r: updated both Path and Prefix to use dotted style

atollena:
lgtm.

olix0r:
I understand trepidation around a change like this. How should we proceed?

olix0r:
@atollena If we're all good with this, could you please merge this internally?

finagle-http/src/main/scala/com/twitter/finagle/http/codec/HttpDtab.scala
mosesn: of which file?

mosesn:
LGTM

mosesn:
@olix0r also, can you add a note in CHANGES?

olix0r:
Added a note to CHANGES

finagle-http/src/main/scala/com/twitter/finagle/http/codec/HttpDtab.scala
olix0r: Oh, oh, sorry.  I thought you were referring to prefixFailure -- I'll remove decodePath

olix0r:
Fixed decodePath nit

mosesn:
:+1: pulling it in

finagle-core/src/main/scala/com/twitter/finagle/Dtab.scala
```diff
@@ -154,6 +162,106 @@ object Dentry {
...
+     *
+     * label      ::= (\\x[a-f0-9][a-f0-9]|[0-9A-Za-z:.#$%-_])+
+     *
+     * }}}
+     *
+     * for example
+     *
+     * {{{
+     * /foo/bar/baz
+     * /foo&#47;*&#47;/bar/baz
```
atollena: I'm having trouble interpreting this, what is this supposed to mean? /foo//*//bar/baz?

finagle-core/src/main/scala/com/twitter/finagle/Dtab.scala
```diff
@@ -154,6 +162,106 @@ object Dentry {
...
+          case Prefix.Label(buf) if buf != path.elems(i) =>
+            return false
+          case _ => // matches
+            i += 1
+        }
+      true
+    }
+
+    // A prefix acts somewhat like a Seq[Elem]
+    def take(n: Int) = Prefix(elems.take(n):_*)
```
edma2: I don't see take, drop, ++, etc being used anywhere for prefixes. Any reason to keep them?

olix0r: I was just trying to keep this totally in line with Path api.  I can omit them if you think that would be better

finagle-core/src/main/scala/com/twitter/finagle/Dtab.scala
olix0r: you can't have `/*` or `*/` in a javadoc, since it messes with comment structure.  This renders as ` /foo/*/bar/baz` in rendered javadocs.

atollena: Ha yes! /foo&#47;*&#47;/bar/baz reads as foo/*//bar/baz with a double slash though, no?

finagle-core/src/main/scala/com/twitter/finagle/Dtab.scala
edma2: Keeping it consistent with Path seems reasonable to me. I don't feel strongly about it.

finagle-core/src/main/scala/com/twitter/finagle/Dtab.scala
olix0r: good catch, I'll remove the dupe slash

finagle-core/src/main/scala/com/twitter/finagle/Dtab.scala
```diff
@@ -154,6 +162,106 @@ object Dentry {
...
+     * {{{
+     * /foo/bar/baz
+     * /foo&#47;*&#47;/bar/baz
+     * /
+     * }}}
+     *
+     * parses into the path
+     *
+     * {{{
+     * Prefix(Label(foo),Label(bar),Label(baz))
```
atollena: Also as pointed out by dschoebel it will parse into `Prefix(Label(foo),AnyElem,Label(bar),Label(baz))`

olix0r:
comments fixed, rebased against develop. @atollena

finagle-core/src/main/scala/com/twitter/finagle/Dtab.scala
```diff
@@ -127,26 +128,33 @@ case class Dtab(dentries0: IndexedSeq[Dentry])
...
   /**
    * Parse a Dentry from the string `s` with concrete syntax:
    * {{{
-   * dentry     ::= path '=>' tree
+   * dentry     ::= prefix '=>' tree
    * }}}
    *
-   * where the productions `path` and `tree` are from the grammar
-   * documented in [[com.twitter.finagle.NameTree.read NameTree.read]].
+   * where the production `prefix` is from the grammer documented in
```
edma2: "grammar"

finagle-core/src/main/scala/com/twitter/finagle/Path.scala
```diff
@@ -103,6 +84,27 @@ object Path {
     true
   }
 
+  // We're extra careful with allocation here because any time
+  // there are nonbase delegations, we need to serialize the paths
+  // to strings
+  private[finagle] val showElem: Buf => String = { buf =>
```
edma2: ffti: 
`def showElem(buf: Buf): String`

olix0r: @mosesn told me that the def incurs an unnecessary allocation!

finagle-core/src/main/scala/com/twitter/finagle/Dtab.scala
olix0r: oopsie!

olix0r:
fixed "grammer" nit and CHANGES

finagle-core/src/main/scala/com/twitter/finagle/Path.scala
mosesn: ya

edma2: wow :confounded:

atollena:
This is merged, thank you Oliver.

dschobel:
merged in 1ff06d07e1e53589d828a96238adb6d5fc4db562. thanks @olix0r

PR #473, opened 2016-03-04
Make MapHeaderMap completely case insensitive

grandbora:
Http header field names are case-insensitive ([rfc2616#4.2](https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2)). Therefore [MapHeaderMap](https://github.com/twitter/finagle/blob/develop/finagle-http/src/main/scala/com/twitter/finagle/http/HeaderMap.scala#L60) is designed to be case-insensitive. However only a subset of methods are obeying to this contract.

For example with the current implementation the tests below fail;

```scala
  test("test set and getAll") {
    val map = MapHeaderMap()
    map.set("KEY", "1")
    assert(map.getAll("key") == List("1"))
  }

  test("test +=") {
    val map = MapHeaderMap()
    map += ("KEY" -> "1")
    assert(map.getAll("key") == List("1"))
  }
```
```bash
[info] - test set and getAll *** FAILED ***
[info]   List() did not equal List("1") (MapHeaderMapTest.scala:13)
[info] - test += *** FAILED ***
[info]   List() did not equal List("1") (MapHeaderMapTest.scala:19)

```

There are easier solutions for this problem (details are below), however in order to be backward compatible (to a sufficient degree) this pr introduces a slightly more complicated logic.

In this pr, all the headers are stored with their original header names, but they are grouped by the [canonical name of the header](https://github.com/grandbora/finagle/blob/make-headermap-case-insensitive/finagle-http/src/main/scala/com/twitter/finagle/http/HeaderMap.scala#L65) on write (see `set`, `add` and `+=` methods). 

In order to keep things sane, I created the [HeaderValuePair](https://github.com/grandbora/finagle/blob/make-headermap-case-insensitive/finagle-http/src/main/scala/com/twitter/finagle/http/HeaderMap.scala#L60) abstraction, which is an implementation detail and is not visible to library users.

---- 

For the curious, other possible solutions;

* **Modify on write**
On write, lowercase every header field name (set, add, +=, apply). In [MapHeaderMap#apply](https://github.com/twitter/finagle/blob/develop/finagle-http/src/main/scala/com/twitter/finagle/http/HeaderMap.scala#L113) method this approach has been taken but it is not propagated to other write methods.

* **Merge on read**
 On write, use given header field names as they are. On read, find every entry that matches the given header field name an a case-insensitive manner and merge their values.

* **Use a case-insensitive data structure as underlying storage object**
Delegate the responsibility of handling case-insensitivity to the underlying object. Use something like `new TreeMap<String, String>(String.CASE_INSENSITIVE_ORDER);`

The disadvantage of first approach is that it relies on `.toLowerCase`. In theory [header field names can contain any ISO-8859-1 character](https://tools.ietf.org/html/rfc5987#section-1). I am not sure if there would be some edge cases in the behavior of `.toLowerCase` if a non ASCII character appears in the header field name (not even sure if this is a valid concern).

Second approach doesn't have the disadvantage mentioned above, as long as it uses `String#equalsIgnoreCase` method. However keeping various representations in the underlying storage, for a single header seems like an unnecessary complication.

Third approach can actually be the cleanest implementation, however I don't know if using case-insensitive TreeMap has performance impacts or if it is feasible.

@fwbrasil care to take a look :)

### DIFF
--- a/finagle-http/src/main/scala/com/twitter/finagle/http/HeaderMap.scala
+++ b/finagle-http/src/main/scala/com/twitter/finagle/http/HeaderMap.scala
@@ -1,12 +1,14 @@
 package com.twitter.finagle.http
 
-import com.twitter.finagle.http.netty.HttpMessageProxy
-import com.twitter.util.TwitterDateFormat
 import java.text.SimpleDateFormat
 import java.util.{Date, Locale, TimeZone}
+
+import com.twitter.finagle.http.netty.HttpMessageProxy
+import com.twitter.util.TwitterDateFormat
+
 import scala.annotation.varargs
-import scala.collection.mutable
 import scala.collection.JavaConverters._
+import scala.collection.mutable
 
 /**
  * Mutable message headers map.
@@ -52,67 +54,84 @@ abstract class HeaderMap
   def += (kv: (String, Date)): HeaderMap =
     += ((kv._1, HeaderMap.format(kv._2)))
 
-  override def empty: HeaderMap = new MapHeaderMap(mutable.Map.empty)
+  override def empty: HeaderMap = MapHeaderMap()
 }
 
+private[finagle] case class HeaderValuePair(header: String, value: String) {
+  val canonicalName = HeaderValuePair.canonicalName(header)
+}
+
+private[finagle] object HeaderValuePair {
+  def canonicalName(header: String) = header.toLowerCase(Locale.US)
+}
 
 /** Mutable-Map-backed [[HeaderMap]] */
-class MapHeaderMap(underlying: mutable.Map[String, Seq[String]]) extends HeaderMap {
+class MapHeaderMap extends HeaderMap {
+
+  private[this] val underlying = mutable.Map.empty[String, List[HeaderValuePair]]
+
+  // This is here to provide backward compatibility
+  def this(underlying: mutable.Map[String, Seq[String]]) {
+    this
+    for ((k, vs) <- underlying; v <- vs) yield {
+      add(k, v)
+    }
+  }
 
   def getAll(key: String): Iterable[String] =
-    underlying.getOrElse(key, Nil)
+    underlying.getOrElse(HeaderValuePair.canonicalName(key), Nil).map(_.value)
 
   def add(k: String, v: String): MapHeaderMap = {
-    underlying(k) = underlying.getOrElse(k, Nil) :+ v
+    val t = HeaderValuePair(k, v)
+    underlying(t.canonicalName) = underlying.getOrElse(t.canonicalName, Nil) :+ t
     this
   }
 
   def set(key: String, value: String): MapHeaderMap = {
-    underlying.retain { case (a, _) => !a.equalsIgnoreCase(key) }
-    underlying(key) = Seq(value)
+    val t = HeaderValuePair(key, value)
+    underlying(t.canonicalName) = List(t)
     this
   }
 
   // For Map/MapLike
   def get(key: String): Option[String] = {
-    underlying.find { case (k, v) => k.equalsIgnoreCase(key) }.flatMap { _._2.headOption }
+    getAll(key).headOption
   }
 
   // For Map/MapLike
   def iterator: Iterator[(String, String)] = {
-    for ((k, vs) <- underlying.iterator; v <- vs) yield
-      (k, v)
+    for ((_, vs) <- underlying.iterator; v <- vs) yield
+      (v.header, v.value)
   }
 
   // For Map/MapLike
   def += (kv: (String, String)): MapHeaderMap.this.type = {
-    underlying(kv._1) = Seq(kv._2)
+    val t = HeaderValuePair(kv._1, kv._2)
+    underlying(t.canonicalName) = List(t)
     this
   }
 
   // For Map/MapLike
   def -= (key: String): MapHeaderMap.this.type = {
-    underlying.retain { case (a, b) => !a.equalsIgnoreCase(key) }
+    underlying.remove(HeaderValuePair.canonicalName(key))
     this
   }
 
   override def keys: Iterable[String] =
-    underlying.keys
+    keySet
 
   override def keySet: Set[String] =
-    underlying.keySet.toSet
+    underlying.values.flatten.map(_.header).toSet
 
   override def keysIterator: Iterator[String] =
-    underlying.keysIterator
+    keySet.iterator
 }
-
 
 object MapHeaderMap {
   def apply(headers: Tuple2[String, String]*): MapHeaderMap = {
-    val map = headers
-      .groupBy { case (k, v) => k.toLowerCase }
-      .mapValues { case values => values.map { _._2 } } // remove keys
-    new MapHeaderMap(mutable.Map() ++ map)
+    val tmp = new MapHeaderMap
+    headers.foreach(t => tmp.add(t._1, t._2))
+    tmp
   }
 }
 
--- a/finagle-http/src/main/scala/com/twitter/finagle/http/HeaderMap.scala
+++ b/finagle-http/src/main/scala/com/twitter/finagle/http/HeaderMap.scala
@@ -70,14 +70,6 @@ class MapHeaderMap extends HeaderMap {
 
   private[this] val underlying = mutable.Map.empty[String, List[HeaderValuePair]]
 
-  // This is here to provide backward compatibility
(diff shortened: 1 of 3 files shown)


kevinoliver:
I haven't read the diff yet, but one consideration is how this changes iteration of existing keys. Specifically, if you had a user that did `Map.add("UP", "X")` and then iterated over the keys, they'd now see "up" instead of "UP". 

I _think_ that could in turn break users who need to look at all the headers to calculate a signature, for example in OAuth 1.1.

grandbora:
@kevinoliver thanks for looking into this. You pointed a valid concern. What do you think if I implement the second approach (Merge on read)? Would that be ok?

kevinoliver:
I can't think of a way to do merge on read, but maybe you have ideas.

grandbora:
Hey @kevinoliver , I gave it a shot and introduced a slightly more complex logic to work around the problem you described. I updated the pr description and added some tests for backward compatibility. I'd appreciate if you can have a look.

finagle-http/src/main/scala/com/twitter/finagle/http/HeaderMap.scala
```diff
@@ -52,67 +54,88 @@ abstract class HeaderMap
...
-  override def empty: HeaderMap = new MapHeaderMap(mutable.Map.empty)
+  override def empty: HeaderMap = MapHeaderMap()
 }
 
+private[finagle] case class HeaderValuePair(header: String, value: String) {
+  val canonicalName = HeaderValuePair.canonicalName(header)
+}
+
+private[finagle] object HeaderValuePair {
+  def canonicalName(header: String) = header.toLowerCase
```
kevinoliver: the default encoding will be the platform if unspecified, so we should use `c.t.io.Charsets.UsAscii` i believe.

finagle-http/src/main/scala/com/twitter/finagle/http/HeaderMap.scala
```diff
@@ -52,67 +54,88 @@ abstract class HeaderMap
...
+}
 
 /** Mutable-Map-backed [[HeaderMap]] */
-class MapHeaderMap(underlying: mutable.Map[String, Seq[String]]) extends HeaderMap {
+class MapHeaderMap extends HeaderMap {
+
+  private val underlying = mutable.Map.empty[String, List[HeaderValuePair]]
```
kevinoliver: nitpick: use `private[this]`

finagle-http/src/main/scala/com/twitter/finagle/http/HeaderMap.scala
```diff
@@ -52,67 +54,88 @@ abstract class HeaderMap
...
+
+  // This is here to provide backward compatibility
+  def this(underlying: mutable.Map[String, Seq[String]]) {
+    this
+    for ((k, vs) <- underlying; v <- vs) yield{
```
kevinoliver: style nit: add a space after `yield`

finagle-http/src/main/scala/com/twitter/finagle/http/HeaderMap.scala
```diff
@@ -52,67 +54,88 @@ abstract class HeaderMap
...
   }
 
   // For Map/MapLike
   def -= (key: String): MapHeaderMap.this.type = {
-    underlying.retain { case (a, b) => !a.equalsIgnoreCase(key) }
+    underlying.remove(HeaderValuePair.canonicalName(key))
     this
   }
 
   override def keys: Iterable[String] =
```
kevinoliver: do you think we should document that these will return the keys in a normalized form (here and below)?

finagle-http/src/main/scala/com/twitter/finagle/http/HeaderMap.scala
```diff
@@ -52,67 +54,88 @@ abstract class HeaderMap
...
+}
 
 /** Mutable-Map-backed [[HeaderMap]] */
-class MapHeaderMap(underlying: mutable.Map[String, Seq[String]]) extends HeaderMap {
+class MapHeaderMap extends HeaderMap {
+
+  private val underlying = mutable.Map.empty[String, List[HeaderValuePair]]
+
+  // This is here to provide backward compatibility
+  def this(underlying: mutable.Map[String, Seq[String]]) {
```
kevinoliver: hmm i realize why you left it, but i think its a bit deceptive to the end-user. given that they pass in a mutable map, there is likely some expectation that their modifications to the map are reflected in this `HeaderMap`. but they won't.

this is a big enough change in behavior that i think we should at a minimum mark this as deprecated, and strongly consider breaking the API by removing the constructor.

kevinoliver:
@grandbora this looks like a good approach. i think its worthy of a note in `CHANGES` under the "Runtime Behavior Changes" section, if you could add that as well. or you could call it a "Bug Fix" — either way is fine.

finagle-http/src/main/scala/com/twitter/finagle/http/HeaderMap.scala
grandbora: I was aware of that point, I saw no harm in leaving it to the default because I think as long as it is consistent through out the life cycle of the running app (or most likely within the life cycle of a single request) it *should* be fine. So that was my reasoning.

However I am no expert on this stuff, so tried to add what you suggested. It didn't work out because it expects a `Locale` not a `Charset`. I ended up using `Locale.US`. If there is a twitter convention for this please let me know.

finagle-http/src/main/scala/com/twitter/finagle/http/HeaderMap.scala
grandbora: Correct me if I am missing something here. The purpose of the rework was to make sure that we return the original header names, not the normalized ones. [There is a test case specifically for that](https://github.com/grandbora/finagle/blob/make-headermap-case-insensitive/finagle-http/src/test/scala/com/twitter/finagle/http/MapHeaderMapTest.scala#L113-L124). So behavior here should not change.

finagle-http/src/main/scala/com/twitter/finagle/http/HeaderMap.scala
grandbora: You have a valid point here. I was going to suggest adding a deprecation here. But I personally prefer breaking the api by removing this constructor. Reasons:
* there is already an apply method which takes tuples. I would expect that most people are using that over this constructor.
* this is a pretty easy fix on the users' side. I don't think it worths the trouble of deprecating it.

Let me know which way you want to go with. Happy to do the changes.

grandbora:
I updated the `CHANGES` accordingly.

finagle-http/src/main/scala/com/twitter/finagle/http/HeaderMap.scala
kevinoliver: yeah, you're right.

finagle-http/src/main/scala/com/twitter/finagle/http/HeaderMap.scala
```diff
@@ -52,67 +54,88 @@ abstract class HeaderMap
...
+  def canonicalName(header: String) = header.toLowerCase(Locale.US)
+}
 
 /** Mutable-Map-backed [[HeaderMap]] */
-class MapHeaderMap(underlying: mutable.Map[String, Seq[String]]) extends HeaderMap {
+class MapHeaderMap extends HeaderMap {
+
+  private[this] val underlying = mutable.Map.empty[String, List[HeaderValuePair]]
+
+  // This is here to provide backward compatibility
```
kevinoliver: please either deprecate or remove. as mentioned, i think this API is too weird now.

grandbora: Removed https://github.com/twitter/finagle/pull/473/commits/8f01e6bf3b02809cc60b242ea9d6c1aea1640d26

kevinoliver:
thanks @grandbora. working on pulling this in locally. looks like there is some usage of the mutable map constructor so it may take some time.

kevinoliver:
This just landed in `develop` with commit https://github.com/twitter/finagle/commit/603c0226029f8cc9da2ea628e793e80a91c170ef

Thanks for the contribution @grandbora.

grandbora:
Yay! Looking forward to use this on prod. Thanks @kevinoliver

PR #474, opened 2016-03-04
Allow control over thrift protocol upgrade.

adleong:
Problem

Some thrift server implementations react badly to unknown methods. In
particular, the go implementation throws up its hands and closes the
connection. This makes finagle thrift clients unusable with go servers
as finagle's attempt to upgrade the protocol results in a closed
connection.

Solution

Introduce the ability to control protocol upgrading. This is done
through a new Param (for Stack clients) and a new parameter to
ThriftClient(Buffered|Framed)Codec(Factory)? (for older clients). In
both cases, the preparer simply returns the underlying ServiceFactory
when the parameter is set to false. The parameter default is true to
by default to retain backwards compatibility.

Result

Existing code will continue to run as-is - protocol upgrading will be
attempted. New code gains the ability to turn off upgrading. When
upgrading is off, the upgrade message will not be sent and clients will
work with funky servers.

Also-by: Alex Leong <alex@buoyant.io>

### DIFF
--- a/finagle-thrift/src/main/scala/com/twitter/finagle/Thrift.scala
+++ b/finagle-thrift/src/main/scala/com/twitter/finagle/Thrift.scala
@@ -11,7 +11,7 @@ import com.twitter.finagle.thrift.service.ThriftResponseClassifier
 import com.twitter.finagle.thrift.{ClientId => _, _}
 import com.twitter.finagle.tracing.Tracer
 import com.twitter.finagle.transport.Transport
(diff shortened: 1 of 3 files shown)


dschobel:
thanks @adleong, @cacoco fixed the nits locally on liam's version and is currently bringing it in

cacoco:
Already merged #467. Thanks!

adleong:
Thanks, @cacoco!

Issue #476, opened 2016-03-07
Support server side entity body decoding


nrinaudo:
Finagle currently has support for clients accepting encoded response entities - if a client sets its `Accept-Encoding` to, say, `gzip`, Finagle knows to compress its response and set the `Content-Encoding` header appropriately.

What is not yet supported is the symmetric of this: it's legal, although not terribly common, for clients to submit request entity bodies in a content encoding they know the server to support (`gzip`, for instance, is guaranteed by spec to be supported by all compliant servers). This is not something that Finagle knows to deal with.

The feature appears simple enough to implement - all the necessary bricks are there, even the configuration option, they've just not been assembled. I'd be happy to submit a PR should this be considered a desirable feature.

kevinoliver:
Is this the same issue as in #469? If so, a PR implementing it would be great (cc @adriancole)

nrinaudo:
It absolutely is, you're right. Dont't know how I failed to find it, I'm sorry.
I might be a bit busy this week but I think I can get something done this weekend. It looks like a 2 lines modifications (@vkostyukov pointed me in the right direction).

nrinaudo:
The problem turns out not to be server side decoding implementation - I have something locally that works fine - but client side encoding, which I'll need at least for testing.

From what I managed to piece together, content encoding in netty must go through `HttpContentEncoder`, which is implemented in such a way that it *must* be called in response to a request: if `messageReceived` isn't called, netty doesn't populate its accepted-encoding list and freaks out when asked to encode content. It thinks it's trying to respond to a non-existing request and throws a scary sounding `IllegalStateException("cannot send more responses than requests")`.

I could probably hack together a custom implementation of `HttpContentEncoder`, but before I go down that road, is there a known, clean way of working around that issue? My netty knowledge is next to non-existent, so I might be overlooking something obvious.

vkostyukov:
@nrinaudo In what position do you insert `HttpContentDecompressor`? My take would be to put it right after `HttpChunkAggregator`. I looked through the implementation of `HttpContentDecompressor` and it looks like it should handle streaming requests just fine.

I'm not sure what do you mean by `messageReceiver` isn't called. Do you mean `HttpContentDecompressor` doesn't work if the incoming message isn't encoded?

nrinaudo:
Sorry, I wasn't clear. The server side decompression works fine, it's the *client side* compression that's tricky. The issue I'm having is not with `HttpContentDecompressor` But `HttpContentCompressor`.

[...]

Issue #469, opened 2016-02-24
Support http decompression for servers


adriancole:
Can `Http.Server().withDecompression(true)` be implemented? Currently, this is silently ignored (conceding it works on the client)

It is not obvious how to affect the netty pipeline to add server support for this, without changing finagle classes directly. If there's a clean way to affect the pipeline, that'd work for me, too.

kevinoliver:
Does `Http.server.withCompressionLevel` work for you? (Looks like we could
use some scaladoc on that method and the others in Http.server to explain
what they do).

I think it gets wired into `c.t.f.http.Codec` which in turn wires up
netty's `HttpContentCompressor` for a server.

On Wed, Feb 24, 2016 at 12:33 AM, Adrian Cole <notifications@github.com>
wrote:

> Can (gzip) decompression be added for http servers?
>
> The _decompressionEnabled flag and similar threw me for a loop, as it
> seems to be placed on Http, but only implemented on the client. It is not
> obvious how to affect the netty pipeline to add server support for this,
> without changing finagle classes directly.
>
> If there's a way to hook into the netty pipeline to add this (without
> changing finagle classes), that'd work for me, too.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/twitter/finagle/issues/469>.
>



-- 
Kevin Oliver | Twitter, Inc | follow me: @kevino <http://twitter.com/kevino>

adriancole:
> Does `Http.server.withCompressionLevel` work for you? (Looks like we could
> use some scaladoc on that method and the others in Http.server to explain
> what they do).
>
> I think it gets wired into `c.t.f.http.Codec` which in turn wires up
> netty's `HttpContentCompressor` for a server.

Compression works, iirc. The question here is about decompression. ex can
the netty layer unzip a POST request body.

kevinoliver:
Derp, reading fail on my part.

Looks like it is not wired up but could be done in a similar fashion.

taylorleese:
Can we resolve this based on #478 or is there something else I'm missing here?

vkostyukov:
Yeah, I think it might be closed now. Thanks @nrinaudo for fixing this!

PR #478, opened 2016-03-14
Implement server-side request entity decoding (#476 & #469).

nrinaudo:
Problem

Finagle currently fails to decode legal client-submitted encoded request entities (typically, gzipped or deflated).

See #476 and #469 for more details.

Solution

Hook `HttpContentDecompressor` into server pipelines whenever configured to support it.

### DIFF
--- a/finagle-http/src/test/scala/com/twitter/finagle/http/ServerSideDecodingTest.scala
+++ b/finagle-http/src/test/scala/com/twitter/finagle/http/ServerSideDecodingTest.scala
@@ -3,7 +3,7 @@ package com.twitter.finagle.http
 import java.io.{ByteArrayOutputStream, OutputStream, PrintStream}
 import java.net.InetSocketAddress
 import java.util.zip.{DeflaterOutputStream, GZIPOutputStream}
-
+import com.twitter.conversions.time._
 import com.twitter.finagle
 import com.twitter.finagle.Service
 import com.twitter.finagle.builder.ClientBuilder
@@ -88,7 +88,7 @@ class ServerSideDecodingTest extends FunSuite with GeneratorDrivenPropertyChecks
         .setHeader("Content-Type", "text/plain;charset=utf-8")
         .url(url)
 
-      assert(content == Await.result(client(req.buildPost(encoder.encode(content)))).contentString)
+      assert(content == Await.result(client(req.buildPost(encoder.encode(content))), 1.second).contentString)
     }
   }
 }
--- a/finagle-http/src/test/scala/com/twitter/finagle/http/ServerSideDecodingTest.scala
+++ b/finagle-http/src/test/scala/com/twitter/finagle/http/ServerSideDecodingTest.scala
@@ -16,12 +16,13 @@ import org.scalatest.junit.JUnitRunner
 import org.scalatest.prop.GeneratorDrivenPropertyChecks
 import org.scalatest.{BeforeAndAfterAll, FunSuite}
 
-/** Provides tests for server side content decoding.
-  *
-  * If and when client side compression is implemented, this test should probably be removed in favour of a complete
-  * entry in [[EndToEndTest]]. Client side compression is currently made problematic by netty
-  * (see https://github.com/netty/netty/issues/4970).
-  */
+/**
+ *  Provides tests for server side content decoding.
+ *
+ * If and when client side compression is implemented, this test should probably be removed in favour of a complete
+ * entry in [[EndToEndTest]]. Client side compression is currently made problematic by netty
+ * (see https://github.com/netty/netty/issues/4970).
+ */
 @RunWith(classOf[JUnitRunner])
 class ServerSideDecodingTest extends FunSuite with GeneratorDrivenPropertyChecks with BeforeAndAfterAll {
   // Echo server (with decoding)
--- a/finagle-http/src/test/scala/com/twitter/finagle/http/ServerSideDecodingTest.scala
+++ b/finagle-http/src/test/scala/com/twitter/finagle/http/ServerSideDecodingTest.scala
@@ -19,12 +19,14 @@ import org.scalatest.{BeforeAndAfterAll, FunSuite}
 /**
  *  Provides tests for server side content decoding.
  *
- * If and when client side compression is implemented, this test should probably be removed in favour of a complete
- * entry in [[EndToEndTest]]. Client side compression is currently made problematic by netty
+ * If and when client side compression is implemented, this test should probably
+ * be removed in favour of a complete entry in [[EndToEndTest]]. Client side
+ * compression is currently made problematic by netty
  * (see https://github.com/netty/netty/issues/4970).
  */
 @RunWith(classOf[JUnitRunner])
-class ServerSideDecodingTest extends FunSuite with GeneratorDrivenPropertyChecks with BeforeAndAfterAll {
+class ServerSideDecodingTest extends FunSuite with GeneratorDrivenPropertyChecks
+                                     with BeforeAndAfterAll {
   // Echo server (with decoding)
   val server = finagle.Http.server
     .withLabel("server")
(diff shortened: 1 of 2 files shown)


finagle-http/src/test/scala/com/twitter/finagle/http/ServerSideDecodingTest.scala
```diff
@@ -0,0 +1,94 @@
+package com.twitter.finagle.http
+
+import java.io.{ByteArrayOutputStream, OutputStream, PrintStream}
```
mosesn: please sort your imports

finagle-http/src/test/scala/com/twitter/finagle/http/ServerSideDecodingTest.scala
```diff
@@ -0,0 +1,94 @@
...
+import java.util.zip.{DeflaterOutputStream, GZIPOutputStream}
+
+import com.twitter.finagle
+import com.twitter.finagle.Service
+import com.twitter.finagle.builder.ClientBuilder
+import com.twitter.io.Buf
+import com.twitter.util.{Await, Closable, Future}
+import org.jboss.netty.handler.codec.http.HttpHeaders
+import org.junit.runner.RunWith
+import org.scalacheck.{Arbitrary, Gen}
```
mosesn: love to see these imports 😻

finagle-http/src/test/scala/com/twitter/finagle/http/ServerSideDecodingTest.scala
```diff
@@ -0,0 +1,94 @@
...
+import org.scalatest.junit.JUnitRunner
+import org.scalatest.prop.GeneratorDrivenPropertyChecks
+import org.scalatest.{BeforeAndAfterAll, FunSuite}
+
+/** Provides tests for server side content decoding.
```
mosesn: twitter style guide on scaladoc:

```
/**
 * Blah blah blah
 * @blah blah blah
 */
```

here's an example: https://github.com/twitter/util/blob/develop/util-core/src/main/scala/com/twitter/util/Future.scala#L16-L20

finagle-http/src/test/scala/com/twitter/finagle/http/ServerSideDecodingTest.scala
```diff
@@ -0,0 +1,94 @@
...
+  *
+  * If and when client side compression is implemented, this test should probably be removed in favour of a complete
```
mosesn: wrap at 80 columns please https://github.com/twitter/finagle/blob/develop/CONTRIBUTING.md#style

finagle-http/src/test/scala/com/twitter/finagle/http/ServerSideDecodingTest.scala
```diff
@@ -0,0 +1,94 @@
...
+  // Standard client
+  val client = ClientBuilder()
+    .codec(Http())
+    .hosts(Seq(server.boundAddress.asInstanceOf[InetSocketAddress]))
+    .hostConnectionLimit(1)
+    .build()
+
+  // URL at which all requests should be made
+  val url: String = {
+    val adr = server.boundAddress.asInstanceOf[InetSocketAddress]
```
mosesn: feel free to ignore: the short name we use for address is usually `addr`

finagle-http/src/test/scala/com/twitter/finagle/http/ServerSideDecodingTest.scala
```diff
@@ -0,0 +1,94 @@
...
+  implicit val arbEncoder: Arbitrary[Encoder] = Arbitrary(Gen.oneOf(Gzip, Deflate, Identity))
+
+  test("decode client-side encoded entity bodies") {
+    forAll { (content: String, encoder: Encoder) ⇒
+      val req = RequestBuilder()
+        .setHeader("Content-Encoding", encoder.name)
+        .setHeader("Content-Type", "text/plain;charset=utf-8")
+        .url(url)
+
+      assert(content == Await.result(client(req.buildPost(encoder.encode(content)))).contentString)
```
mosesn: can you add a timeout to `Await.result` so this won't hang even if we regress while developing?

finagle-http/src/test/scala/com/twitter/finagle/http/ServerSideDecodingTest.scala
```diff
@@ -0,0 +1,94 @@
...
+/** Provides tests for server side content decoding.
+  *
+  * If and when client side compression is implemented, this test should probably be removed in favour of a complete
+  * entry in [[EndToEndTest]]. Client side compression is currently made problematic by netty
+  * (see https://github.com/netty/netty/issues/4970).
+  */
+@RunWith(classOf[JUnitRunner])
+class ServerSideDecodingTest extends FunSuite with GeneratorDrivenPropertyChecks with BeforeAndAfterAll {
+  // Echo server (with decoding)
+  val server = finagle.Http.server
```
mosesn: feel free to ignore: I'd just spin up a new client and server for the test if I were you, and then we don't have to worry about spinning it down in an afterAll, we can just do it directly in the test.

mosesn:
LGTM, just a few nits, thanks!

finagle-http/src/test/scala/com/twitter/finagle/http/ServerSideDecodingTest.scala
nrinaudo: This makes sense, but I feel I must point out I failed to do so because `EndToEndTest` mostly does not specify timeouts - I assumed this was the way things were done in Finagle.

By which I do not mean to say I won't fix this problem - I will, you're obviously right - but rather that `EndToEndTest` might need some work.

finagle-http/src/test/scala/com/twitter/finagle/http/ServerSideDecodingTest.scala
nrinaudo: I'm fine with that, but how do you ensure everything is properly closed down if the test fails? `afterAll` has the advantage of being executed even if individual tests throw exceptions, which is why I tend to implement resource cleanup that way when I can. What's the proper Finagle pattern for dealing with that?

nrinaudo:
Mmm, that wasn't very clever, I just overwrote all my questions with that last commit. Let's start over.

1. adding a timeout to `Await.result`: you're absolutely right, and I've updated the code accordingly. I feel I must point out I didn't initially specify a timeout because very few such calls in the tests do set a timeout, and I thought this was the way things were done in Finagle.
2. spawning a new server for each individual test: I'm fine with that, but what's the recommended pattern for enforcing resource cleanup if an exception if thrown during a test? `afterAll` has the advantage of being guaranteed to be called, which is why I tend to use if to free resources when I can, can you point me to some idiomatic Finagle code that deals with this issue?

finagle-http/src/test/scala/com/twitter/finagle/http/ServerSideDecodingTest.scala
mosesn: :+1: yeah, our style has evolved since we first started adding end to end tests, but we typically make these kinds of style fixes lazily.

mosesn:
@nrinaudo if the test throws an exception we have bigger fish to fry :wink:.  internally, we can't merge in code that makes our tests fail, and the tests will get torn down at the end of the day, so resource consumption isn't that bad.

nrinaudo:
@mosesn understood. I've updated the PR accordingly, let me know if you feel anything else is dodgy.

mosesn:
LGTM

vkostyukov:
Looks great! Thanks @nrinaudo!

mosesn:
We merged this internally, it should show  up on the `develop` branch on Monday.  Thanks!

mosesn:
@nrinaudo boom: https://github.com/twitter/finagle/commit/7b5eaaebf4a9b7c11808207af51bd3001f0ef609  thanks so much!

PR #479, opened 2016-03-15
finagle-ci: Updates finagle to not sudo anymore

mosesn:
Just hanging out to get travised


mosesn:
Nope, still doesn't work :(

PR #480, opened 2016-03-16
Added Round Robin load balancer for comparison purposes and improved Simulation.scala

stevej:
Problem

We wanted to compare Round Robin load balancing with the choices available in Finagle.

Solution

We added a simple round robin loadbalancer to the finagle-core package along with a unit test suite.

We also added improvements to Simulation.scala such as printing individual latency numbers, adding a slow client in the middle of a simulation run.

We added a .rr() function to Balancers.scala as well as a newline to .aperture()'s comment to improve the scaladoc output.

We added a type constraint to DistributorT#This to enforce by the compiler what was being done by convention.

We changed protection modifiers to several member classes and traits of loadbalancer package to allow running their tests from sbt.

Result
  * Added Round Robin load balancer and unit tests
  * Added .rr() to Balancers
  * Changed protection modifiers to several member classes and traits of loadbalancer package
  * Modified Simulation to allow a slow client in the middle.
  * Modified Simulation to allow tracing latency per request.
  * Added a shell script for running Simulation.scala from a shell.
  * Added some R code in a literate file to help me build nice graphs.
  * Added a newline for improved scaladoc in Balancers.scala

### DIFF
--- a/finagle-benchmark/src/main/scripts/build_graphs.R
+++ b/finagle-benchmark/src/main/scripts/build_graphs.R
@@ -1,0 +1,91 @@
+This is not really an R script, it is a literate transcript of the R
+functions I wrote to generate a graph combined with some commentary on
+the data generated. How I use this script is to copy and paste
+snippets into my R workspace.
+
+# returns a data.frame with response times from different load balancers, sampled at the given rate.
+buildData <- function(rr, p2c, ewma, samples=1000) {
+    rr <- read.table(rr)
+    p2c <- read.table(p2c)
+    ewma <- read.table(ewma)
+    data = data.frame(seq(1, samples),
+         sample(rr$V1, samples),
+         sample(p2c$V1, samples),
+         sample(ewma$V1, samples))
+
+    names(data) <- c("id", "rr", "p2c", "ewma")
+
+    return(data)
+}
+
+allGood <- buildData(rr="all_good_1000_rr.txt", p2c="all_good_1000_p2c.txt", ewma="all_good_1000_ewma.txt")
+
+slowMiddle <- buildData(rr="slow_middle_1000_rr.txt", p2c="slow_middle_1000_p2c.txt", ewma="slow_middle_1000_ewma.txt")
+
+slowStart <- buildData(rr="slow_start_1000_rr.txt", p2c="slow_start_1000_p2c.txt", ewma="slow_start_1000_ewma.txt")
+
+# Overlapping ewma vs rr histograms
+> nines = c(0.5, 0.9, 0.95, 0.96, 0.97, 0.98, 0.99, 0.999, 0.9999, 0.99999)
+> quantile(slowMiddle$p2c, nines)
+     50%      90%      95%      96%      97%      98%      99%    99.9%   99.99%  99.999%
+ 174.000  184.000  189.000  191.000  193.000  197.000  205.000 1723.004 1825.201 1907.291
+> quantile(slowMiddle$ewma, nines)
+     50%      90%      95%      96%      97%      98%      99%    99.9%   99.99%  99.999%
+ 174.000  183.000  188.000  189.000  191.000  194.000  199.000  211.000 1780.103 1819.110
+> quantile(slowMiddle$rr, nines)
+    50%     90%     95%     96%     97%     98%     99%   99.9%  99.99% 99.999%
+ 175.00  187.00  203.00 1664.00 1673.00 1683.00 1714.00 1862.00 1977.70 1994.56
+
+### How to think about the latency graphic where Round Robin jumps up
+to 2 seconds after 95%
+
+What does it mean for our service? If your timeout was 1 second, your
+success rate dropped to 95% during this period of time if you used
+round robin. Your success rate only drops to 99.9% if you used EWMA.
+99% if you used queue-depth as your metric.
+
+We setup a load balancing simulation with repsonse latencies from some
+collected ping data. The median was 167ms with a stddev of 5ms and a
+max of 200ms.
+
+10 good clients and 1 client that slows to 2 seconds in the middle of
+the test run and then recovers back to normal. This acts similar to a
+bad GC pause in a backend. No timeouts enabled and no retries.
+
+[TODO: find a way to do this automatically]
+I hand-converted those quantiles into the following file (rr_latencies.txt):
+
+Please note that the file uses tabs.
+
+percentile         queue       ewma rr
+"90"               184         183  187
+"95"               189         188  203
+"97"               193         191  1673
+"98"               197         194  1683
+"99"               205         199  1714
+"99.9"             1723        211  1862
+"99.99"            1825        1780 1977
+"99.999"           1907        1819 1994
+
+Here's how we generate the png of response times.
+
+rr_latencies <- read.table("rr_latencies.txt", header=TRUE)
+# turn this into a new dataframe for my own sanity. read.table is not being helpful
+df <- data.frame(percentile=rr_latencies$percentile, ewma=rr_latencies$ewma, queue=rr_latencies$queue, roundrobin=rr_latencies$rr)
+# builds an x-axis along the number of 9s we care about.
+# the graph looks nuts along the automatically inferred scale so we'll force a continuous scale.
+df$nines <- abs(log10(1.0 - df$percentile / 100))
+
+png("rr_latencies.png", width=960, height=960)
+theme_set(theme_bw(base_size = 27) + theme(legend.key.size=unit(35, "pt")))
+ggplot(data=df, aes(x=nines, y=roundrobin)) +
+       scale_x_continuous(breaks=df$nines,
+                     labels=c("90", "95", "97", "98", "99", "99.9", "99.99", "99.999")) +
+       geom_point(aes(y=roundrobin, col="Round Robin")) +
+       geom_line(aes(y=roundrobin, col="Round Robin")) +
+       geom_point(aes(y=ewma, col="EWMA")) +
+       geom_line(aes(y=ewma, col="EWMA")) +
+       geom_point(aes(y=queue, col="Least Loaded")) +
+       geom_line(aes(y=queue, col="Least Loaded")) +
+       labs(title="Latency by Load Balancer", x="Percentile", y="Latency (ms)", color="Balancers")
+dev.off()
--- a/finagle-core/src/main/scala/com/twitter/finagle/loadbalancer/Balancers.scala
+++ b/finagle-core/src/main/scala/com/twitter/finagle/loadbalancer/Balancers.scala
@@ -123,6 +123,7 @@ object Balancers {
    *  1. The concurrent load, measured over a window specified by
    *     `smoothWin`, to each service stays within the load band, delimited
    *     by `lowLoad` and `highLoad`.
+   *
    *  2. Services receive load proportional to the ratio of their
    *     weights.
    *
@@ -153,6 +154,32 @@ object Balancers {
           gauge.remove()
           super.close(when)
         }
+      }
+    }
+  }
+
+  /**
+   * A simple round robin balancer that chooses the next backend in
+   * the list for each request.
+   *
+   * WARNING: Unlike other balancers available in finagle, this does
+   * not take latency into account and will happily direct load to
+   * slow or oversubscribed services. We recommend using one of the
+   * other load balancers for typical production use.
+   *
+   * @param maxEffort the maximum amount of "effort" we're willing to
+   * expend on a load balancing decision without reweighing.
+   */
+  def rr(
+    maxEffort: Int = MaxEffort
+  ): LoadBalancerFactory = new LoadBalancerFactory {
+    def newBalancer[Req, Rep](
+      endpoints: Activity[Set[ServiceFactory[Req, Rep]]],
+      sr: StatsReceiver,
+      exc: NoBrokersAvailableException
+    ): ServiceFactory[Req, Rep] = {
+      new RoundRobinBalancer(endpoints, sr, exc) {
+        private[this] val gauge = sr.addGauge("rr")(1)
       }
     }
   }
--- a/finagle-core/src/main/scala/com/twitter/finagle/loadbalancer/RoundRobin.scala
+++ b/finagle-core/src/main/scala/com/twitter/finagle/loadbalancer/RoundRobin.scala
@@ -1,0 +1,99 @@
+package com.twitter.finagle.loadbalancer
+
+import com.twitter.finagle.{ClientConnection, NoBrokersAvailableException, Service,
+  ServiceFactory, ServiceFactoryProxy, Status}
+import com.twitter.finagle.service.FailingFactory
+import com.twitter.finagle.stats.StatsReceiver
+import com.twitter.finagle.util.OnReady
+import com.twitter.util.{Activity, Future, Promise, Time}
+
+/**
+ * A simple round robin balancer that chooses the next backend in
+ * the list for each request.
+ */
+class RoundRobinBalancer[Req, Rep](
+  val activity: Activity[Traversable[ServiceFactory[Req, Rep]]],
+  val statsReceiver: StatsReceiver,
+  val emptyException: NoBrokersAvailableException,
+  val maxEffort: Int = 5
+) extends ServiceFactory[Req, Rep]
+    with Balancer[Req, Rep]
+    with Updating[Req, Rep] {
+
+  // For the OnReady mixin
+  private[this] val ready = new Promise[Unit]
+  override def onReady: Future[Unit] = ready
+
+  protected[this] val maxEffortExhausted = statsReceiver.counter("max_effort_exhausted")
+
+  override def apply(conn: ClientConnection): Future[Service[Req,Rep]] = {
+    dist.pick()(conn)
+  }
+
+  protected class Node(val factory: ServiceFactory[Req, Rep])
+      extends ServiceFactoryProxy[Req,Rep](factory)
+      with NodeT[Req,Rep] { self =>
+    type This = Node
+    // Note: These stats are never updated.
+    override def load = 0.0
+    override def pending = 0
+    override def token = 0
+
+    override def close(deadline: Time): Future[Unit] = factory.close(deadline)
+    override def apply(conn: ClientConnection): Future[Service[Req,Rep]] = factory(conn)
+  }
+
+  /**
+    * A simple round robin distributor.
+    */
+  protected class Distributor(val vector: Vector[Node])
+      extends DistributorT[Node] {
+    type This = Distributor
+
+    @volatile private[this] var sawDown = false
+    private[this] var currentNode = 0
+
+    private[this] val nodeUp: Node => Boolean = {_.isAvailable}
+    private[this] val nodeDown: Node => Boolean = {!_.isAvailable}
+
+    private[this] val (up, down) = vector.partition(nodeUp) match {
+      case (Vector(), down) => (down, Vector.empty)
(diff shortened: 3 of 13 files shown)


finagle-benchmark/src/main/scripts/build_graphs.R
```diff
@@ -0,0 +1,91 @@
...
+### How to think about the latency graphic where Round Robin jumps up
+to 2 seconds after 95%
+
+What does it mean for our service? If your timeout was 1 second, your
+success rate dropped to 95% during this period of time if you used
+round robin. Your success rate only drops to 99.9% if you used EWMA.
+99% if you used queue-depth as your metric.
+
+We setup a load balancing simulation with repsonse latencies we
+gathered from pinging two datacenters. Ping times were about 175ms
```
roanta: It's not really relevant if this is between data centers or not. I'm not sure if it was, looking at some notes from steve gury who collected the data, it just says "distribution of ping latency on some remote server". If we want more control over the distribution we can eventually synthesize the latency in Simulation to fit a pre-defined distribution.

roanta:
Just a minor nit, otherwise LGTM. Thanks so much for the great contribution!

finagle-benchmark/src/main/scala/com/twitter/finagle/loadbalancer/RoundRobin.scala
```diff
@@ -0,0 +1,84 @@
+package com.twitter.finagle.loadbalancer
+
+import com.twitter.finagle.{ClientConnection, NoBrokersAvailableException, Service, ServiceFactory}
+import com.twitter.finagle.service.FailingFactory
+import com.twitter.finagle.stats.StatsReceiver
+import com.twitter.finagle.util.OnReady
+import com.twitter.util.{Activity, Future, Promise, Time}
+
+object RoundRobinBalancer {
```
mosesn: how would you feel about moving this into finagle-core?  we have a few use cases which could use this for real.

finagle-benchmark/src/main/scala/com/twitter/finagle/loadbalancer/RoundRobin.scala
```diff
@@ -0,0 +1,84 @@
...
+  def rr(): LoadBalancerFactory =
+    new LoadBalancerFactory {
+      def newBalancer[Req, Rep](
+        endpoints: Activity[Set[ServiceFactory[Req, Rep]]],
+        sr: StatsReceiver,
+        exc: NoBrokersAvailableException
+      ): ServiceFactory[Req, Rep] = {
+        new RoundRobinBalancer(endpoints, sr, exc) {
+          private[this] val gauge = sr.addGauge("rr")(1)
```
mosesn: ffti: use a counter and decrement on close?

finagle-benchmark/src/main/scripts/build_graphs.R
stevej: Ok, I've updated this paragraph to describe the data (median, stddev, max) instead.

finagle-benchmark/src/main/scala/com/twitter/finagle/loadbalancer/RoundRobin.scala
stevej: Sure, I can move it and improve the scaladoc to match what you'd expect from something in finagle-core. I'm surprised you have a serious use case for Round Robin that isn't better served by almost any other choice currently in finagle. Would you also want it exposed easily as Balancers.rr?

roanta: I think Moses is referring to a few systems that want to optimize for equity (and have other means of detecting and mitigating the effects of unhealthy nodes). Although I don't think we need round-robin for that. We could use P2C with a constant cost per-node. I would actually vote for keeping this here.

finagle-benchmark/src/main/scala/com/twitter/finagle/loadbalancer/RoundRobin.scala
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  /**
+    * A simple round robin distributor.
+    */
+  protected class Distributor(val vector: Vector[Node])
+      extends DistributorT {
+    type This = Distributor
+
+    var currentNode = 0
```
dschobel: private[this]

finagle-benchmark/src/main/scala/com/twitter/finagle/loadbalancer/RoundRobin.scala
mosesn: Why use P2C with a constant cost per-node?  There's no benefit, right?  RoundRobin is very simple, it probably adds about as much complexity as P2C with constant cost per-node, and it seems easier to reason about.

finagle-benchmark/src/main/scala/com/twitter/finagle/loadbalancer/RoundRobin.scala
```diff
@@ -0,0 +1,84 @@
...
+
+      vector(nodeToPick)
+    }
+
+    override def needsRebuild: Boolean = false
+    override def rebuild(): This = new Distributor(vector)
+    override def rebuild(vector: Vector[Node]): This = new Distributor(vector)
+  }
+
+  protected def maxEffort: Int = Int.MaxValue
```
dschobel: why expend this much effort as opposed to a single scan over the nodes and getting to the rebuild sooner?

finagle-benchmark/src/main/scala/com/twitter/finagle/loadbalancer/RoundRobin.scala
roanta: The only benefit is having fewer implementations to maintain. I'd just prefer waiting to promote it into core until we are sure that it solves the use cases you have in mind.

mosesn: The RoundRobin implementation is simple enough that I don't think it significantly complicates things.  I'll ask someone from search to take a look.  Happy to merge this in as-is.

finagle-benchmark/src/main/scala/com/twitter/finagle/loadbalancer/RoundRobin.scala
stevej: I went ahead and set it to 5 which is the default in Balancers. Based on your comment, I'd ideally set it to the size of the vector but I don't have access to that here. Maybe activity.values.size? That doesn't seem totally right to me.

dschobel:
lgtm, thanks for the new lb!

mosesn:
looks good to me too!

finagle-benchmark/src/main/scala/com/twitter/finagle/loadbalancer/RoundRobin.scala
cacoco: I'm with @mosesn on moving this into finagle-core. I have at least one (major) use-case that would benefit from this and we've been hoping to get this LB for a while. cc @roanta

roanta: Sounds good to me if @stevej is up for it.

stevej:
I'm working on moving this to finagle-core and adding tests today.

stevej:
While writing a unit test suite, I ended up hardening RoundRobin a bit. Most of the test scenarios were originally borrowed from P2CBalancerTest.

I also added a type constraint to DistributorT#This to enforce by the compiler what we had been expecting users to do by convention.

sbt needs wider access permissions for some of the traits/classes in the loadbalancer package in order to compile and run their tests. This is probably not an issue with your own build system, just with the open source release.

finagle-core/src/main/scala/com/twitter/finagle/loadbalancer/Balancers.scala
```diff
@@ -156,4 +157,30 @@ object Balancers {
       }
     }
   }
+
+  /**
+   * A simple round robin balancer that chooses the next backend in
+   * the list for each request.
+   *
+   * WARNING: Unlike other balancers available in finagle, this does
+   * not take latency into account and will happily direct load to
```
roanta: latency or endpoint status.

roanta: whoops, just saw that you added it to the implementation.

finagle-core/src/main/scala/com/twitter/finagle/loadbalancer/RoundRobin.scala
```diff
@@ -0,0 +1,95 @@
...
+    override def pick(): Node = {
+      if (vector.isEmpty)
+        return failingNode(emptyException)
+
+      var node = vector(chooseNext())
+
+      if (node.status != Status.Open) {
+        // If we have an error, we'll rebuild so don't worry about the
+        // wheel, just find any good node to return or else throw.
+        sawDown = true
```
roanta: why do need to rebuild if we find a down node?

finagle-core/src/test/scala/com/twitter/finagle/loadbalancer/RoundRobinBalancerTest.scala
```diff
@@ -0,0 +1,160 @@
...
+
+@RunWith(classOf[JUnitRunner])
+class RoundRobinBalancerTest extends FunSuite with App with RoundRobinSuite {
+  case class LoadedFactory(id: Int) extends RRServiceFactory {
+    @volatile var stat: Status = Status.Open
+    var load = 0
+    var sum = 0
+    var count = 0
+
+    // This isn't quite the right notion of mean load, but it's good enough.
```
roanta: How so? Seems correct to me.

roanta:
Nice! Thanks for taking the time to do this. 

LGTM, just a few questions.

finagle-core/src/test/scala/com/twitter/finagle/loadbalancer/RoundRobinBalancerTest.scala
stevej: That's a good question, I cribbed this method from P2CBalancerTest and marius wrote this comment in 1179a23b. I mean, it seems fine to me.

roanta: I think it's vestigial from when the balancers supported weights.

finagle-core/src/main/scala/com/twitter/finagle/loadbalancer/RoundRobin.scala
stevej: If I rebuild, will the Closed node still be in the list? If yes, then rebuilding is not an effective strategy here. The lifecycle for the list of nodes is not entirely clear to me. 

If a rebuild is the wrong choice here, I can either partition the list or move the index forward one and keep trying but I'll need some way to signal that all nodes are down in that case or else we'll spin forever. I could also do what I've done here (find any node ready to serve traffic) and the index will be updated on the next call and not set sawBegin to true.

roanta: Yes, the status != open node will still be in the list as written. Rebuild only signals to the `Balancer` trait to schedule a call to your distributors rebuild method (with the same nodes). It is effectively like calling `new Distributor(vector)` but it is serialized with other updates from the dynamic `Activity[Seq[Nodes]]`.

We should do something similar to what we do in P2C. Partition the nodes into up and down and rebuild if you see a down node in up or a node has recovered in down. If we need to, we should be able to maintain a consistent ring across rebuilds by sorting the nodes via the token field.

stevej: Ok, I'm now partitioning into down and up Vectors and triggering rebuild if a member of either should be moved. That's a lot nicer and separates concerns neatly. If somebody really needs a consistent ring across rebuilds, I think it's best if they add that when they have specific requirements.

finagle-core/src/main/scala/com/twitter/finagle/loadbalancer/RoundRobin.scala
```diff
@@ -0,0 +1,105 @@
...
+import com.twitter.finagle.stats.StatsReceiver
+import com.twitter.finagle.util.OnReady
+import com.twitter.util.{Activity, Future, Promise, Time}
+
+/**
+ * A simple round robin balancer that chooses the next backend in
+ * the list for each request.
+ */
+class RoundRobinBalancer[Req, Rep](
+  protected val activity: Activity[Traversable[ServiceFactory[Req, Rep]]],
```
mosesn: do these need to be protected val?  do we have to specify a visibility for them, or is it OK to be default / private?

finagle-core/src/main/scala/com/twitter/finagle/loadbalancer/RoundRobin.scala
```diff
@@ -0,0 +1,105 @@
...
+
+  override def apply(conn: ClientConnection): Future[Service[Req,Rep]] = {
+    dist.pick()(conn)
+  }
+
+  protected class Node(val factory: ServiceFactory[Req, Rep])
+      extends ServiceFactoryProxy[Req,Rep](factory)
+      with NodeT[Req,Rep] { self =>
+    type This = Node
+    // We don't care about these metrics for round robin
```
mosesn: we don't use them for balancing, but they could be useful for metrics, since they're used to calculate load on a service.  maybe worth mentioning that the stats we export for "load" is not right.

https://github.com/twitter/finagle/blob/develop/finagle-core/src/main/scala/com/twitter/finagle/loadbalancer/Balancer.scala#L94-L96

finagle-core/src/main/scala/com/twitter/finagle/loadbalancer/RoundRobin.scala
```diff
@@ -0,0 +1,105 @@
...
+      case (Vector(), down) => (down, Vector.empty)
+      case updown => updown
+    }
+
+    // For each node that's requested, we move the currentNode index
+    // around the wheel using mod arithmetic. This is the round robin
+    // of our balancer.
+    private def chooseNext(): Int = {
+      synchronized {
+        currentNode = (currentNode + 1) % up.size
```
mosesn: ffti: cache up.size?  I'm not sure if vector.size is efficient.

finagle-core/src/main/scala/com/twitter/finagle/loadbalancer/RoundRobin.scala
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    /**
+      * Will only return Nodes that have Status.Open
+      */
+    override def pick(): Node = {
+      if (up.isEmpty)
+        return failingNode(emptyException)
+
+      var node = up(chooseNext())
```
mosesn: we don't reassign node, let's make this a val

finagle-core/src/main/scala/com/twitter/finagle/loadbalancer/RoundRobin.scala
```diff
@@ -0,0 +1,105 @@
...
+        currentNode = (currentNode + 1) % up.size
+        currentNode
+      }
+    }
+
+    /**
+      * Will only return Nodes that have Status.Open
+      */
+    override def pick(): Node = {
+      if (up.isEmpty)
```
mosesn: I don't think having multiple exit points for this method makes it easier to read.

rewriting it to avoid multiple exits:

```scala
if (up.isEmpty) failingNode(emptyException)
else {
  val node = up(chooseNext())
  node.status match {
    case Status.Open => node
    case _ => up.find(_.isAvailable).getOrElse(failingNode(emptyException)
  }
}
```

seems just as clear to me.

finagle-core/src/test/scala/com/twitter/finagle/loadbalancer/RoundRobinBalancerTest.scala
```diff
@@ -0,0 +1,159 @@
...
+    val ml = fs.head.meanLoad
+    for (f <- fs) {
+      assert(math.abs(f.meanLoad - ml) < ε,
+        "ml=%f; f.ml=%f; ε=%f".format(ml, f.meanLoad, ε))
+    }
+  }
+}
+
+@RunWith(classOf[JUnitRunner])
+class RoundRobinBalancerTest extends FunSuite with App with RoundRobinSuite {
```
mosesn: do we need App here?  looks like we need it in p2c because we parse flags, but since we don't here, we shouldn't need it.

mosesn:
just a few more nits.

finagle-core/src/test/scala/com/twitter/finagle/loadbalancer/RoundRobinBalancerTest.scala
stevej: thanks, removed App.

finagle-core/src/main/scala/com/twitter/finagle/loadbalancer/RoundRobin.scala
stevej: Thanks, I changed that.

finagle-core/src/main/scala/com/twitter/finagle/loadbalancer/RoundRobin.scala
stevej: this is now gone due to the other change you suggested.

finagle-core/src/main/scala/com/twitter/finagle/loadbalancer/RoundRobin.scala
stevej: I'm digging into whether vector.size is efficient.

stevej: Vector.size looks efficient, it is integer subtraction.

https://github.com/scala/scala/blob/2.11.x/src/library/scala/collection/immutable/Vector.scala#L81

finagle-core/src/main/scala/com/twitter/finagle/loadbalancer/RoundRobin.scala
stevej: the convention of the other balancers is to make these protected. I don't know why that choice was made but it's not strictly required so I've reverted back to default access modifiers for RoundRobinBalancer

finagle-core/src/main/scala/com/twitter/finagle/loadbalancer/RoundRobin.scala
stevej: noted in the comment now.

mosesn:
LGTM

mosesn:
@stevej some feedback I've gotten so far internally:

Can we change the metric name to `round_robin` instead of `rr`?
Can we change the method name to `roundRobin` instead of `rr`?
Can we change the name of the `updown` variable to `updog`?

If you're OK with these changes, I can make them and merge it in, if not we can talk about it and figure something out :)

cc @kevinoliver @dschobel

stevej:
I am totally fine with those changes. (please remember to change Simluation which calls Balancer.rr)

updog is good. I'll inform the office about this change.

mosesn:
This was merged internally, thanks so much for the contribution!

mosesn:
@stevej this hit the develop branch today!  https://github.com/twitter/finagle/commit/fc6452179168db49797ae620fafe1f70206b9aaf  Really appreciate you working with us to get it in, this will be really useful for us too. :100: :sparkling_heart:

PR #481, opened 2016-03-18
Insert DelayedReleaseService into the HTTP client stack

### LINKED ISSUE #471: streaming client appears to hang when making a second request
We've observed a streaming http client hang when more than one streaming request is made.

Example code to observe this behavior:
```scala
Http.server
  .withStreaming(enabled = true)
  .serve("0.0.0.0:8080", Service.mk[Request, Response] { req =>
    val writable = Reader.writable() // never gets closed
    Future.value(Response(req.version, Status.Ok, writable))
})

val client = Http.client.withStreaming(enabled = true).newService(s"/$$/inet/localhost/8080")

val rsp0 = Await.result(client(Request()), 2.seconds)
// rsp0 stream is never closed
println("first request succeeded")

val rsp1 = Await.result(client(Request()), 2.seconds)
// hangs
println("second request succeeded")
```

For full context, we observed this in our Kubernetes namer, which watches for changes using this API:
`/api/v1/namespaces/foo/endpoints?watch=true`
We worked around it by creating a new client for each streaming request:
https://github.com/BuoyantIO/linkerd/pull/137

olix0r:
Problem

FactoryToService releases HTTP services before their response stream has
completed.  This causes busy connections to be returned to the available
connection pool prematurely such that multiple streaming requests may not be
processed simultaneously by an Http service.

Solution

Use the StackClient.Role.prepFactory slot as it was intended: to prevent
premature closure of a Service.  The http.DelayedRelease module wraps all
services in a http.DelayedReleaseService to ensure that close() is not called
on the underlying service until chunked responses have been read or discarded.

Fixes #471

Also-by: Andrew Seigner <siggy@buoyant.io>

### DIFF
--- a/finagle-http/src/main/scala/com/twitter/finagle/Http.scala
+++ b/finagle-http/src/main/scala/com/twitter/finagle/Http.scala
@@ -4,7 +4,7 @@ import com.twitter.conversions.storage._
 import com.twitter.finagle.client._
 import com.twitter.finagle.dispatch.GenSerialClientDispatcher
 import com.twitter.finagle.filter.PayloadSizeFilter
-import com.twitter.finagle.http.{HttpClientTraceInitializer, HttpServerTraceInitializer, HttpTransport, Request, Response}
+import com.twitter.finagle.http.{DelayedRelease, HttpClientTraceInitializer, HttpServerTraceInitializer, HttpTransport, Request, Response}
 import com.twitter.finagle.http.codec.{HttpClientDispatcher, HttpServerDispatcher}
 import com.twitter.finagle.http.filter.{ClientContextFilter, DtabFilter, HttpNackFilter, ServerContextFilter}
 import com.twitter.finagle.netty3._
@@ -106,6 +106,9 @@ object Http extends Client[Request, Response] with HttpRichClient
   object Client {
     val stack: Stack[ServiceFactory[Request, Response]] =
       StackClient.newStack
+        // XXX this prepConn is not exactly equivalent with prepareConnFactory in Codec.scala
+        .replace(StackClient.Role.prepConn, DelayedRelease.module)
+        .replace(StackClient.Role.prepFactory, DelayedRelease.module)
         .replace(TraceInitializerFilter.role, new HttpClientTraceInitializer[Request, Response])
         .prepend(http.TlsFilter.module)
         .prepend(nonChunkedPayloadSize)
--- a/finagle-http/src/main/scala/com/twitter/finagle/http/DelayedReleaseService.scala
+++ b/finagle-http/src/main/scala/com/twitter/finagle/http/DelayedReleaseService.scala
@@ -1,10 +1,24 @@
 package com.twitter.finagle.http
 
+import com.twitter.finagle.client.StackClient
 import com.twitter.finagle.util.AsyncLatch
-import com.twitter.finagle.{Service, ServiceProxy}
+import com.twitter.finagle._
 import com.twitter.io.{Buf, Reader, Writer}
 import com.twitter.util.{Future, Promise, Return, Throw, Time}
 import java.util.concurrent.atomic.AtomicBoolean
+
+private[finagle] object DelayedRelease {
+  val role = StackClient.Role.prepFactory
+  val description = "Prevents an HTTP service from being closed until its response completes"
+  val module: Stackable[ServiceFactory[Request, Response]] =
+    new Stack.Module1[FactoryToService.Enabled, ServiceFactory[Request, Response]] {
+      val role = DelayedRelease.role
+      val description = DelayedRelease.description
+      def make(_enabled: FactoryToService.Enabled, next: ServiceFactory[Request, Response]) =
+        if (_enabled.enabled) next.map(new DelayedReleaseService(_))
+        else next
+    }
+}
 
 /**
  * Delay release of the connection until all chunks have been received.
@@ -13,13 +27,13 @@ private[finagle] class DelayedReleaseService[-Req <: Request](
   service: Service[Req, Response]
 ) extends ServiceProxy[Req, Response](service) {
 
-  protected[this] val counter = new AsyncLatch
+  protected[this] val latch = new AsyncLatch
 
   private[this] def proxy(in: Response) = {
     val released = new AtomicBoolean(false)
     def done() {
       if (released.compareAndSet(false, true)) {
-        counter.decr()
+        latch.decr()
       }
     }
 
@@ -46,19 +60,19 @@ private[finagle] class DelayedReleaseService[-Req <: Request](
   }
 
   override def apply(request: Req): Future[Response] = {
-    counter.incr()
+    latch.incr()
     service(request) transform {
       case Return(r) if r.isChunked =>
         Future.value(proxy(r))
       case t =>
-        counter.decr()
+        latch.decr()
         Future.const(t)
     }
   }
 
   override final def close(deadline: Time): Future[Unit] = {
     val p = new Promise[Unit]
-    counter.await { p.become(service.close(deadline)) }
+    latch.await { p.become(service.close(deadline)) }
     p
   }
 
--- a/CHANGES
+++ b/CHANGES
@@ -7,6 +7,11 @@ Note that ``RB_ID=#`` correspond to associated messages in commits.
 6.x
 -----
 
+Runtime Behavior Changes
+~~~~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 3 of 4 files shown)


finagle-http/src/test/scala/com/twitter/finagle/http/HttpStreamingEndToEndTest.scala
```diff
@@ -0,0 +1,35 @@
+package com.twitter.finagle.http
+
+import com.twitter.conversions.time._
+import com.twitter.finagle.{Http => _Http, Service}
+import com.twitter.io.{Buf, Reader}
+import com.twitter.util.{Await, Future, JavaTimer}
+import org.scalatest.FunSuite
+import java.net.InetSocketAddress
+
+class HttpStreamingEndToEndTest extends FunSuite {
```
luciferous: What do you think of adding this to [StreamingTest.scala](https://github.com/twitter/finagle/blob/develop/finagle-http/src/test/scala/com/twitter/finagle/http/StreamingTest.scala)?

finagle-http/src/main/scala/com/twitter/finagle/Http.scala
```diff
@@ -107,6 +107,7 @@ object Http extends Client[Request, Response] with HttpRichClient
     val stack: Stack[ServiceFactory[Request, Response]] =
       StackClient.newStack
         .replace(TraceInitializerFilter.role, new HttpClientTraceInitializer[Request, Response])
+        .replace(StackClient.Role.prepFactory, DelayedRelease.module)
```
luciferous: Do you also need `prepConn`?

I was going to propose making these changes look more like ClientBuilder, e.g.:

```
    val clientStack = {
      val stack0 = stack
        .replace(StackClient.Role.prepConn, prepConn)
        .replace(StackClient.Role.prepFactory, (next: ServiceFactory[Req, Rep]) =>
        codec.prepareServiceFactory(next))
```

But I think I like your way better, because it means less reliance on `Codec`.

finagle-http/src/main/scala/com/twitter/finagle/http/DelayedReleaseService.scala
```diff
@@ -1,11 +1,25 @@
...
 
+import com.twitter.finagle.client.StackClient
 import com.twitter.finagle.util.AsyncLatch
-import com.twitter.finagle.{Service, ServiceProxy}
+import com.twitter.finagle._
 import com.twitter.io.{Buf, Reader, Writer}
 import com.twitter.util.{Future, Promise, Return, Throw, Time}
 import java.util.concurrent.atomic.AtomicBoolean
 
+object DelayedRelease {
```
vkostyukov: Let's make that `private[finagle]` by analogy with `DelayedReleaseService`.

finagle-http/src/main/scala/com/twitter/finagle/http/DelayedReleaseService.scala
```diff
@@ -13,13 +27,38 @@ private[finagle] class DelayedReleaseService[-Req <: Request](
...
 
-  protected[this] val counter = new AsyncLatch
+  protected[this] val latch = new AsyncLatch
 
-  private[this] def proxy(in: Response) = {
+  override def apply(request: Req): Future[Response] = {
+    latch.incr()
+    service(request) transform {
+      case Return(r) if r.isChunked =>
+        Future.value(DelayedReleaseResponse(r, latch.decr))
```
vkostyukov: What do you think about moving `DelayedReleaseResponse.apply` here as `def private[this]`? This would not only reduce the diff, but would allow as to access `latch` directly (w/o passing a closure) so we can save those 8 bytes per closure allocation on each chunked request.

dschobel:
lgtm after vladimir's DelayedRelease visibility suggestion. 

thanks @olix0r / @siggy, that was a bad broken window.

finagle-http/src/main/scala/com/twitter/finagle/http/DelayedReleaseService.scala
olix0r: Yeah, I can revert this.  DelayedReleaseResponse seemed like a nice utility to have outside of this implementation, but now that I'm no longer using it in two places it seems best to merge back..

mosesn:
LGTM with the caveat that this won't let folks who are managing their connection manually close them the way they would expect, but that's *probably* OK.

finagle-http/src/main/scala/com/twitter/finagle/Http.scala
olix0r: `prepConn` seems to be orthogonal:

```scala
    val prepConn = new Stack.ModuleParams[ServiceFactory[Req, Rep]] {
      override def parameters: Seq[Stack.Param[_]] = Nil
      override val role = StackClient.Role.prepConn
      override val description = "Connection preparation phase as defined by a Codec"
      def make(ps: Stack.Params, next: ServiceFactory[Req, Rep]) = {
        val Stats(stats) = ps[Stats]
        val underlying = codec.prepareConnFactory(next, ps)
        new ServiceFactoryProxy(underlying) {
          val stat = stats.stat("codec_connection_preparation_latency_ms")
          override def apply(conn: ClientConnection) = {
            val begin = Time.now
            super.apply(conn) ensure {
              stat.add((Time.now - begin).inMilliseconds)
            }
          }
        }
      }
    }
```

Let's leave this for another change, if at all

olix0r:
@mosesn You can't close a Service today(!) because FactoryToService ignores closes.  If you're actually managing your own connections, you are likely using the ServiceFactory interface (via newClient), where none of this applies.

finagle-http/src/main/scala/com/twitter/finagle/Http.scala
luciferous: Leaving prepConn out will mean HTTP clients built using Stack and ClientBuilder have be different – see: https://github.com/twitter/finagle/blob/develop/finagle-http/src/main/scala/com/twitter/finagle/http/Codec.scala#L154.

This comment is more for posterity than recommending the work be done in this PR. Eventually, I think we should reconcile the gap.

olix0r:
Updated per feedback

mosesn:
@olix0r oh, I see, because if your FactoryToServices aren't installed you don't have to worry about FactoryToService.  I get it now, thanks!

finagle-http/src/main/scala/com/twitter/finagle/Http.scala
olix0r: Oh, interesting. finagle-http's codec _does_ install DelayedReleaseService in the prepConn:
```scala

      override def prepareConnFactory(
        underlying: ServiceFactory[Request, Response],
        params: Stack.Params
      ): ServiceFactory[Request, Response] =
        // Note: This is a horrible hack to ensure that close() calls from
        // ExpiringService do not propagate until all chunks have been read
        // Waiting on CSL-915 for a proper fix.
        underlying.map { u =>
          val filters =
            new ClientContextFilter[Request, Response].andThenIf(!_streaming ->
              new PayloadSizeFilter[Request, Response](
                params[param.Stats].statsReceiver, _.content.length, _.content.length
              )
            )

          filters.andThen(new DelayedReleaseService(u))
        }
```

Perhaps we should install DelayedRelease.module in prepConn as well?

luciferous: Right, the relevant (to this issue) difference between Stack and ClientBuilder clients is that ClientBuilder has access to the codec, via which it can inject codec-specific logic as prepConn and prepFactory. In the HTTP codec, this is implemented as DelayedReleaseService wrapping.

olix0r:
Added DelayedRelease to the prepConn phase as well so that it is more in line with the logic in http/Codec.scala

finagle-http/src/main/scala/com/twitter/finagle/Http.scala
```diff
@@ -106,6 +106,9 @@ object Http extends Client[Request, Response] with HttpRichClient
   object Client {
     val stack: Stack[ServiceFactory[Request, Response]] =
       StackClient.newStack
+        // XXX this prepConn is not exactly equivalent with prepareConnFactory in Codec.scala
```
luciferous: fingers crossed

luciferous:
LGTM

mosesn:
LGTM

dschobel:
this landed in d928ba5f01175e1ab8e1eebbaa5d461252a6f2c4. thanks @olix0r / @siggy

PR #486, opened 2016-03-25
finagle-http: Encode local dtabs in the Dtab-Local header

### LINKED ISSUE #485: Change HttpDtab to encode Dtab-local headers
In August 2014, finagle-http changed to read `Dtab-local` headers (https://github.com/twitter/finagle/commit/88ede12847c996547381a2b990b3aaf4f8f99759).  The intent of this change was that we would later be able to stop encoding `X-Dtab` headers (which are substantially more complicated).

I would like to change HttpDtab.write to encode `Dtab-local` headers.  `X-Dtab` headers should be considered deprecated at this point, though we will continue to support _reading_ these headers.

I'll prepare a change.  Please feel free to raise questions before I do so, @kevinoliver @mosesn @atollena @edma2.

olix0r:
Problem

Finagle's HTTP server may read Dtabs in two forms: the `Dtab-local` header and `X-Dtab-NN-[A|B]` header pairs.  The latter was designed before the Dtab syntax stabilized; and `Dtab-local` was introduced some to simplify this in 88ede12847c996547381a2b990b3aaf4f8f99759.  The old `X-Dtab` is more complicated and is no longer necessary, since the `Dtab-local` header has been supported since August 2014.

Solution

Change Finagle's HTTP client to encode `Dtab-local` headers.  The old `X-Dtab` syntax may still be read by servers, but clients will set only the `Dtab-local` header.

Fixes #485

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -7,8 +7,15 @@ Note that ``RB_ID=#`` correspond to associated messages in commits.
 6.x
 -----
 
+Runtime Behavior Changes
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+  * finagle-http: Local Dtabs are now encoded into the `Dtab-Local` header.  `X-Dtab` headers
+    may still be read but should be considered deprecated.
+
 New Features
 ~~~~~~~~~~~~
+
   * finagle-http: new stack params MaxChunkSize, MaxHeaderSize, and MaxInitialLineLength
     are available to configure the http codec. ``RB_ID=811129``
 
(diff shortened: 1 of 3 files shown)


mosesn:
LGTM

atollena:
lgtm, I'm pulling this internally.

atollena:
This was merged and will be in the next release, thanks @olix0r .

mosesn:
Thanks, made it into develop here: https://github.com/twitter/finagle/commit/21a1ba9cbfc8ae1106d024d3681cece5495dfc8a !

PR #491, opened 2016-04-04
Admit comments in Dtabs

olix0r:
Problem

There is no way to annotate Dtabs with commentary.

Solution

Support line-oriented comments beginning with '#'.

NameTreeParser's `eatWhitespace` method has been modified to consume lines or
line-endings once the '#' character is encountered.

### DIFF
--- a/doc/src/sphinx/Names.rst
+++ b/doc/src/sphinx/Names.rst
@@ -161,8 +161,9 @@ is bound by Finagle to the Internet address ``localhost:8080``. Similarly,
 is the path describing the serverset_ ``/foo/bar`` on the ZooKeeper
 ensemble ``zk.local.twitter.com:2181``.
 
-Dtabs may contain line-oriented comments beginning with ``#``.  For
-example, this Dtab with commentary:
+Dtabs may contain line-oriented comments beginning with ``#``. ``#``
+must be preceded by a whitespace character or delimeter such as ``;``,
+``|``, or ``&``.  For example, this Dtab with commentary:
 
 ::
 
@@ -176,7 +177,7 @@ is equivalent to this Dtab without commentary:
 
 ::
 
-       /s => /t | (/u & /v);
+       /s => /a | (/b & /c);
 
 We use dtabs to define how logical names (e.g. ``/s/crawler``)
 translate into addresses. Because rewriting is abstracted away, we can
--- a/doc/src/sphinx/Names.rst
+++ b/doc/src/sphinx/Names.rst
@@ -161,6 +161,23 @@ is bound by Finagle to the Internet address ``localhost:8080``. Similarly,
 is the path describing the serverset_ ``/foo/bar`` on the ZooKeeper
 ensemble ``zk.local.twitter.com:2181``.
 
+Dtabs may contain line-oriented comments beginning with ``#``.  For
+example, this Dtab with commentary:
+
+::
+
+        # delegation for /s
+        /s => /a      # prefer /b
+            | ( /b    # or share traffic between /b and /c
+              & /c
+              );
+
+is equivalent to this Dtab without commentary:
+
+::
+
+       /s => /t | (/u & /v);
+
 We use dtabs to define how logical names (e.g. ``/s/crawler``)
 translate into addresses. Because rewriting is abstracted away, we can
 adapt a Finagle process to its environment by manipulating its dtab.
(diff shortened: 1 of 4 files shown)


olix0r:
cc @atollena @edma2 @mosesn 

I'll update docs, etc, if this is seems useful.

This is something that I've implemented in various clients so that we can have comments in our stored dtabs.  It seems useful to just put this into the Dtab syntax directly...

atollena:
This is going to be useful for us, thanks for contributing that back.

lgtm, pending notes in CHANGES and doc changes.

olix0r:
rebased against new develop.  Added CHANGES and documentation in Names.rst.  Let me know if additional documentation would be helpful.

dschobel:
lgtm

finagle-core/src/main/scala/com/twitter/finagle/NameTreeParsers.scala
```diff
@@ -55,8 +55,17 @@ private class NameTreeParsers private (str: String) {
   }
 
   private[this] def eatWhitespace() {
-    while (!atEnd && str(idx).isWhitespace)
+    while (!atEnd && (str(idx).isWhitespace || str(idx) == '#')) {
+      if (str(idx) == '#') eatLine()
```
kevinoliver: what does this do to dtabs using #s?

doc/src/sphinx/Names.rst
```diff
@@ -161,6 +161,23 @@ is bound by Finagle to the Internet address ``localhost:8080``. Similarly,
...
+        /s => /a      # prefer /b
+            | ( /b    # or share traffic between /b and /c
+              & /c
+              );
+
+is equivalent to this Dtab without commentary:
+
+::
+
+       /s => /t | (/u & /v);
```
atollena: nit: `/s => /a | (/b & /c)`

doc/src/sphinx/Names.rst
```diff
@@ -161,6 +161,23 @@ is bound by Finagle to the Internet address ``localhost:8080``. Similarly,
 is the path describing the serverset_ ``/foo/bar`` on the ZooKeeper
 ensemble ``zk.local.twitter.com:2181``.
 
+Dtabs may contain line-oriented comments beginning with ``#``.  For
```
mosesn: maybe add a note that `#` must be preceded by a space or must be the first thing in a line?

mosesn:
LGTM

finagle-core/src/main/scala/com/twitter/finagle/NameTreeParsers.scala
edma2: Using a non-dtab character (like ;) for comments would avoid this ambiguity.

olix0r: As illustrated in the test case, /#foo names are safe. The comment character has to be found in the context of whitespace and not in a name. eatWhitespace isn't called while processing names.

Re: character, `;` is meaningful as a delimiter which Is problematic:
```
/goo => /car ; comment
          | /cab ;
```
Doesn't work.

I'm open to other suggestions but # seems like a natural comment char.

kevinoliver: i saw the test case after i commented.

seems reasonable to me, but i'll defer to others working more closely on dtabs.

doc/src/sphinx/Names.rst
```diff
@@ -161,6 +161,24 @@ is bound by Finagle to the Internet address ``localhost:8080``. Similarly,
 is the path describing the serverset_ ``/foo/bar`` on the ZooKeeper
 ensemble ``zk.local.twitter.com:2181``.
 
+Dtabs may contain line-oriented comments beginning with ``#``. ``#``
+must be preceded by a whitespace character or delimeter such as ``;``,
```
gtcampbell: typo: "delimeter" should be "delimiter"

dschobel: this is up for review internally so I'll fix it there.

finagle-core/src/main/scala/com/twitter/finagle/NameTreeParsers.scala
mariusae: it's a little confusing to treat comments in this way.  how about using '%'?

kevinoliver:
This was merged in 4d6dfc1

PR #493, opened 2016-04-15
Fix typo in FailureAccrualFactory example

samstarling:
Problem

The documentation refers to `FailureAccrual.Param`, which does not exist.

Solution

Correct the reference so that it is `FailureAccrualFactory.Param`.


vkostyukov:
Nice catch! Thanks @samstarling!

mosesn:
LGTM

vkostyukov:
This was merged internally and will show up here on Monday. Thanks for the contribution, @samstarling!

PR #494, opened 2016-04-16
Clarify argument names in SessionPoolingParams

olix0r:
Problem

The argument name `sessionsPerHost` is misleading. The value represents the
total number of sessions across the entire load balancing pool and not a single
host.

Solution

Change the argument name to `sessions`.


olix0r:
Am I right about this?

olix0r:
Nevermind, I think I'm wrong about this.

PR #495, opened 2016-04-18
Fix histogram snapshotting after a lapse in snapshotting     

olix0r:
Problem

When snapshotting does not occur for serveral minutes, the 60-second snapshotting is not honored. If snapshotting does not occur for two minutes, histogram windowin will not be properly honored until snapshot has been called twice.

Solution

Change snapshot windowing to be based on the current time.
 
Fixes twitter/twitter-server#33

### DIFF
--- a/finagle-stats/src/test/scala/com/twitter/finagle/stats/MetricsBucketedHistogramTest.scala
+++ b/finagle-stats/src/test/scala/com/twitter/finagle/stats/MetricsBucketedHistogramTest.scala
@@ -127,20 +127,79 @@ class MetricsBucketedHistogramTest extends FunSuite {
       h.add(1)
 
       // initial user access to start histogram snapshots
-      val init = details.counts
-      assert(init == Seq(BucketAndCount(1, 2, 1)))
+      assert(details.counts == Seq(BucketAndCount(1, 2, 1)))
 
       // call .snapshot() to recompute counts
       h.snapshot()
-      val countsSnap0 = details.counts
-      assert(countsSnap0 == Seq(BucketAndCount(1, 2, 1)))
+      assert(details.counts == Seq(BucketAndCount(1, 2, 1)))
 
       h.add(Int.MaxValue)
       // roll to window 2 (this should make data A visible after a call to snapshot)
       roll()
-      val countsSnap1 = details.counts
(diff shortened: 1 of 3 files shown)


olix0r:
@yokoshin i put together a proof of concept test for your change.  Please review and let me know if it looks right.

kevinoliver:
lgtm @olix0r

yokoshin:
lgtm @olix0r

finagle-stats/src/test/scala/com/twitter/finagle/stats/MetricsBucketedHistogramTest.scala
```diff
@@ -132,15 +132,63 @@ class MetricsBucketedHistogramTest extends FunSuite {
...
 
       h.add(Int.MaxValue)
       // roll to window 2 (this should make data A visible after a call to snapshot)
       roll()
-      val countsSnap1 = details.counts
-      assert(countsSnap1 == Seq(BucketAndCount(1, 2, 1),
+      assert(details.counts == Seq(BucketAndCount(1, 2, 1),
         BucketAndCount(2137204091, Int.MaxValue, 1)))
+
+      // Advance several minutes without snapshotting (i.e. to
```
mosesn: break this off into its own test?

mosesn:
LGTM except for a nit in the test

finagle-stats/src/test/scala/com/twitter/finagle/stats/MetricsBucketedHistogramTest.scala
olix0r: The test name, _histogram snapshot respects refresh window_, seemed appropriate...  i'll split it out if that's clearer, though.

mosesn: I don't mind the test name, but I don't like tests that are too long.  it makes it hard to tell what has broken when a test fails, and it's harder to tell what a test is trying to do when you read the code.

olix0r:
updated test

mosesn:
LGTM

cacoco:
This has been merged internally and should be pushed on the next sync.

kevinoliver:
looks like this landed in 186f7a42f but we messed up and lost the commit info :( 
apologies.

Issue #496, opened 2016-04-18
Feature Request / Question: Load balancer with no rotation


johanstenberg92:
Hi,

I'm using Finagle for my Master's Thesis project and so far, I'm really liking it! Thanks for a great library.

My thesis involves creating a distributed, resilient in-memory database, which is using strong eventual consistency. With strong eventual consistency, I want to be able to create a finagle client, tell it it's closest replicas but since the synchronization between replicas aren't instant, I don't want any rotation (think that you write a value then read it and it isn't there...). However I still want the client to auto-magically change replica if the first replica is down.

I've whipped up a very simple load balancer which sticks with the same node until it needs to change because of the status of the node has changed, would this be an interesting PR or maybe it's too trivial?

Finally, I would like to know if I should approach this problem any other way, I haven't found any other way in your docs. It might be semantically odd to have it reconnect to another replica instantly.

Thanks!

vkostyukov:
Thanks a lot for sharing this @johanstenberg92, this is really interesting!

I wonder if having a tiny wrapper around `ServiceFactory` returned from `$Ptotocol.newClient` would allow you to achieve that. When you ask for a client (`newClient: ServiceFactory`) not a service (`newService: Service`), it provides you more fine grained control over the underlying session (represented as a `Service`), which maps to a concrete replica and concrete connection. So basically you can create one and used it until `status != open` so you can recreate that and use again.

Hope that's not complete non-sense.

olix0r:
I had a similar use case recently, and @mosesn suggested that the _aperture_ load balancer could be used with `maxEffort = Int.MaxValue (see https://github.com/BuoyantIO/linkerd/pull/300).  This should cause _all_ traffic to be sent to a single endpoint as long as that endpoint is available. When it is no longer available another endpoint in the load balancing pool will be used.

johanstenberg92:
@vkostyukov and @olix0r thanks for replying. I've solved it by creating my own load-balancer.

Would a PR be interesting to take a look at @vkostyukov? With a new load balancer? Otherwise you are free to close this!

mosesn:
I think it would be useful.  A few people have asked about it, so clearly there's a use for it.

johanstenberg92:
@mosesn cool, I'll create a PR in the next 24h and reference this issue, cheers!

roanta:
I agree with @vkostyukov, why can't we build this on top of a client using the ServiceFactory API? That's the intended use case.

mariusae:
[...]

PR #497, opened 2016-04-20
Add Sticky Load Balancer which connects to one host and sticks to it.

johanstenberg92:
Problem

If you use Finagle for services which aren't identical replicas such as
systems using strong eventual consistency, you can't currently use any provided
load balancers since they treat all hosts as replicas.

Solution

This PR adds a new load balancer sticking to the first backend it can connect
to. It only changes backend if the initial backend is not reachable. This allows
Finagle clients to be used in systems using (strong) eventual consistency.

### DIFF
--- a/finagle-core/src/main/scala/com/twitter/finagle/loadbalancer/Sticky.scala
+++ b/finagle-core/src/main/scala/com/twitter/finagle/loadbalancer/Sticky.scala
@@ -1,0 +1,95 @@
+package com.twitter.finagle.loadbalancer
+
+import com.twitter.finagle.{ClientConnection, NoBrokersAvailableException, Service,
+  ServiceFactory, ServiceFactoryProxy, Status}
+import com.twitter.finagle.service.FailingFactory
+import com.twitter.finagle.stats.StatsReceiver
+import com.twitter.util.{Activity, Future, Promise, Time}
+
+/**
+ * A simple sticky load balancer which selects the first backend it
+ * can connect to and sticks to it unless that backend is not functioning
+ * properly, in which case it will change to another functioning backend.
+ */
+class StickyBalancer[Req, Rep](
+    val activity: Activity[Traversable[ServiceFactory[Req, Rep]]],
+    val statsReceiver: StatsReceiver,
+    val emptyException: NoBrokersAvailableException,
+    val maxEffort: Int = 5)
+  extends ServiceFactory[Req, Rep]
+  with Balancer[Req, Rep]
+  with Updating[Req, Rep] {
+
+  // For the OnReady mixin
+  private[this] val ready = new Promise[Unit]
+  override def onReady: Future[Unit] = ready
+
+  protected[this] val maxEffortExhausted = statsReceiver.counter("max_effort_exhausted")
+
+  override def apply(conn: ClientConnection): Future[Service[Req,Rep]] = {
+    dist.pick()(conn)
+  }
+
+  protected class Node(val factory: ServiceFactory[Req, Rep])
+      extends ServiceFactoryProxy[Req,Rep](factory)
+      with NodeT[Req,Rep] { self =>
+    type This = Node
+    // Note: These stats are never updated.
+    def load = 0.0
+    def pending = 0
+    def token = 0
+
+    override def close(deadline: Time): Future[Unit] = factory.close(deadline)
+    override def apply(conn: ClientConnection): Future[Service[Req,Rep]] = factory(conn)
+  }
+
+  /**
+    * A simple sticky distributor.
+    */
+  protected class Distributor(val vector: Vector[Node])
+      extends DistributorT[Node] {
+    type This = Distributor
+
+    private[this] var currentNode = 0
+
+    private[this] val nodeUp: Node => Boolean = {_.isAvailable}
+
+    private[this] val nodes = vector
+
+    /**
+      * Will only return Nodes that have Status.Open
+      */
+    def pick(): Node = {
+      if (nodes.isEmpty) failingNode(emptyException)
+      else {
+        val node = nodes(currentNode)
+        node.status match {
+          case Status.Open => node
+          case _ => synchronized {
+            var idx = 0
+            var res = -1
+
+            while (idx < nodes.size && res == -1) {
+              if (nodes(idx).isAvailable) res = idx
+              else idx += 1
+            }
+
+            if (res != -1) {
+              currentNode = res
+              nodes(res)
+            } else failingNode(emptyException)
+          }
+        }
+      }
+    }
+
+    def needsRebuild: Boolean = false
+
+    def rebuild(): This = new Distributor(vector)
+    def rebuild(vector: Vector[Node]): This = new Distributor(vector)
+  }
+
+  protected def initDistributor(): Distributor = new Distributor(Vector.empty)
+  protected def newNode(factory: ServiceFactory[Req,Rep], statsReceiver: StatsReceiver): Node = new Node(factory)
+  protected def failingNode(cause: Throwable): Node = new Node(new FailingFactory(cause))
+}
--- a/finagle-core/src/test/scala/com/twitter/finagle/loadbalancer/StickyBalancerTest.scala
+++ b/finagle-core/src/test/scala/com/twitter/finagle/loadbalancer/StickyBalancerTest.scala
@@ -1,0 +1,150 @@
+package com.twitter.finagle.loadbalancer
+
+import com.twitter.app.App
+import com.twitter.conversions.time._
+import com.twitter.finagle._
+import com.twitter.finagle.stats.{StatsReceiver, NullStatsReceiver, InMemoryStatsReceiver}
+import com.twitter.util.{Function => _, _}
+import org.junit.runner.RunWith
+import org.scalatest.FunSuite
+import org.scalatest.junit.JUnitRunner
+import scala.collection.mutable
+
+private[loadbalancer] trait StickySuite {
+  // number of servers
+  val N: Int = 100
+  // number of reqs
+  val R: Int = 100000
+
+  trait SServiceFactory extends ServiceFactory[Unit, Int]
+
+  val noBrokers = new NoBrokersAvailableException
+
+  def newBal(
+    fs: Var[Traversable[SServiceFactory]],
+    sr: StatsReceiver = NullStatsReceiver
+  ): ServiceFactory[Unit, Int] = new StickyBalancer(
+    Activity(fs.map(Activity.Ok(_))),
+    statsReceiver = sr,
+    emptyException = noBrokers,
+    maxEffort = 1
+  )
+}
+
+@RunWith(classOf[JUnitRunner])
+class StickyBalancerTest extends FunSuite with StickySuite {
+  case class LoadedFactory(id: Int) extends SServiceFactory {
+    @volatile var stat: Status = Status.Open
+    var load = 0
+    var sum = 0
+    var count = 0
+
+    def apply(conn: ClientConnection) = {
+      load += 1
+      sum += load
+      count += 1
+
+      Future.value(new Service[Unit, Int] {
+        def apply(req: Unit) = Future.value(id)
+        override def close(deadline: Time) = {
+          load -= 1
+          sum += load
+          count += 1
+          Future.Done
+        }
+      })
+    }
+
+    def close(deadline: Time) = Future.Done
+    override def toString = "LoadedFactory(%d, %d, %s)".format(id, load, stat)
+    override def status = stat
+  }
+
+  test("Sticks to one backend") {
+    val init = Vector.tabulate(N) { i => new LoadedFactory(i) }
+    val bal = newBal(Var(init))
+    for (_ <- 0 until R) bal()
+
+    assert(init.filter(_.count == R).size == 1)
+    assert(init.filter(_.count == 0).size == N - 1)
+  }
+
+  test("Empty load balancer throws") {
+    val vec = Var(Vector.empty[LoadedFactory])
+    val bal = newBal(vec)
+    val exc = intercept[NoBrokersAvailableException] { Await.result(bal()) }
+    assert(exc eq noBrokers)
+
+    vec() :+= new LoadedFactory(0)
+    for (_ <- 0 until R) Await.result(bal())
+    assert(vec().head.load == R)
+
+    vec() = Vector.empty
+    intercept[NoBrokersAvailableException] { Await.result(bal()) }
+  }
+
+  test("Closing a node removes it from load balancing") {
+    val init = Vector.tabulate(N) { i => new LoadedFactory(i) }
+    val bal = newBal(Var.value(init))
+
+    for (_ <- 0 until R) {
+      assert(Await.result(bal()).status == Status.Open)
+    }
+    assert(init(0).count > 0)
+
+    assert(init(0).status == Status.Open)
+    init(0).stat = Status.Closed
+    assert(init(0).status == Status.Closed)
+
+    val init0count = init(0).count
+
+    // There are no closed nodes returned from bal()
+    for (_ <- 0 until R) {
+      assert(Await.result(bal()).status == Status.Open)
+    }
+
+    // no new traffic has been sent to init(0) which is closed.
+    assert(init0count == init(0).count)
+  }
+
+  test("Changing a node from Closed to Open re-adds it to load balancing") {
+    val init = Vector.tabulate(2) { i => new LoadedFactory(i) }
+    val bal = newBal(Var(init))
+
+    // checkpoint for seeing how much traffic init(0) has seen
+    var init0count = init(0).count
+
+    assert(init(0).status == Status.Open)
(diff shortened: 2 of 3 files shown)


johanstenberg92:
References issue #496.

mosesn:
I have a couple questions about your commit message.  What exactly do you mean by "strong eventual consistency".  Are you referring to read-your-writes consistency?  Delayed-t consistency?

The nodes *are* still identical replicas, right?  The only difference is that you want to be pegged to a single remote peer if possible, which makes it easier for a server to maintain state for a given client.

finagle-core/src/main/scala/com/twitter/finagle/loadbalancer/Sticky.scala
```diff
@@ -0,0 +1,95 @@
...
+import com.twitter.finagle.{ClientConnection, NoBrokersAvailableException, Service,
+  ServiceFactory, ServiceFactoryProxy, Status}
+import com.twitter.finagle.service.FailingFactory
+import com.twitter.finagle.stats.StatsReceiver
+import com.twitter.util.{Activity, Future, Promise, Time}
+
+/**
+ * A simple sticky load balancer which selects the first backend it
+ * can connect to and sticks to it unless that backend is not functioning
+ * properly, in which case it will change to another functioning backend.
```
mosesn: please motivate here

finagle-core/src/main/scala/com/twitter/finagle/loadbalancer/Sticky.scala
```diff
@@ -0,0 +1,95 @@
...
+ */
+class StickyBalancer[Req, Rep](
+    val activity: Activity[Traversable[ServiceFactory[Req, Rep]]],
```
mosesn: do these all need to be public fields?

finagle-core/src/main/scala/com/twitter/finagle/loadbalancer/Sticky.scala
```diff
@@ -0,0 +1,95 @@
...
+    val statsReceiver: StatsReceiver,
+    val emptyException: NoBrokersAvailableException,
+    val maxEffort: Int = 5)
+  extends ServiceFactory[Req, Rep]
+  with Balancer[Req, Rep]
+  with Updating[Req, Rep] {
+
+  // For the OnReady mixin
+  private[this] val ready = new Promise[Unit]
```
mosesn: you never satisfy this promise

finagle-core/src/main/scala/com/twitter/finagle/loadbalancer/Sticky.scala
```diff
@@ -0,0 +1,95 @@
...
+  override def apply(conn: ClientConnection): Future[Service[Req,Rep]] = {
+    dist.pick()(conn)
+  }
+
+  protected class Node(val factory: ServiceFactory[Req, Rep])
+      extends ServiceFactoryProxy[Req,Rep](factory)
+      with NodeT[Req,Rep] { self =>
+    type This = Node
+    // Note: These stats are never updated.
+    def load = 0.0
```
mosesn: make these vals instead of defs if they're never updated

finagle-core/src/main/scala/com/twitter/finagle/loadbalancer/Sticky.scala
```diff
@@ -0,0 +1,95 @@
...
+  /**
+    * A simple sticky distributor.
+    */
+  protected class Distributor(val vector: Vector[Node])
+      extends DistributorT[Node] {
+    type This = Distributor
+
+    private[this] var currentNode = 0
+
+    private[this] val nodeUp: Node => Boolean = {_.isAvailable}
```
mosesn: function literals should have spaces at the beginning and the end, like so: `{ node => node.isAvailable }`.  I'm also personally allergic to using placeholder syntax with curly braces, but feel free to ignore that nit.

finagle-core/src/main/scala/com/twitter/finagle/loadbalancer/Sticky.scala
```diff
@@ -0,0 +1,95 @@
...
+
+    private[this] val nodes = vector
+
+    /**
+      * Will only return Nodes that have Status.Open
+      */
+    def pick(): Node = {
+      if (nodes.isEmpty) failingNode(emptyException)
+      else {
+        val node = nodes(currentNode)
```
mosesn: because this isn't synchronized, we may not see updates to currentNode.  you should either mark it volatile or synchronize the read.

finagle-core/src/test/scala/com/twitter/finagle/loadbalancer/StickyBalancerTest.scala
```diff
@@ -0,0 +1,150 @@
...
+  ): ServiceFactory[Unit, Int] = new StickyBalancer(
+    Activity(fs.map(Activity.Ok(_))),
+    statsReceiver = sr,
+    emptyException = noBrokers,
+    maxEffort = 1
+  )
+}
+
+@RunWith(classOf[JUnitRunner])
+class StickyBalancerTest extends FunSuite with StickySuite {
```
mosesn: please check that load balancer `status` is the best of its constituent nodes

finagle-core/src/main/scala/com/twitter/finagle/loadbalancer/Sticky.scala
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  /**
+    * A simple sticky distributor.
+    */
+  protected class Distributor(val vector: Vector[Node])
+      extends DistributorT[Node] {
+    type This = Distributor
+
+    private[this] var currentNode = 0
```
mosesn: I don't think the TrafficDistributor reorders nodes, so everyone will start on the same node if their host list or service discovery gives them the nodes in the same order.

finagle-core/src/main/scala/com/twitter/finagle/loadbalancer/Sticky.scala
```diff
@@ -0,0 +1,95 @@
...
+      else {
+        val node = nodes(currentNode)
+        node.status match {
+          case Status.Open => node
+          case _ => synchronized {
+            var idx = 0
+            var res = -1
+
+            while (idx < nodes.size && res == -1) {
+              if (nodes(idx).isAvailable) res = idx
```
mosesn: check status instead, imo.  it's more clear–`isAvailable` was a mistake.

finagle-core/src/main/scala/com/twitter/finagle/loadbalancer/Sticky.scala
```diff
@@ -0,0 +1,95 @@
...
+    /**
+      * Will only return Nodes that have Status.Open
+      */
+    def pick(): Node = {
+      if (nodes.isEmpty) failingNode(emptyException)
+      else {
+        val node = nodes(currentNode)
+        node.status match {
+          case Status.Open => node
+          case _ => synchronized {
```
mosesn: since this is written in an imperative style anyway, let's go whole hog.  let's use multiple exits, which will simplify the code.

```scala
var idx = 0

while (idx < nodes.size) {
  if (nodes(idx).isAvailable) {
    currentNode = idx
    return nodes(idx)
  }
  idx += 1
}
failingNode(emptyException)
```

mosesn: This synchronized block is either at the wrong point, or we should check whether status is open again.  Even if someone else beats us in the race here, we're still going to do all of this work as soon as they they let go of the lock.

Some alternatives:
1. double check the condition.
2. expand the synchronized block to include the original status check (I agree with your assessment that it's not the best tbh)
3.  use a regular java Lock instance so you can tryLock and if you fail to acquire the lock, busywait by recursing.

finagle-core/src/main/scala/com/twitter/finagle/loadbalancer/Sticky.scala
```diff
@@ -0,0 +1,95 @@
...
+            var idx = 0
+            var res = -1
+
+            while (idx < nodes.size && res == -1) {
+              if (nodes(idx).isAvailable) res = idx
+              else idx += 1
```
mosesn: it's not clear to me that incrementing the index is the right approach here, because I don't think the traffic distributor reorders nodes.  if clients are evenly distributed across remote peers but one peer goes down, suddenly the subsequent peer will get double the traffic.  for example:

node 2 still alive:
[(node 1, clients 5), (node 2, clients 5), (node 3, clients 5)]

node 2 dies:
[(node 1, clients 5), (node 3, clients 10)]

which seems like the wrong behavior.  let's either randomize on node down or ensure that the traffic distributor will reorder for us.

finagle-core/src/main/scala/com/twitter/finagle/loadbalancer/Sticky.scala
```diff
@@ -0,0 +1,95 @@
...
+    val maxEffort: Int = 5)
+  extends ServiceFactory[Req, Rep]
+  with Balancer[Req, Rep]
+  with Updating[Req, Rep] {
+
+  // For the OnReady mixin
+  private[this] val ready = new Promise[Unit]
+  override def onReady: Future[Unit] = ready
+
+  protected[this] val maxEffortExhausted = statsReceiver.counter("max_effort_exhausted")
```
mosesn: this counter is never used

finagle-core/src/main/scala/com/twitter/finagle/loadbalancer/Sticky.scala
mosesn: let's also add a note that if many nodes go down, and then later are revived, but the old nodes never go down, there's no mechanism to later move load back to the revived nodes.

finagle-core/src/test/scala/com/twitter/finagle/loadbalancer/StickyBalancerTest.scala
mosesn: please add a test to check that it behaves correctly if traffic distributor hands you the same nodes as everyone else in the same order.

finagle-core/src/main/scala/com/twitter/finagle/loadbalancer/Sticky.scala
```diff
@@ -0,0 +1,95 @@
...
+/**
+ * A simple sticky load balancer which selects the first backend it
+ * can connect to and sticks to it unless that backend is not functioning
+ * properly, in which case it will change to another functioning backend.
+ */
+class StickyBalancer[Req, Rep](
+    val activity: Activity[Traversable[ServiceFactory[Req, Rep]]],
+    val statsReceiver: StatsReceiver,
+    val emptyException: NoBrokersAvailableException,
+    val maxEffort: Int = 5)
```
mosesn: why let maxEffort be configurable?  don't we check all of the nodes anyway?

finagle-core/src/main/scala/com/twitter/finagle/loadbalancer/Sticky.scala
```diff
@@ -0,0 +1,95 @@
...
+      if (nodes.isEmpty) failingNode(emptyException)
+      else {
+        val node = nodes(currentNode)
+        node.status match {
+          case Status.Open => node
+          case _ => synchronized {
+            var idx = 0
+            var res = -1
+
+            while (idx < nodes.size && res == -1) {
```
mosesn: the way this is phrased means that it won't ever go back and recheck nodes that have been marked busy or dead.

finagle-core/src/test/scala/com/twitter/finagle/loadbalancer/StickyBalancerTest.scala
mosesn: please add a test that check that adding a new node to the cluster doesn't change which node is stuck to.

finagle-core/src/main/scala/com/twitter/finagle/loadbalancer/Sticky.scala
```diff
@@ -0,0 +1,95 @@
...
+              nodes(res)
+            } else failingNode(emptyException)
+          }
+        }
+      }
+    }
+
+    def needsRebuild: Boolean = false
+
+    def rebuild(): This = new Distributor(vector)
```
mosesn: this doesn't preserve `currentNode` so we might end up choosing a completely different `currentNode`.

finagle-core/src/main/scala/com/twitter/finagle/loadbalancer/Sticky.scala
```diff
@@ -0,0 +1,95 @@
...
+
+            while (idx < nodes.size && res == -1) {
+              if (nodes(idx).isAvailable) res = idx
+              else idx += 1
+            }
+
+            if (res != -1) {
+              currentNode = res
+              nodes(res)
+            } else failingNode(emptyException)
```
mosesn: this isn't actually empty, right?  it's that everyone is dead.  instead of failing we typically just pick one in that case, because we're optimistic =)

finagle-core/src/main/scala/com/twitter/finagle/loadbalancer/Sticky.scala
mosesn: also this val seems unused, as far as I can tell.

finagle-core/src/main/scala/com/twitter/finagle/loadbalancer/Sticky.scala
```diff
@@ -0,0 +1,95 @@
...
+    def load = 0.0
+    def pending = 0
+    def token = 0
+
+    override def close(deadline: Time): Future[Unit] = factory.close(deadline)
+    override def apply(conn: ClientConnection): Future[Service[Req,Rep]] = factory(conn)
+  }
+
+  /**
+    * A simple sticky distributor.
```
mosesn: indentation

finagle-core/src/main/scala/com/twitter/finagle/loadbalancer/Sticky.scala
```diff
@@ -0,0 +1,95 @@
...
+    type This = Distributor
+
+    private[this] var currentNode = 0
+
+    private[this] val nodeUp: Node => Boolean = {_.isAvailable}
+
+    private[this] val nodes = vector
+
+    /**
+      * Will only return Nodes that have Status.Open
```
mosesn: indentation

mosesn: add a note that this is where we're sticky too.

finagle-core/src/test/scala/com/twitter/finagle/loadbalancer/StickyBalancerTest.scala
mosesn: please add a test that checks that if a node is marked dead, it can later be marked alive again.

mosesn:
This is a good first draft, thanks for the PR!  I'll see if I can get a potential user internally to take a look too.

mosesn:
cc @olix0r who cares about this too.

johanstenberg92:
Thanks @mosesn for the feedback! However @olix0r your comment you stated in the issue (#496) made me think. 

I guess the purpose of using a load balancer is to balance the load of the provided hosts, not to provide some layer of resilience management (even though it naturally does, it's first a load balancer and second a resilience-manager). To evaluate on "strong eventual consistency", let's just scratch the word strong and use eventual consistency. 

With the original Finagle client, I can write a value to my in-memory, eventually consistent database and then when I read it instantly after, the Finagle client has rotated to another host and I can't find my value! If you want to support eventually consistent systems and the use case I just described - should it be done with load balancers such as this PR or is it up to the user of Finagle to write a thin wrapper around the client?

mosesn:
Eventual consistency is still not well defined.  Unless it's OK that remote servers sometimes don't have the data you just wrote, it sounds like what you want is read-your-writes consistency.  If that's the case, your data server needs to ensure that it writes to a replicated log, and tells clients the position of the write in the replicated log, so that it can ensure that when it reconnect to another data node, that the other node is caught up to that position in the replicated log.

When we get to that layer, this load balancer is not going to be the right tool.  You will have to do what you're describing where you use the ServiceFactory directly, so that you can have special behavior when you switch nodes.

Does that sound right?

johanstenberg92:
Now we are talking about consistency from two different perspectives, you're talking about the client's perspective. Let's focus on the PR - @mosesn do you think this PR is bloaty and it's provided logic should be handled by the Finagle user through e.g. ServiceFactory? Or should I fix it according to your feedback? If we forget about eventual consistency for a moment, I would argue this LB can be useful in testing etc.

mosesn:
Yeah, from talking to other users who want sticky sessions, it looks like their use cases are also satisfied by using the ServiceFactory directly.  We probably don't want to add this until we're sure we have a use case where we need the load balancer to own this.

johanstenberg92:
@mosesn feel free to close this PR and the corresponding issue then! Cheers

PR #502, opened 2016-05-02
Netty 4 server fixes

lukiano:
- Fix casting bug at shutdown. 

- Call retain() on ByteBuf before offering to an AsyncQueue

### DIFF
--- a/finagle-netty4/src/main/scala/com/twitter/finagle/netty4/Netty4Listener.scala
+++ b/finagle-netty4/src/main/scala/com/twitter/finagle/netty4/Netty4Listener.scala
@@ -13,7 +13,7 @@ import io.netty.channel._
 import io.netty.channel.nio.NioEventLoopGroup
 import io.netty.channel.socket.SocketChannel
 import io.netty.channel.socket.nio.NioServerSocketChannel
-import io.netty.util.concurrent.GenericFutureListener
+import io.netty.util.concurrent.{Future => NettyFuture, FutureListener}
 import java.lang.{Boolean => JBool, Integer => JInt}
 import java.net.SocketAddress
 import java.util.concurrent.TimeUnit
@@ -130,8 +130,8 @@ private[finagle] case class Netty4Listener[In, Out](
         // Existing tasks have ``deadline`` time to finish executing.
         bossLoop
           .shutdownGracefully(0 /* quietPeriod */ , deadline.inMillis /* timeout */ , TimeUnit.MILLISECONDS)
-          .addListener(new GenericFutureListener[Nothing] {
-            def operationComplete(future: Nothing): Unit = p.setDone()
+          .addListener(new FutureListener[Any] {
+            def operationComplete(future: NettyFuture[Any]) = p.setDone()
           })
         p
       }
--- a/finagle-netty4/src/main/scala/com/twitter/finagle/netty4/transport/ChannelTransport.scala
+++ b/finagle-netty4/src/main/scala/com/twitter/finagle/netty4/transport/ChannelTransport.scala
@@ -130,7 +130,7 @@ private[netty4] class ChannelTransport[In, Out](ch: Channel) extends Transport[I
 
   override def toString = s"Transport<channel=$ch, onClose=$closed>"
 
-  ch.pipeline().addLast("finagleChannelTransport", new SimpleChannelInboundHandler[Out]() {
+  ch.pipeline().addLast("finagleChannelTransport", new SimpleChannelInboundHandler[Out](false) {
 
     override def channelReadComplete(ctx: ChannelHandlerContext): Unit = {
       if (!ch.config.isAutoRead && msgsNeeded.get > 0) ch.read()
--- a/finagle-netty4/src/test/scala/com/twitter/finagle/netty4/Netty4ListenerTest.scala
+++ b/finagle-netty4/src/test/scala/com/twitter/finagle/netty4/Netty4ListenerTest.scala
@@ -152,7 +152,7 @@ class Netty4ListenerTest extends FunSuite with Eventually with IntegrationPatien
     eventually { counterEquals("connects")(3) }
 
     // listening socket is closed
-    Await.ready(server.close(), Duration.fromSeconds(15))
+    Await.ready(server.close(Time.Top), Duration.fromSeconds(15))
 
     // new connection attempts fail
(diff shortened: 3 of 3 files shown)


finagle-netty4/src/main/scala/com/twitter/finagle/netty4/Netty4Listener.scala
```diff
@@ -128,11 +128,16 @@ private[finagle] case class Netty4Listener[In, Out](
 
         // The boss loop immediately starts refusing new work.
         // Existing tasks have ``deadline`` time to finish executing.
-        bossLoop
+        val shutdown: NettyFuture[_] = bossLoop
           .shutdownGracefully(0 /* quietPeriod */ , deadline.inMillis /* timeout */ , TimeUnit.MILLISECONDS)
-          .addListener(new GenericFutureListener[Nothing] {
-            def operationComplete(future: Nothing): Unit = p.setDone()
-          })
+        def shutdownListener[N <: NettyFuture[_]]: GenericFutureListener[N] =
```
vkostyukov: ffti: `NettyFuture[Any]` and `GenericFutureListener[Any]` should work as well.

I have an internal patch that also fixes that but it's been soaking for review for a long time so I'm totally fine to merge this in.

finagle-netty4/src/main/scala/com/twitter/finagle/netty4/channel/Netty4ServerChannelInitializer.scala
```diff
@@ -103,13 +104,25 @@ private[netty4] class Netty4ServerChannelInitializer(
...
+    pipeline.addLast("ContentRetainer", ContentRetainer)
+
     // The bridge handler must be last in the pipeline to ensure
     // that the bridging code sees all encoding and transformations
     // of inbound messages.
     pipeline.addLast("finagle bridge", newBridge())
   }
 }
 
+private[netty4] object ContentRetainer extends ChannelInboundHandlerAdapter {
```
vkostyukov: Ah that's interesting! I guess you're right and we should totally ratain those objects.

What do you think about fixing `ChannelTransport` to do `ReferenceCountUtil.ratain(msg)` before offering the message into `AsyncQueue`, instead of introducing a new handler here?

dschobel:
thanks @lukiano, can you add a unit test for the casting bug and updating the git commit to follow https://github.com/twitter/finagle/blob/develop/CONTRIBUTING.md ?

Taking a step back from the question of the reference leak this patches for pipelines that let undecoded buffers through, I don't think letting netty bytebufs leak into the finagle layer of transports and dispatchers is something we should encourage so I'd rather install a terminal bytebuf 2 `com.twitter.util.Buf` handler and then ask protocol implementors who want to use bytebufs and the netty convenience classes to build their protocols in a channel handler. 

That lets us sequester the netty types to the pipeline and nudges people towards using buf and gaining the benefit of not being tied to a specific version of netty.

finagle-netty4/src/main/scala/com/twitter/finagle/netty4/channel/Netty4ServerChannelInitializer.scala
lukiano: I thought of that as well, but wasn't sure about maybe mixing concerns

lukiano:
@dschobel Will do!
About your second paragraph, this isn't just ByteBuf, but actual Netty messages (HttpContent and inheritance) which also extends ByteBufHolder, which itself extends ReferenceCounted, so we can call retain()
Related, eventually this will need to be improved to support pooled ByteBufs

mosesn:
@lukiano yah, we were thinking about how to implement pooled ByteBufs, we were thinking we could use PhantomReferences?  What do you think?

lukiano:
@mosesn Not sure yet as I'm not entirely familiar with ByteBufs, but was thinking something along the lines of when the Buf readerIndex reaches the end, then release the object.

mosesn:
@lukiano but what if we want to read it again, or we don't want to finish reading it?

dschobel:
> About your second paragraph, this isn't just ByteBuf, but actual Netty messages (HttpContent and inheritance) which also extends ByteBufHolder, which itself extends ReferenceCounted, so we can call retain()
> Related, eventually this will need to be improved to support pooled ByteBufs

@lukiano thinking about it a bit more, since this never releases buffers it's not usable with the pooled allocator unless the protocol implementation copies every buffer, in which case, why use the pooled allocator?

It seems like it would be safer and less confusing to let netty's leak detectors blow up early if someone tries to use the pooled allocator rather than force that constraint on protocol implementors.

As @mosesn says,  there are options for making pooled work without copying but that will be a bigger effort.

lukiano:
I think the user should be able to choose between non-pooled or pooled buffers. Also they should be able to choose if they are heap or direct based. I think Netty defaults to direct, but in case of Unpooled buffers, maybe heap ones are better as a default.
If the user chooses to go in the direct/pooled route, it must be aware of the compromises. These are examples:

- If there's only one buffer, like a short message, or a chunked message put together via HttpObjectAggregator, then it will be released when the service responds. That is, Future[Response] is completed.
- If there are chunk of buffers, provided by the `com.twitter.io.Reader` of the request, each time `read(n: Int)` is called to get a new chunk, the previous one is released.

I'm sure there are corner cases. This PR is to make it work with unpooled buffers

dschobel:
@lukiano okay I was missing that unpooled buffers can be freed and disabling the leak detection code doesn't stop it from being a runtime exception to access its contents.

I wrote two tests for the listener and the transporter with no protocol implementation and the listener's buffers hit finagle space with `refCnt == 1` which seems correct.

There is a bug on the transporter side because the `ChannelTransport`'s SimpleChannelHandler instance auto-releases its buffers and will cause a runtime failure if the protocol implementation doesn't explicitly retain them.

lukiano:
@dschobel I have updated this PR with the SimpleChannelHandler change, updated the commit messages, and updated the tests to reflect the casting issue.

mosesn:
LGTM

dschobel:
thanks @lukiano, I'm merging it internally. it will sync out to the develop branch on Monday/Tuesday.

dschobel:
merged in 2f1786b4658c60a0f22dafc83b0cdfcccba5619d. thanks @lukiano

PR #503, opened 2016-05-03
Lazily instantiate StackServer's JvmFilter

olix0r:
Problem

StackServer statically instantiates a `MkJvmFilter` with a `c.t.jvm.Jvm`. This creates a timer thread that samples JVM stats. We'd like to disable the JvmFilter stack module during profiling, but JVM stats polling is initiated regardless of whether the module is used.

Solution

Lazily instantiate StackServer's JvmFilter.

Result

The stats polling timer is only initiated when the JvmFilter server module is
enabled.

Also-by: Steve Jenson <stevej@buoyant.io>

### DIFF
--- a/finagle-core/src/main/scala/com/twitter/finagle/server/StackServer.scala
+++ b/finagle-core/src/main/scala/com/twitter/finagle/server/StackServer.scala
@@ -20,7 +20,7 @@ import scala.collection.JavaConverters._
 
 object StackServer {
 
-  private[this] val newJvmFilter = new MkJvmFilter(Jvm())
+  private[this] lazy val newJvmFilter = new MkJvmFilter(Jvm())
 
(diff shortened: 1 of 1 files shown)


kevinoliver:
I understand the need for the change in `StackServer`. I'm not sure I see why the change is necessary in `MkJvmFilter`.

olix0r:
it seemed surprising that polling is started by the factory regardless of whether it builds anything. i can back it out, though.

kevinoliver:
yeah, i'm not a fan of side-effects from constructors either. wanted to make sure i understood the change. i'm fine leaving it as is.

kevinoliver:
🚢

mosesn:
LGTM

mosesn:
This was merged in a while ago =) https://github.com/twitter/finagle/commit/dd5736a31f44ce145aed6327586504e75fda753c

PR #504, opened 2016-05-03
Requeue retryable failures

olix0r:
Problem

ResponseClassifiers provide an explicit mechanism for marking failures as retryable, yet this response classification is not taken into account when determining whether a request should be requeued.

Solution

Provide RequeueFilter with a ResponseClassifier.  RequeueFilter considers a request to be requeued when its response is classified as a RetryableFailure.

### DIFF
--- a/finagle-core/src/main/scala/com/twitter/finagle/service/RequeueFilter.scala
+++ b/finagle-core/src/main/scala/com/twitter/finagle/service/RequeueFilter.scala
@@ -36,13 +36,14 @@ import com.twitter.util._
  *      for more details.
  */
 private[finagle] class RequeueFilter[Req, Rep](
-    retryBudget: RetryBudget,
-    retryBackoffs: Stream[Duration],
-    statsReceiver: StatsReceiver,
-    canRetry: () => Boolean,
-    maxRetriesPerReq: Double,
-    timer: Timer)
-  extends SimpleFilter[Req, Rep] {
+  retryBudget: RetryBudget,
+  retryBackoffs: Stream[Duration],
+  statsReceiver: StatsReceiver,
+  canRetry: () => Boolean,
+  maxRetriesPerReq: Double,
+  timer: Timer,
+  classifier: ResponseClassifier = PartialFunction.empty
+) extends SimpleFilter[Req, Rep] {
 
   require(maxRetriesPerReq >= 0,
     s"maxRetriesPerReq must be non-negative: $maxRetriesPerReq")
@@ -68,37 +69,40 @@ private[finagle] class RequeueFilter[Req, Rep](
     retriesRemaining: Int,
     backoffs: Stream[Duration]
   ): Future[Rep] = {
-    service(req).transform {
-      case t@Throw(RetryPolicy.RetryableWriteException(_)) =>
-        if (!canRetry()) {
-          canNotRetryCounter.incr()
-          responseFuture(attempt, t)
-        } else if (retriesRemaining > 0 && retryBudget.tryWithdraw()) {
-          backoffs match {
-            case Duration.Zero #:: rest =>
-              // no delay between retries. Retry immediately.
-              requeueCounter.incr()
-              applyService(req, service, attempt + 1, retriesRemaining - 1, rest)
-            case delay #:: rest =>
-              // Delay and then retry.
-              timer.doLater(delay) {
+    service(req).transform { result =>
+      classifier.applyOrElse(ReqRep(req, result), ResponseClassifier.Default) match {
+        case ResponseClass.RetryableFailure =>
+          if (!canRetry()) {
+            canNotRetryCounter.incr()
+            responseFuture(attempt, result)
+          } else if (retriesRemaining > 0 && retryBudget.tryWithdraw()) {
+            backoffs match {
+              case Duration.Zero #:: rest =>
+                // no delay between retries. Retry immediately.
                 requeueCounter.incr()
                 applyService(req, service, attempt + 1, retriesRemaining - 1, rest)
-              }.flatten
-            case _ =>
-              // Schedule has run out of entries. Budget is empty.
+              case delay #:: rest =>
+                // Delay and then retry.
+                timer.doLater(delay) {
+                  requeueCounter.incr()
+                  applyService(req, service, attempt + 1, retriesRemaining - 1, rest)
+                }.flatten
+              case _ =>
+                // Schedule has run out of entries. Budget is empty.
+                budgetExhaustCounter.incr()
+                responseFuture(attempt, result)
+            }
+          } else {
+            if (retriesRemaining > 0)
               budgetExhaustCounter.incr()
-              responseFuture(attempt, t)
+            else
+              requestLimitCounter.incr()
+            responseFuture(attempt, result)
           }
-        } else {
-          if (retriesRemaining > 0)
-            budgetExhaustCounter.incr()
-          else
-            requestLimitCounter.incr()
-          responseFuture(attempt, t)
-        }
-      case t =>
-        responseFuture(attempt, t)
+
+        case _ =>
+          responseFuture(attempt, result)
+      }
     }
   }
 
--- a/CHANGES
+++ b/CHANGES
@@ -7,6 +7,12 @@ Note that ``RB_ID=#`` correspond to associated messages in commits.
 6.x
 -----
 
+Runtime Behavior Changes
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+  * finagle-core: RequeueFilter now uses the stack's ResponseClassifier to determine whether
+    a response should be considered retryable.
+
 6.35.0
 ------
 
--- a/finagle-core/src/main/scala/com/twitter/finagle/service/Retries.scala
+++ b/finagle-core/src/main/scala/com/twitter/finagle/service/Retries.scala
@@ -93,7 +93,13 @@ object Retries {
    * (see [[RetryPolicy.RetryableWriteException]]).
    */
   private[finagle] def moduleRequeueable[Req, Rep]: Stackable[ServiceFactory[Req, Rep]] =
-    new Stack.Module3[Stats, Budget, HighResTimer, ServiceFactory[Req, Rep]] {
+    new Stack.Module4[
+      Stats,
+      Budget,
+      param.ResponseClassifier,
+      HighResTimer,
+      ServiceFactory[Req, Rep]
+    ] {
       def role: Stack.Role = Retries.Role
 
       def description: String =
@@ -102,6 +108,7 @@ object Retries {
       def make(
         statsP: param.Stats,
         budgetP: Budget,
+        classifierP: param.ResponseClassifier,
         timerP: HighResTimer,
         next: ServiceFactory[Req, Rep]
       ): ServiceFactory[Req, Rep] = {
@@ -109,12 +116,19 @@ object Retries {
         val scoped = statsRecv.scope("retries")
         val requeues = scoped.counter("requeues")
         val retryBudget = budgetP.retryBudget
+        val param.ResponseClassifier(classifier) = classifierP
         val timer = timerP.timer
 
-        val filters = newRequeueFilter(
-          retryBudget, budgetP.requeueBackoffs, withdrawsOnly = false, scoped, timer, next
+        val filter = newRequeueFilter(
+          retryBudget,
+          budgetP.requeueBackoffs,
+          withdrawsOnly = false,
+          scoped,
+          timer,
+          classifier,
+          next
         )
-        svcFactory(retryBudget, filters, scoped, requeues, next)
+        svcFactory(retryBudget, filter, scoped, requeues, next)
       }
     }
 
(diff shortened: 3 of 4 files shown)


olix0r:
It occurs to me that, for the purposes of linkerd, we probably want to move the requeue filter up the stack:  Right now requeueing occurs on the same client.  Given a situation like `/host/foo => 1 * /srv/foo & 9 * /srv/bar` where /srv/foo fails some portion of requests, we should be able to retry onto /srv/bar.  Furthermore, in an ideal world, we'd perform load balancing, failure accrual, etc in the same fashion...

olix0r:
nudge.  I copied this functionality into linkerd to start playing with it, and it seems like something we definitely want.  I got the impression from Gitter that there might be some reticence to a change like this.  From my point of view: this will be a no-op change to anyone who isn't setting a response classifier.  For those setting response classifiers _that return RetryableFailure_, I assume this is actually the behavior they would want by returning such a response classification?

kevinoliver:
@olix0r i personally haven't had time to look. hoping to by the end of the week.

olix0r:
@kevinoliver no worries, thanks for the update

finagle-core/src/main/scala/com/twitter/finagle/service/RequeueFilter.scala
```diff
@@ -36,13 +36,14 @@ import com.twitter.util._
...
  */
 private[finagle] class RequeueFilter[Req, Rep](
-    retryBudget: RetryBudget,
-    retryBackoffs: Stream[Duration],
-    statsReceiver: StatsReceiver,
-    canRetry: () => Boolean,
-    maxRetriesPerReq: Double,
-    timer: Timer)
-  extends SimpleFilter[Req, Rep] {
+  retryBudget: RetryBudget,
```
kevinoliver: please don't reformat — we follow the standard scala style guide: http://docs.scala-lang.org/style/declarations

finagle-core/src/main/scala/com/twitter/finagle/service/RequeueFilter.scala
```diff
@@ -36,13 +36,14 @@ import com.twitter.util._
...
+  retryBackoffs: Stream[Duration],
+  statsReceiver: StatsReceiver,
+  canRetry: () => Boolean,
+  maxRetriesPerReq: Double,
+  timer: Timer,
+  classifier: ResponseClassifier = PartialFunction.empty
```
kevinoliver: lets not give a default argument since we do not need to keep backwards compat for package private changes.

kevinoliver:
`RequeueFilter` is intended for nacks and these retries are usually going to not be that. 

i think this type of code belongs in `RetryFilter` instead.

kevinoliver:
or barring that, we could consider adding a new filter that has this behavior and sits after requeues.

finagle-core/src/main/scala/com/twitter/finagle/service/RequeueFilter.scala
olix0r: oops!

olix0r:
Great feedback, @kevinoliver. Thanks.

A couple clarifications:
- RetryFilter isn't on the stack by default (that's okay, just making sure I'm not missing it somewhere).
- How do multiple retry modules interact with RetryBudgets?  I assume they should share retry budgets so that retries in either place are counted -- but the lower one would have to be wrapped in a `WithdrawOnlyRetryBudget`?

kevinoliver:
> RetryFilter isn't on the stack by default (that's okay, just making sure I'm not missing it somewhere).

yeah, i think this would need to change for Stack-based clients and perhaps slight modifications to ClientBuilder as well.

> How do multiple retry modules interact with RetryBudgets? I assume they should share retry budgets so that retries in either place are counted -- but the lower one would have to be wrapped in a WithdrawOnlyRetryBudget?

by default, they share a budget and yeah the withdrawals need to be coordinated — this is all handled in the `Retries` code today. i'd recommend trying to keep the logic there if you want to keep sane.

it might be good to make sure others are on board with this direction (@mosesn @dschobel @roanta) before we go too deep here.

mosesn:
Yeah, maybe we should keep RetryFilter in ClientBuilder, and add the improvement there?  @olix0r we've been exploring using ClientBuilder as a tool for configuring application-level logic, and keeping finagle transport / session-only.

olix0r:
@mosesn That's fair, though it gets hard to reason about if these features are spread around...

For context, linkerd uses several stacks to model features at different phases of routing. We'd want this type of retry to be instrumented [in the highest (logical name) stack](https://github.com/BuoyantIO/linkerd/pull/370/files), while the requeues should probably be down in the lowest (concrete client) stack.

As I understand it I should pursue this as:
- Change RetryFilter (and not RequeueFilter) to use a ResponseClassifier
- Change ClientBuilder.newService (and not newClient) to install a RetryFilter module at the top of the stack.
  - Since ClientBuilder.newService already replaces the requeue filter with Retries.moduleWithRetryPolicy -- that does _not_ count withdrawls when a retry policy is set, the RetryFilter module _should_ count with drawls.

Then, linkerd may reuse these stack modules in slightly different locations.

olix0r:
I notice that there's some overlap between ResponseClassifier and RetryPolicy: both include a PartialFunction that determines whether something is retryable. There's a bit of a mismatch between Retries.Policy and, ResponseClassifiers, and RetryFilter, though:

Retrie.Policy only considers a `Try[Nothing]`:
```scala
object Retries {
  ...
  private[twitter] case class Policy(retryPolicy: RetryPolicy[Try[Nothing]]) {
    def mk(): (Policy, Stack.Param[Policy]) =
      (this, Policy.param)
  }
}
```

whereas RetryFilter uses a policy on `(Req, Try[Req])`
```scala
class RetryFilter[Req, Rep](
    retryPolicy: RetryPolicy[(Req, Try[Rep])],
    timer: Timer,
    statsReceiver: StatsReceiver,
    retryBudget: RetryBudget,
    classifier: ResponseClassifier = PartialFunction.empty)
  extends Filter[Req, Rep, Req, Rep] {
```

and ResponseClassifier matches simply on `ReqRep`:
```scala
  type ResponseClassifier = PartialFunction[ReqRep, ResponseClass]
```

Furthemore, `Retries.Budget` provides _requeueBackoffs_:
```scala
  case class Budget(
      retryBudget: RetryBudget,
      requeueBackoffs: Stream[Duration] = Budget.emptyBackoffSchedule) {
    def this(retryBudget: RetryBudget) =
      this(retryBudget, Budget.emptyBackoffSchedule)

    def mk(): (Budget, Stack.Param[Budget]) =
      (this, Budget)
  }
```

which duplicates information in a RetryPolicy.

Why are there so many ways of expressing essentially the same configuration?  Is it really a practical that someone would want to have disjoint policies between requeues and retries?  Certainly the budgets should be accounted together... wouldn't it be desirable to consider backoffs together as well?  I.e. if a request is retried and then requeued -- I would expect this to consume two backoffs from the stream and not one backoff from two streams.

What's the motivation for separating retry and requeue policies? It seems that it will make it much harder to construct a coherent retry policy.

olix0r:
I see that _moduleWithRetryPolicy_ (which is configured in ClientBuilderClient.newService) applies a RetryExceptionsFilter arund the RequeueFilter:
```scala
  private[finagle] def moduleWithRetryPolicy[Req, Rep]: Stackable[ServiceFactory[Req, Rep]] =
    new Stack.Module5[
      Stats,
      Budget,
      Policy,
      HighResTimer,
      ServiceFactory[Req, Rep]
    ] {
      def role: Stack.Role = Retries.Role

      def description: String =
        "Retries requests, at the service application level, that have been rejected " +
          "or meet the application-configured retry policy for transport level failures."

      def make(
        statsP: param.Stats,
        budgetP: Budget,
        policyP: Policy,
        timerP: HighResTimer,
        next: ServiceFactory[Req, Rep]
      ): ServiceFactory[Req, Rep] = {
        val statsRecv = statsP.statsReceiver
        val scoped = statsRecv.scope("retries")
        val requeues = scoped.counter("requeues")
        val retryBudget = budgetP.retryBudget
        val retryPolicy = policyP.retryPolicy

        val filters =
          if (retryPolicy eq RetryPolicy.Never) {
            newRequeueFilter(
              retryBudget,
              budgetP.requeueBackoffs,
              withdrawsOnly = false,
              scoped,
              timerP.timer,
              next
            )
          } else {
            val retryFilter = new RetryExceptionsFilter[Req, Rep](
              retryPolicy, timerP.timer, statsRecv, retryBudget)
            // note that we wrap the budget, since the retry filter wraps this
            val requeueFilter = newRequeueFilter(
              retryBudget,
              budgetP.requeueBackoffs,
              withdrawsOnly = true,
              scoped,
              timerP.timer,
              next
            )
            retryFilter.andThen(requeueFilter)
          }

        svcFactory(retryBudget, filters, scoped, requeues, next)
      }
    }
```

I suppose that this could be changed to use a RetryPolicy that is built around a ResponseClassifier.

Still, I'm unclear on the overlap between requeue and retry backoffs.

kevinoliver:
Requeues are meant to handle known-to-be-safe failures in a transparent manner. The initial assumption was that there should not be any backoffs between failures. That is not always a valid assumption depending on cluster topology and flexibility was given to allow customization. I think its reasonable to question if that original assumption is worth the potential risk.

As for why the backoffs should be separate, that part makes sense to me. You may want an exponential backoff for application failures while something else may be appropriate for a nack. Ultimately I think you should be able to share them across if you choose to. 

Hope that helps clarify.

olix0r:
That does help clarify backoffs, thanks.

olix0r:
I'm still a bit stuck on this.

I tried changing Retries.Policy to take a `RetryPolicy[ReqRep]` instead of a `RetryPolicy[Try[Nothing]]`.  Then, we could set a RetryPolicy on the stack with:

```scala

  def classified(backoffs: Stream[Duration], classifier: ResponseClassifier): RetryPolicy[ReqRep] =
    new RetryPolicy[ReqRep] {
      def apply(input: ReqRep): Option[(Duration, RetryPolicy[ReqRep])] =
        classifier.applyOrElse(reqrep, ResponseClassifier.Default) match {
          case ResponseClass.RetryableFailure =>
            backoffs match {
              case howlong #:: rest => Some((howlong, classified(rest, classifier)))
              case _ => None
            }
          case _ => None
        }
    }

  /**
   * Converts a `RetryPolicy[ReqRep]` to a `RetryPolicy[(Req, Try[Rep])]`.
   */
  private[finagle] def convertReqRepPolicy[Req, Rep](
    policy: RetryPolicy[ReqRep]
  ): RetryPolicy[(Req, Try[Rep])] =
    new RetryPolicy[(Req, Try[Rep])] {
      def apply(input: (Req, Try[Rep])): Option[(Duration, RetryPolicy[(Req, Try[Rep])])] = {
        val (req, rep) = input
        policy(ReqRep(req, rep)) match {
          case Some((howlong, nextPolicy)) => Some((howlong, convertReqRepPolicy(nextPolicy)))
          case None => None
        }
      }
    }
```

Then, I would change `Retries.moduleWithRetryPolicy` to install a RetryFilter instead of a RetryExceptionsFilter.

I thought this approach would be equitable, since Retries.Policy is `private[twitter]`, however ClientBuilder exposes a `retryPolicy(rp: RetryPolicy[Try[Nothing]])`.  Furthermore, it seems undesirable that this RetryPolicy is distinct from the ResponseClassifier on the stack -- we'd want a way to unify these...  We could introduce a new stack module, e.g. Retries.moduleWithClassifier, that installs the proper retryFilter based on the response classifier, though it would need a new parameter type to provide the backoff stream.  Having _both_ `Retry.Policy(backoff, policy)` and `Retry.Backoff(backoff)` veers into a redundant, confusing world.

At this point, I'm satisfied that I can introduce classification-based retries to linkerd properly using RetryFilter.

I could use some guidance, however, on where to cut this feature in finagle, if at all.  The question is now about Finagle's multiple configuration APIs and how to expose this without having conflicting ways to configure retries... and we also probably don't want to break public APIs due to this? I'm happy to spend more time on this if this is a desired feature and I can get a clearer sense of how this should be configured. Otherwise, I may just put this down.

olix0r:
Unrelated aside:

is RetryPolicy.backoff safe?
```scala
  def backoff[A](
    backoffs: Stream[Duration]
  )(shouldRetry: PartialFunction[A, Boolean]): RetryPolicy[A] =
    RetryPolicy { e =>
      if (shouldRetry.applyOrElse(e, AlwaysFalse)) {
        backoffs match {
          case howlong #:: rest =>
            // XXX this can blow the stack, right?
            Some((howlong, backoff(rest)(shouldRetry)))
          case _ =>
            None
        }
      } else {
        None
      }
    }
```

It seems to be recursive in a way that can blow the stack.  Is the assumption here that retries are constrained by some maximum number less than the max stack size?  Should this be enforced---or at least documented?

mosesn:
RetryPolicy construction is safe because it's lazy–it requires that you have the failure in hand before constructing the next one.  We usually pop back up the stack before using it again, so we don't go deeper in the stack.

I agree that we don't want to break the ClientBuilder API.  I think it's worth noting that RetryFilter does what you want already without having to use ReqRep, as long as you do it outside of the finagle stack.

I don't really know what the right thing to do with this is.  It seems tricky that we want to be able to use application-layer information to inform session-layer decisions.  I think we don't know the right balance yet.

olix0r:
@mosesn Thanks -- super helpful.  I'll share what we do in linkerd and we can spend some more time thinking about how to integrate this sort of thing into finagle.

olix0r:
linkerd's implementation of a ResponseClassifier-backed RetryFilter: https://github.com/BuoyantIO/linkerd/pull/403

mosesn:
We ended up that this wasn't the right way forward.  We may come up with richer application-level tools in the future, but for now we don't want to add this to finagle.

PR #507, opened 2016-05-19
Add channel stats to Netty4 clients

### LINKED ISSUE #506: netty4: no channel stats on clients
finagle-netty4 includes a ChannelStatsHandler (analogous to netty3's channel stats handler). However, it appears that netty4 channel stats are only recorded on servers and not clients:
```
:; git grep -Fl ChannelStatsHandler finagle-netty4/src/main
finagle-netty4/src/main/scala/com/twitter/finagle/netty4/channel/ChannelStatsHandler.scala
finagle-netty4/src/main/scala/com/twitter/finagle/netty4/channel/Netty4ServerChannelInitializer.scala
```

It seems stats just aren't wired into the netty4 client yet. I'm happy to give it a go if no one's working on it.

olix0r:
Problem

Netty3 includes channel stats (connects, closes, etc) for clients, but the
Netty4 client pipeline does not include channel stats.

Solution

Add ChannelStatsHandler to the Netty4 client pipeline.

Fixes #506

### DIFF
--- a/finagle-netty4/src/main/scala/com/twitter/finagle/netty4/channel/ChannelStatsHandler.scala
+++ b/finagle-netty4/src/main/scala/com/twitter/finagle/netty4/channel/ChannelStatsHandler.scala
@@ -50,12 +50,21 @@ private[netty4] class ChannelStatsHandler(statsReceiver: StatsReceiver)
     connectionCount.get()
   }
 
-
+  private[this] def getOrMkChannelStats(ctx: ChannelHandlerContext): ChannelStats = {
+    val channelStatsAttr = ctx.attr(ConnectionStatsKey)
+    channelStatsAttr.get match {
+      case null =>
+        val cs = ChannelStats(new AtomicLong(0), new AtomicLong(0))
+        channelStatsAttr.set(cs)
+        cs
+      case cs => cs
+    }
+  }
 
   override def channelActive(ctx: ChannelHandlerContext): Unit = {
     ctx.attr(ChannelWasWritableKey).set(true) //netty channels start in writable state
     ctx.attr(ChannelWritableDurationKey).set(Stopwatch.start())
-    ctx.attr(ConnectionStatsKey).set(ChannelStats(new AtomicLong(0), new AtomicLong(0)))
+    getOrMkChannelStats(ctx)
     connects.incr()
     connectionCount.incrementAndGet()
 
@@ -64,7 +73,7 @@ private[netty4] class ChannelStatsHandler(statsReceiver: StatsReceiver)
   }
 
   override def write(ctx: ChannelHandlerContext, msg: Object, p: ChannelPromise) {
-    val channelWriteCount = ctx.attr(ConnectionStatsKey).get.bytesWritten
+    val channelWriteCount = getOrMkChannelStats(ctx).bytesWritten
 
     msg match {
       case buffer: ByteBuf =>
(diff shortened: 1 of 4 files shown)


olix0r:
This isn't particularly tested, etc. Happy to polish more if this change is helpful.

mosesn:
LGTM

mosesn:
Removing "ship it" label temporarily until you have more confidence in it.

olix0r:
What I've observed is that the lifecycle (for an http/1.0 request) appears to be:

1. write
2. channelActive
3. channelRead
4. close
5. channelInactive

The _write_ blows up because it tries to access `ctx.attr(ConnectionStatsKey)` before channelActive is called, and then _channelInactive_ is called, which may also throw an exception once it tries to access the ctx.

olix0r:
I believe that there are additional conditions (errors) where channelInactive may be called without channelActive being called -- I'll address that if I can reproduce.

dschobel:
oh I think I know what this is... we had a change a few weeks ago so that the transport promise is satisfied outside the pipeline which means that you can see writes before channel active fires. handlers that assume writes come after channel active and rely on the latter to init state will do what you're seeing. 

We moved the init logic in all of ours handlers to handlerAdded but looks like we missed one...

@olix0r can you move  the NPE'ing state's init code  (just `ctx.attr(ConnectionStatsKey).set(ChannelStats(new AtomicLong(0), new AtomicLong(0)))` I think) into handlerAdded and see if that doesn't fix it?

olix0r:
(btw, @dschobel, updated)

finagle-netty4/src/main/scala/com/twitter/finagle/netty4/channel/ChannelStatsHandler.scala
```diff
@@ -99,13 +101,16 @@ private[netty4] class ChannelStatsHandler(statsReceiver: StatsReceiver)
...
-    val channelStats = ctx.attr(ConnectionStatsKey).get
 
+    val channelStats = ctx.attr(ConnectionStatsKey).get
     connectionReceivedBytes.add(channelStats.bytesRead.get)
     connectionSentBytes.add(channelStats.bytesWritten.get)
 
-    val elapsed = ctx.attr(ConnectionDurationKey).get()
-    connectionDuration.add(elapsed().inMilliseconds)
+    ctx.attr(ConnectionDurationKey).get match {
+      case null => // the connection didn't initialize
```
dschobel: can you reproduce `channelInactive` firing when `channelActive` didn't?

dschobel:
can you add a test case or update the existing ones so that writes happen before `channelActive`?

dschobel:
lgtm otherwise

finagle-netty4/src/main/scala/com/twitter/finagle/netty4/channel/ChannelStatsHandler.scala
olix0r: i believe that this happens if there are errors in the write() call (which precedes channelActive in clients)

olix0r:
minor fixup of tests -- it was hard to reproduce the null case fully in tests (because of mocks and whatnot)

mosesn:
LGTM

mosesn:
Merged here: https://github.com/twitter/finagle/commit/6b277536a49fb6c4a40032c7e03635c45bc0479b thanks!

PR #514, opened 2016-06-06
Move HTTP Context filters into the Http client and server stacks

olix0r:
Problem

The headers used to propagate context information are hardcoded into the Http
client and server, provided by `Http.{Client,Server}.newDispatcher`.

In linkerd, we would like to admit linkerd-specific headers, and so we would
prefer to replace these filters.

Solution

Insert a ServerContextFilter module at the top of the server stack and a
ClientContextFilter module at the bottom of the client stack. Include the
DtabFilter in the ServerContextFilter module since it extracts similar
contextual information.


mosesn:
This seems reasonable to me.  I can't think of any reason why it shouldn't work this way.

olix0r:
I'm still trying to sort through a few things I'd like to do as a part of this change:
- I'd like to move Dtab-encoding out of the dispatcher and into the stack
- I think there's a minor problem with retries and dtab-encoding: if a request with Dtab overrides is retried, an error will be logged because Dtab-local headers will be present on the request when retried.
- I don't yet understand how this should interact with clients ClientBuilder -- currently the Http codec installs context filters. It appears that, for example, ClientBuilder only ever uses `StackClient.newStack` and never `Http.Client.stack`, and so we have to replicate customizations from http's stack into http's codec...

olix0r:
OK, at this point I've moved the http client dispatcher's Dtab behavior into a DtabFilter.Injector class and added this into the client context module.  Similarly, I've deprecated the (pre-httpx) DtabFilter.Finagle in favor of DtabFilter.Extractor, which is installed via the server context module.

Still needs CHANGES update. What else?

Thoughts?

finagle-http/src/main/scala/com/twitter/finagle/http/filter/DtabFilter.scala
```diff
@@ -40,7 +40,41 @@ object DtabFilter {
...
+  class Injector extends SimpleFilter[Request, Response] {
+    def apply(req: Request, service: Service[Request, Response]): Future[Response] = {
+      // XXX won't this behave poorly with retries?
+      // val dtabHeaders = HttpDtab.strip(req)
+      // if (dtabHeaders.nonEmpty) {
+      //   // Log an error immediately if we find any Dtab headers already in the request and report them
+      //   val headersString = dtabHeaders.map({case (k, v) => s"[$k: $v]"}).mkString(", ")
+      //   log.error(s"discarding manually set dtab headers in request: $headersString\n" +
+      //     s"set Dtab.local instead to send Dtab information.")
+      // }
```
olix0r: I just commented out for the purposes of discussion. As I mentioned earlier, I think that this error reporting should be removed, since it does not play well with retries.

finagle-http/src/main/scala/com/twitter/finagle/http/filter/DtabFilter.scala
```diff
@@ -40,7 +40,41 @@ object DtabFilter {
...
+  }
+
+  /**
+   * Modifies each request with Dtab encoding from Dtab.local and
+   * streams chunked responses via `Reader`.  If the request already
+   * contains Dtab headers they will be dropped silently.
+   */
+  class Injector extends SimpleFilter[Request, Response] {
+    def apply(req: Request, service: Service[Request, Response]): Future[Response] = {
+      // XXX won't this behave poorly with retries?
```
atollena: Good catch. Could we check if the Dtab has already been injected? 
You could simply add a value in `request.ctx` for that (or perhaps you have a better idea?).

atollena: The fact that Dtab headers were silently removed from finagle requests has cost many hours of debugging, I would prefer that we keep it if it's not too hairy.

olix0r: Good idea! I didn't know about request.ctx -- let me know what you think.

finagle-http/src/main/scala/com/twitter/finagle/http/filter/DtabFilter.scala
```diff
@@ -40,7 +47,45 @@ object DtabFilter {
...
+    def respondToInvalid(req: Request, msg: String) = invalidResponse(msg)
+  }
+
+  /**
+   * Modifies each request with Dtab encoding from Dtab.local and
+   * streams chunked responses via `Reader`.  If the request already
+   * contains Dtab headers they will be dropped silently.
+   */
+  class Injector extends SimpleFilter[Request, Response] {
+    private[this] val hasSetDtab = Request.Schema.newField[Boolean](false)
```
atollena: since this is a constant it should be capitalized and go into the companion object.

atollena:
lgtm.

roanta:
Thanks @olix0r! This was merged and landed in https://github.com/twitter/finagle/commit/ce4ab7ffd917104efe333b58d1409df00aa4b408

PR #513, opened 2016-05-31
Extract transport agnostic zipkin code into finagle-zipkin-core

sveinnfannar:
Problem

Address issue https://github.com/twitter/finagle/issues/465 (Add http and kafka transports to zipkin-tracer).
Zipkin can receive traces via Kafka, HTTP and Scribe. Some Finagle users (including Quizup) already using Kafka are reluctant to introduce Scribe into their stack as it is no longer maintained.
So the purpose of this PR is to support Kafka as a transport for Zipkin traces, as well as structuring the code such that it allows other transports to be added (like HTTP).

Solution

Extract most of finagle-zipkin into finagle-zipkin-core, leaving finagle-zipkin with only scribe specific stuff (with no api breaking changes to finagle-zipkin of course). ~~Then implement finagle-zipkin-kafka, which depends on finagle-zipkin-core~~. Kafka transport will be introduced in a separate PR.

This structure was initially suggested by @kevinoliver in https://github.com/twitter/finagle/issues/465#issuecomment-183012155. 

Todo

 - [x] Extract common code into finagle-zipkin-core (largest part)
 - [ ] ~~Implement finagle-zipkin-kafka (in progress)~~ (separate PR)
 - [x] Clean up Pants and SBT dependencies

Questions

@mosesn can you check this out and provide some feedback?
@adriancole is this in line with what you had envisioned?

Issue #465, opened 2016-02-11
Add http and kafka transports to zipkin-tracer


adriancole:
Many people aren't running scribe and don't want to run scribe. Zipkin used to only work with scribe, but it now works with http and kafka.

I've pasted a copy of a patched http raw zipkin tracer below.. Is something like this possible to add to Finagle directly?

```scala
  /** Logs spans via POST /api/v1/spans. */
  override def logSpans(spans: Seq[Span]): Future[Unit] = {
    // serialize all spans as a thrift list
    val serializedSpans = try {
      val transport = new TMemoryBuffer(0)
      val oproto = new TBinaryProtocol(transport)
      oproto.writeListBegin(new TList(TType.STRUCT, spans.size))
      spans.foreach(_.toThrift.write(oproto))
      oproto.writeListEnd()
      transport.getArray()
    } catch {
      case NonFatal(e) => errorReceiver.counter(e.getClass.getName).incr(); return Future.Unit
    }

    val request = Request(Method.Post, "/api/v1/spans")
    request.headerMap.add("Host", hostname)
    request.headerMap.add("Content-Type", "application/x-thrift")
    request.headerMap.add("Content-Length", serializedSpans.length.toString)
    request.content =  Buf.ByteArray.Owned(serializedSpans)

    Trace.letClear { // Don't recurse tracing by tracing sending spans to the collector.
      client(request)
        .onSuccess(_ => okCounter.incr())
        .rescue {
          case NonFatal(e) => errorReceiver.counter(e.getClass.getName).incr(); Future.Done
        }.unit
    }
  }
```

https://github.com/openzipkin/zipkin/blob/master/zipkin-web/src/main/scala/com/twitter/finagle/zipkin/thrift/HttpZipkinTracer.scala
https://github.com/openzipkin/zipkin/blob/master/zipkin-web/src/test/scala/com/twitter/zipkin/web/HttpZipkinTracerTest.scala

adriancole:
cc @kristofa I know y'all at SoundCloud are using zipkin, though possibly more interested in kafka

kevinoliver:
@adriancole seems reasonable. It'd probably need a bit of work because finagle-zipkin brings in the thrift tracer via [service loading](https://github.com/twitter/finagle/blob/develop/finagle-zipkin/src/main/resources/META-INF/services/com.twitter.finagle.tracing.Tracer).

[...]

PR #515, opened 2016-06-07
Kafka transport for Zipkin traces

sveinnfannar:
Problem

Zipkin can receive traces via Kafka, HTTP and Scribe. Some Finagle users (including Quizup) already using Kafka would like to avoid adding Scribe to their stack as it is no longer maintained.
The purpose of this PR is to support Kafka as a transport for Zipkin traces.
Partially addressing issue #465.

Solution

In #513 I extracted core components of finagle-zipkin into finagle-zipkin-core. Building off that I added a module finagle-zipkin-kafka with kafka specific code and a service loader. It's interface is very similar to [ScribeZipkinTracer] found in finagle-zipkin.

Ping: @adriancole


sveinnfannar:
@mosesn: this PR depends on #513 so the diff is misleading. Is that fine or should I base the PR against the #513 branch? Do you have any suggestions?

sveinnfannar:
@adriancole, I used kafka-clients version 0.8.2.2 because that's the version [zipkin-collector uses](https://github.com/openzipkin/zipkin/blob/master/zipkin-collector/kafka/pom.xml#L33).
Do you think we should use a newer version since 0.8.2.2? One reason to use a newer version is that 0.8.2.2 is quite old. The downside is that clients cannot be newer than the brokers.

eirslett:
Is this running in production anywhere yet? (Or isn't that possible?)

sveinnfannar:
@eirslett: no not yet. I plan on productionizing this at Quizup this week.

eirslett:
Nice! :-D

schrepfler:
Considering the Kafka collector is pretty recent and there's low need to keep compatibility kept perhaps there's a chance to start with an updated stack and use Kafka 0.10.0.0 from the get-go?

sveinnfannar:
@schrepfler thats a very good point. Here are the pros and cons in my opinion:

Pros:
 - Lot 'o bug fixes, features and stability improvements

Cons:
 -  Kafka clients cannot talk to older clusters so users will have to upgrade their clusters
 - The Zipkin consumer uses 0.8.2.2 and there is a [performance impact on 0.10 brokers](http://kafka.apache.org/documentation.html#upgrade_10_performance_impact) when dealing with older consumers. However that can be avoided with some configuration if necessary

As far as I see there are no compelling reasons to keep backwards compatibility with older versions of kafka. I'm going to change to 0.10.

sveinnfannar:
I squashed the commit "Use kafka-clients-0.10 and use new config options" into the previous one

kevinoliver:
@sveinnfannar been busy for the past few days, hoping to take a look at this in the next few days.

finagle-zipkin-kafka/src/main/scala/com/twitter/finagle/zipkin/Flags.scala
```diff
@@ -0,0 +1,13 @@
+package com.twitter.finagle.zipkin
+
+import com.twitter.app.GlobalFlag
+import com.twitter.finagle.zipkin.core.Sampler
+import java.net.InetSocketAddress
+
+object hosts extends GlobalFlag[Seq[InetSocketAddress]](
+  Seq(new InetSocketAddress("localhost", 9092)),
+  "Initial set of kafka servers to connect to, rest of cluster will be discovered (comma separated)")
```
kevinoliver: s/set/list/

finagle-zipkin-kafka/src/main/scala/com/twitter/finagle/zipkin/Flags.scala
```diff
@@ -0,0 +1,13 @@
...
+
+object initialSampleRate extends GlobalFlag[Float](
+  Sampler.DefaultSampleRate,
+  "Initial sample rate")
```
kevinoliver: its good to call out that the value must be `[0.0 - 1.0]`

finagle-zipkin-kafka/src/main/scala/com/twitter/finagle/zipkin/kafka/KafkaRawZipkinTracer.scala
```diff
@@ -0,0 +1,67 @@
...
+import com.twitter.finagle.stats.StatsReceiver
+import com.twitter.finagle.util.DefaultTimer
+import com.twitter.finagle.zipkin.core.{Span, RawZipkinTracer}
+import com.twitter.util._
+import java.util
+import org.apache.kafka.clients.producer._
+import org.apache.thrift.protocol.{TBinaryProtocol, TList, TType}
+import org.apache.thrift.transport.TMemoryBuffer
+
+object KafkaRawZipkinTracer {
```
kevinoliver: can we minimize visibility?

finagle-zipkin-kafka/src/main/scala/com/twitter/finagle/zipkin/kafka/KafkaRawZipkinTracer.scala
```diff
@@ -0,0 +1,67 @@
...
+  private[this] val okCounter = statsReceiver.scope("log_span").counter("ok")
+  private[this] val errorReceiver = statsReceiver.scope("log_span").scope("error")
+
+  private[kafka] lazy val sendCallback = new Callback {
+    override def onCompletion(metadata: RecordMetadata, exception: Exception): Unit =
+      if (exception == null) okCounter.incr()
+      else errorReceiver.counter(exception.getClass.getName).incr()
+  }
+
+  override def sendSpans(spans: Seq[Span]): Future[Unit] = {
```
kevinoliver: rm `override` since its abstract in the base class

finagle-zipkin-kafka/src/main/scala/com/twitter/finagle/zipkin/kafka/KafkaRawZipkinTracer.scala
```diff
@@ -0,0 +1,67 @@
...
+    Try {
+      val serializedSpans = spansToThriftByteArray(spans)
+
+      // Producer#send appends the message to an internal queue and returns immediately.
+      // Messages are then batch sent asynchronously on another thread and the callback invoked.
+      producer.send(new ProducerRecord(topic, serializedSpans), sendCallback)
+    }.onFailure { e =>
```
kevinoliver: we try to hoist function literals into member variables to avoid allocations at runtime.

so this would be:

```
private[this] val onSendFailure: Throwable => Unit = 
  e => errorReceiver.counter(e.getClass.getName).incr()
```

finagle-zipkin-kafka/src/main/scala/com/twitter/finagle/zipkin/kafka/KafkaRawZipkinTracer.scala
```diff
@@ -0,0 +1,67 @@
...
+      errorReceiver.counter(e.getClass.getName).incr()
+    }
+
+    Future.Unit
+  }
+
+  private[this] def spansToThriftByteArray(spans: Seq[Span]): Array[Byte] = {
+    // serialize all spans as a thrift list
+    val transport = new TMemoryBuffer(0)
```
kevinoliver: feel free to ignore: is there a decent guess at an initial size? even if its wrong, i suspect it'll be better than 0.

finagle-zipkin-kafka/src/main/scala/com/twitter/finagle/zipkin/kafka/KafkaRawZipkinTracer.scala
```diff
@@ -0,0 +1,67 @@
...
+    val protocol = new TBinaryProtocol(transport)
```
kevinoliver: feel free to ignore: in finagle-thrift we have `Protocols.binaryFactory` which has stats and does some optimizations.

kevinoliver:
@sveinnfannar apologies for the delay in getting to this. looks good though. i don't have too much in the way of comments.

one question re the client dependency — how do kafka clients work for other versions of servers? and what will happen for users on other versions of the client?

kevinoliver:
So we've discussed this a bit over here and our preference is for this to be a repo in the [finagle organization](https://github.com/finagle/). This gives you full autonomy over the project and lets us stay focused on the core protocols that we use.

adriancole:
ps kafka servers are backwards compatible, but not the other way around

adriancole:
@kevinoliver sounds good.. how does mechanics like version bumps/releases occur from there? Is that something the finagle team conduct?

sveinnfannar:
@kevinoliver I agree, a separate project makes sense.
I'll get everything set up in a separate repository and transfer the ownership to the finagle organization.

kevinoliver:
👍  @sveinnfannar 

@adriancole releases and so on would be owned by the project's owners (perhaps you and @sveinnfannar to start?)

adriancole:
thanks. Reason I ask is that we host some projects in openzipkin org (ex
brave (java tracer), ruby tracer, go tracer).

In OpenZipkin org, we have access to sonatype etc via bintray, can do
tag-driven releases on-demand, change teams etc.

In the Finagle org, we'd need to understand what automation, if any, we
might be lacking, or additional technology folks might need to learn or
become responsible for.

Can you elaborate a little on how releases would work if this were in
Finagle's org?

kevinoliver:
@adriancole to be honest I don't know how this works. @vkostyukov has [finch](https://github.com/finagle/finch) there and @samn has [finagle-clojure](https://github.com/finagle/finagle-clojure) and as far as I know they do their own releases. Maybe one or both can comment. Not sure what concerns you have with regards to automation.

Another option is to not use the Finagle organization!

adriancole:
ahh I see. so looking at finch, it appears it is published under the
"com.github.finagle" group id (as opposed to com.twitter).
http://search.maven.org/#search%7Cga%7C1%7Ccom.github.finagle

This validates what you mentioned, which is that releases are not
coordinated or performed by Twitter folks. In either project, I don't
see evidence of tag-driven release automation (ex. push this tag and
magically things appear in maven central).

Since we aren't re-using any automation or inheriting version bump
efforts (like we would if these transports were in finagle's repo),
I'd suggest we host this in OpenZipkin's org.

@sveinnfannar @kristofa @eirslett you game for us creating a repo in
OpenZipkin for zipkin reporters to finagle?

We would publish them under the group id io.zipkin.finagle (and do all
transports in the same repo). It is easier for us that way as we don't
need to displace time on creating and maintaining a one-off release
process. Also people in the OpenZipkin org are more likely to know the
nature of zipkin updates I think.

The only downside is that people using finagle will have to look at
the OpenZipkin repo for zipkin integration as opposed to the Finagle
one. This implies at least a README change in finagle-zipkin here (in
finagle's repo) so that people can be directed accordingly.

vkostyukov:
@adriancole, @kevinoliver,

Yes, `"com.github.finagle"` (and perhaps `"io.github.finagle"`) is Maven group id we use for community projects (including Finch, finagle-postgres, etc). What we usually do is to create a new project with appropriate ownership. Next step is to file a ticket to Sonatype and ask them for an access from your personal account to one of those group ids. The rest of the magic is primary depends on your imagination. I recall Travis was trying to standardize the way we publish community project, but I don't think it went anywhere. So basically you set up sbt/Maven project and publish things manually (on demand). In most of the cases, it's nothing more than just `sbt +publishSigned`.

I'd probably vote for hosting this project under the Finagle org and I'm happy to help with all the setup, but I can totally understand if you decide to go with OpenZipking org instead.

kevinoliver:
> The only downside is that people using finagle will have to look at
the OpenZipkin repo for zipkin integration as opposed to the Finagle
one. This implies at least a README change in finagle-zipkin here (in finagle's repo) so that people can be directed accordingly.

@adriancole If you go this route, that seems reasonable to me.

sveinnfannar:
Sorry for the radio silence, I've been very busy the couple of days.
I will start working on this again on Monday.

adriancole:
PS I got feedback from gitter from @kristofa that OpenZipkin is preferred, but it doesn't mean the codebase couldn't move to Finagle's at a later time (since it is self-contained).

adriancole:
I've opened https://github.com/openzipkin/zipkin-finagle on this topic. If it turns out folks are against using it, we'll just delete it empty. @sveinnfannar you have write access. fyi @abesto

mosesn:
Looks like the work here has moved to openzipkin/zipkin-finagle#1 so I'm going to close this PR.  Pretty excited about this change folks!

Issue #343, opened 2015-02-16
memcached zipkin integration bug


kristofa:
We recently noticed, since upgrading to Finagle 6.24, that memcached spans representing GET operations look fine initially (using zipkin-web) but when refreshing the full trace minutes later they have a much longer and incorrect duration. This indicates that memcached GET operation spans are submitted twice with different properties.

Initially the spans have expected annotations but are lacking the `Hit` or `Miss` custom annotations that are added in the memcached filter.

After some time when refreshing the span durations are wrongly updated but they also have the 'Hit' or 'Miss' annotations AND `finagle.flush` annotations.  The last one indicates the span was in the `DeadLineSpanMap` till the deadline was reached.

When looking in more detail at the code I think I know what is happening.  When the `Client Received` annotation is set the span is submitted to the collector/scribe and removed from the `DeadLineSpanMap` as expected. However after that the memcached filter kicks in which submits the hit / miss annotation. When this happens `DeadlineSpanMap#update` is executed. The DeadlineSpanMap won't find the trace id anymore in its map and will create a new Span with same id (Trace + span + parent span id).  This new span will contain the cache hit/miss annotation but will never receive any other annotations and so will get send of once the deadline expires.

memcached SET operation spans are fine as those are not updated by the memcached custom filter.

So I think we should not set any more annotations after an annotation that completes the span is submitted.

I'll look into making a pull request to fix this.

mosesn:
Sounds like a plan!  Please let us know if you need any help.

kristofa:
Thanks, I can use some help, yes :)

The problem is that the `MemcachedTraceInitializer` filter is registered in the `TraceInitializerFilter.role`. The next filter in the chain is the `ClientTracingFilter` which submits cs, cr annotations. After the `ClientTracingFilter` ends the span by submitting cr annotation the `MemcachedTraceInitializer` filter still submits cache hit/miss annotations.

So the obvious way to go I guess is to replace the current `ClientTracingFilter` with an updated Memcached filter which combines submitting cs, cr annotations and cache hit/misses.

This does not sound too complicated but the problem is that I don't see a way to do this using the current Memcached clients. The Memcached clients don't seem to use `StackClient` and they only seem to allow overriding the Trace Initializer?

Any thoughts or suggestions would be really helpful. Thanks!

kristofa:
See #354

mosesn:
@kristofa is this something you're still thinking about?

kristofa:
[...]

PR #516, opened 2016-06-13
finagle-memcached: Bugfix memcached Zipkin integration

kristofa:
Problem

When getting a value from memcached the cache hit/miss annotation
was submitted after the cr annotation was submitted.

Submitting the cache hit/miss annotation after the cr annotation resulted
in the span being recreated and ending up in the DeadlineSpanMap.
When the deadline exceeded the span was submitted a 2nd time overwriting
the correct duration of the original span.  Once that happened
the span was also annotated with 'finagle.flush'.

Solution

Update the memcached specific tracing filter so memcached specific
annotations are submitted in between cs and cr annotations.

Result

The memcached span is not re-submitted a 2nd time and does not end
up in DeadlineSpanMap. The duration is correct when viewed in zipkin-web
and there is no 'finagle.flush' annotation anymore.

### DIFF
--- a/finagle-memcached/src/main/scala/com/twitter/finagle/Memcached.scala
+++ b/finagle-memcached/src/main/scala/com/twitter/finagle/Memcached.scala
@@ -11,7 +11,7 @@ import com.twitter.finagle.memcached._
 import com.twitter.finagle.memcached.exp.LocalMemcached
 import com.twitter.finagle.memcached.protocol.text.{MemcachedClientPipelineFactory, MemcachedServerPipelineFactory}
 import com.twitter.finagle.memcached.protocol.{Command, Response, RetrievalCommand, Values}
-import com.twitter.finagle.netty3.{BufChannelBuffer, Netty3Listener, Netty3Transporter}
+import com.twitter.finagle.netty3.{Netty3Listener, Netty3Transporter}
 import com.twitter.finagle.param.{Monitor => _, ResponseClassifier => _, ExceptionStatsHandler => _, Tracer => _, _}
 import com.twitter.finagle.pool.SingletonPool
 import com.twitter.finagle.server.{Listener, StackServer, StdStackServer}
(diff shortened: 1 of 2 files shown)


kristofa:
@mosesn 2nd attempt to fixing #343.

kristofa:
screen shot 2016-06-13 at 9 25 48 am

This screenshot from zipkin-web shows the behavior of the original code. While the memcached 'get' operation didn't even take 1ms the duration of the span ends up being more than 32 seconds because the span is submitted a 2nd time (finagle.flush annotation).

I tested the fix I submit here using a test with a local memcached and zipkin set-up.

finagle-memcached/src/main/scala/com/twitter/finagle/Memcached.scala
```diff
@@ -22,32 +22,31 @@ import com.twitter.finagle.tracing._
...
     }
   }
 
-  class Filter(tracer: Tracer) extends SimpleFilter[Command, Response] {
-    def apply(command: Command, service: Service[Command, Response]): Future[Response] =
-      Trace.letTracerAndNextId(tracer) {
-        val response = service(command)
+  object TracingFilter extends SimpleFilter[Command, Response] {
+    def apply(command: Command, service: Service[Command, Response]) = {
+      val response = service(command)
```
dschobel: doesn't sequencing this request dispatch before the rpc trace below leave the trace out of order with regards to any tracing lower in the stack?

finagle-memcached/src/main/scala/com/twitter/finagle/Memcached.scala
```diff
@@ -22,32 +22,31 @@ import com.twitter.finagle.tracing._
...
+      if (Trace.isActivelyTracing) {
         Trace.recordRpc(command.name)
         command match {
-          case command: RetrievalCommand if Trace.isActivelyTracing =>
+          case command: RetrievalCommand =>
             response onSuccess {
```
dschobel: no infix notation please.

finagle-memcached/src/main/scala/com/twitter/finagle/Memcached.scala
```diff
@@ -58,13 +57,18 @@ private[finagle] object MemcachedTraceInitializer {
                   Trace.recordBinary(key, "Hit")
                   misses.remove(key)
                 }
-                misses foreach { Trace.recordBinary(_, "Miss") }
+                misses foreach {
```
dschobel: ditto, no infix

finagle-memcached/src/main/scala/com/twitter/finagle/Memcached.scala
```diff
@@ -58,13 +57,18 @@ private[finagle] object MemcachedTraceInitializer {
...
+                  Trace.recordBinary(_, "Miss")
```
dschobel: pulling this trace call into a function literal shaves an allocation on the request path

dschobel:
thanks for the patch, one question about traces lower in the stack, otherwise just nits.

finagle-memcached/src/main/scala/com/twitter/finagle/Memcached.scala
kristofa: I think we are fine with these changes.
The order in which tracing events happen for memcached with this commit in place:

1. Initialize span (TraceInitializerFilter part of StackClient)
2. Add Client Send annotation (AnnotatingTracingFilter set-up in MemcachedTracingFilter.Module)
3. Execute memcached request
4. Add cache hit/miss annotations based on memcached response (MemcachedTracingFilter)
5. Add Client Received annotation (AnnotatingTracingFilter set-up in MemcachedTracingFilter.Module)

We can only add the cache hit/miss annotations once we got the response. The important change is that the memcached annotations are submitted in between the cs / cr annotations. The Client Received annotation should be the last annotation that gets submitted because it indicates the span is complete and ready to send off to back-end.

dschobel: Step #3 represents the rest of the client stack and we have no guarantees
that there's no tracing happening there so ideally we'd do the request
level tracing before dispatching the request through the rest of the stack.

I'm okay with leaving it as-is but can you leave a comment that this
implementation assumes no further tracing lower in the stack?

kristofa: I certainly want to add a comment but I would like to understand the problem. 

What could happen is that additional tracing annotations are added lower in the memcached client stack but I don't see why this would give a problem with adding the cache hit/miss annotations on the way out? 

Thank you for the quick feedback!

dschobel: Sorry I should have been more specific, the hit/miss annotations are fine, the issue is sequencing `Trace.recordRpc(command.name)` in line 47 after potential tracing lower in the stack. It seems like we'd want the cache command name to be traced before the remainder for the stack which is dispatching the command.

kristofa:
@dschobel I updated the code to submit the rpc name (command name) before dispatching the request.

dschobel:
We still need to guard the traceRpc call with isActivelyTracing so I think
a comment on the previous version is fine.

kristofa:
Inside 'traceRpc' there is a verification using 'isActivelyTracing' before submitting the annotation. 
But we should have no tracing further down the stack with memcached anyway so I reverted to previous version and added a comment as you suggested.

dschobel:
@kristofa ah good point, I didn't see the isTracing guard in record. anyway, I'm happy with this as-is. thanks again for the patch. 

lgtm

kristofa:
@dschobel Thanks for reviewing! I hope this gets merged as it will make our trace overviews usable again.

dschobel:
@kristofa it's been posted internally for review so I expect it'll merge and get synced out on Monday.

mosesn:
@kristofa this was published to `develop` https://github.com/twitter/finagle/commit/5c8862984b9f01e6863459d8bba4f548981b2629, and we're working on a release now.  thanks for the contribution!

kristofa:
I'm happy that I'm able to contribute. Thanks for open sourcing such a great software project!

Issue #517, opened 2016-06-17
finagle-http: client does not properly handle `Connection: close` 


olix0r:
BuoyantIO/linkerd#482 describes a situation where clients unexpectedly throw a ChannelClosedException.  The scenario is basically:

- Requests being issued from linkerd to a non-Finagle HTTP server on a persistent connection.
- Server responds to a message with `Connection: close`
- Server closes connection.
- Finagle client may issue additional requests on this connection
- requests fail with ChannelClosedException

Per [RFC2616](https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html):
> HTTP/1.1 defines the "close" connection option for the sender to signal that the connection will be closed after completion of the response. For example,
>
>       Connection: close
>
> in either the request or the response header fields indicates that the connection SHOULD NOT be considered `persistent' (section 8.1) after the current request/response is complete.

This seems like a race condition, as the error behavior is not easily triggered unless there are multiple concurrent requests on the connection.

I've written a [small Go HTTP server that reproduces this behavior](https://gist.github.com/olix0r/872f206ed22fce058fcd67c7fdc52e65).  This server can be started with `go run shutter.go`.  Then, configure a finagle HTTP client to issue requests to /$/inet/127.1/8080 (such that there are at least 2 requests issued in parallel) and an error will be encountered.

This issue is observed with both netty3 and netty4 http implementations.

Is this issue known?

I'm happy to help further diagnose/fix this problem, but any pointers are appreciated.

olix0r:
In one of the test, I see that out of about 60K total requests, with ~6.5K server-disconnects, we observe ~550 ChannelClosedExceptions (exceptions caused by 8.5% of disconnects).

olix0r:
I think I have a fix for this. Forthcoming...

vkostyukov:
I've never seen that we've handled that `Connection: close` header in Finagle. I guess we just ignore it right now? Looks like there is a window between receiving a response with `Connection: close` and `channelInactive` event in the pipeline, in which we see `ChannelClosedException`.

olix0r:
It looks like `ConnectionManager` is supposed to handle this:

```scala
  def observeResponse(response: Response, onFinish: Future[Unit]): Unit = synchronized {
    if ((!response.isChunked && response.contentLength.isEmpty) || !response.isKeepAlive) isKeepAlive = false
   ...
```

`Message.isKeepAlive` checks `Connection: close`.

[...]

PR #518, opened 2016-06-20
Ensure that Transport.collate's future is satisfied before its reader is closed

olix0r:
There is a subtle race condition between DelayedReleaseService and
HttpTransport. DelayedReleaseService may be notified that a stream has
completed before the underlying transport has observed this update. This allows
services to be released to the CachingPool before HttpTransport has an
opportunity to update its ConnectionManager's state. This may manifest as a
ConnectionClosedException, when a closed transport is returned as valid by
CachingPool.

Solution

Update Transport.collate to satisfy its closure promise before actually closing
the response stream. This allows the ConnectionManager to observe the state
update before a response's Readers are notified.

Fixes twitter/finagle#517


olix0r:
This seemed like the simplest fix to me, but I'm open to alternate approaches to resolving this race condition.

dschobel:
lgtm. I think this is the most clearest fix for the subtle rat's nest that we built from deferring close and trying to sequence these side effects.

olix0r:
merged https://github.com/twitter/finagle/commit/1b29f4e334f5f24f1c1ff2766369a75cad63561a

PR #520, opened 2016-06-23
RFC: Demonstrate OpenTracing support for Finagle

kcamenzind:
## Status

Request-for-comments about the basic approach before writing tests, documentation, etc.
## Problem

There is currently no direct support in Finagle for tracing requests across the wire using the OpenTracing API. This is an important feature, as RPC calls can be difficult to trace, and developers may use tracers in their systems other than Zipkin (which is Finagle's default tracer). Furthermore, use of the OpenTracing API considerably reduces the tracing surface area in Finagle and also decouples Finagle from implementation choices in tracing systems (like the particular representation of trace/span ids, etc).
## Solution

We implement OpenTracing filters for Finagle HTTP requests and responses that allow tracing of the request across the wire. The client-side filter takes the current tracing span and injects parameters into the HTTP request headers so that the span can be recreated on the server side. The server-side filter extracts these headers and traces the server response. In order to define the parent span of the HTTP requests, we have provided a method letCurrentSpan that sets the parent span for the scope of the request using local Contexts.

An example of how these changes could be implemented can be found in the examples HttpOpenTracingServer and HttpOpenTracingClient, in which client requests and server responses are both traced using these filters. The output of this trace, when using a LightStep tracer, is displayed below.
## Proposed Next Steps

Of course we would like initial reactions/feedback. If the basic approach seems reasonable, we would like to:
- respond to any finagle-team guidance per the above
- add more detail to the traces (beyond start/finish times)
- add tests and documentation
- perhaps add support for a non-HTTP transport (thriftmux? suggestions welcome)

Finally, we are just a few blocks away and are happy to talk in person about any of this if that would be easier. Thanks!

screen shot 2016-06-23 at 12 09 51 pm


olix0r:
Is there a reason why this has to be implemented in finagle-http?  Couldn't this be its own library that implements the proper filters?  It seems onerous for finagle-http to take on a dependency on opentracing (or any tracing implementation, for that matter).

finagle-http/src/main/scala/com/twitter/finagle/http/Codec.scala
```diff
@@ -521,3 +522,107 @@ private[finagle] class HttpServerTracingFilter[Req <: Request, Res](serviceName:
...
+  {
+    def apply(request: Req, service: Service[Req, Res]) = {
+      val rpcName = Contexts.local.getOrElse(rpcCtx, () => "Finagle HTTP Request")
+      val span = createSpanFromContext(tracer, rpcName)
+      var textMap = new HashMap[String,String]()
+
+      // NOTE: OpenTracing uses `inject` to propagate arbitrary states
+      // across process boundaries without tightly coupling to any particular
+      // wire format or object type.
+      tracer.inject(span, textMap)
```
olix0r: It might be preferable to implement a `com.twitter.tracing.Tracer` that's backed by `io.opentracing.Tracer` -- this way all of finagle's built-in annotations will be automatically integrated, without requiring duplication.  This also would make the opentracing backend auto-loadable as a plugin at runtime.

kcamenzind:
> It might be preferable to implement a com.twitter.tracing.Tracer that's backed by io.opentracing.Tracer -- this way all of finagle's built-in annotations will be automatically integrated, without requiring duplication.

While it would be possible to create an OpenTracing implementation of `com.twitter.tracing.Tracer`, there are numerous benefits to the approach suggested in this PR. OpenTracing is decoupled from any notion of a TraceId, so implementing it separately would remove inefficiencies of tracking both the TraceId and OpenTracing span across the wire. It also leaves the encoding of trace information up to the tracing implementation, so that finagle wouldn't have to directly depend on any specific format (such as the finagle-http `X-B3-` headers, finagle-thrift request/response headers, or finagle-mux `Treq` encoding).

Using OpenTracing as a separate tracing pathway also significantly reduces the amount of code needed for its implementation, as well as reducing the complexity of the code that finagle users need to write. This approach does require several new filters that are similar to the ones written for the built-in finagle tracing, but overall this approach leads to a cleaner, more concise, and easier to understand implementation.

> Is there a reason why this has to be implemented in finagle-http?

There's no reason that it has to be in finagle-http; it could definitely be moved though to an finagle-opentracing library (or something of that sort). There are already some http-specific tracing filters in the http codec, which is why I had originally written the opentracing filters there.

bensigelman:
One related issue with the existing Finagle `Tracer` architecture is that things like https://github.com/twitter/finagle/pull/515 are even *relevant* at the Finagle layer (and, for now, they are). With OpenTracing, there can be a Zipkin-Kafka OpenTracing implementation, and — best of all — Finagle doesn't need to know a thing about it. :)

IIRC, the Finagle code at `master` binds itself tightly to the structure and specific encoding of a [TraceId struct](https://github.com/twitter/finagle/blob/develop/finagle-core/src/main/scala/com/twitter/finagle/tracing/Id.scala#L58). The alternative @kcamenzind suggests in this PR focuses on the _semantics_ of the tracing operations (e.g., defining start times, finish times, logging events in between, and injecting/extracting from the wire format), obviously using the OpenTracing API in the process... if we moved in this direction, Finagle could deprecate and eventually remove the `TraceId` (which would be encapsulated by an equivalent OpenTracing implementation).

TL;DR, an approach like the one illustrated here would allow us to remove lots of tracing-related subtlety and complexity from the Finagle codebase while also decoupling it from the encoding and transport concerns of any particular tracing system.

mosesn:
This is a cool proof of concept, thanks for showing it off!

As @olix0r mentioned, finagle has an interface, `Tracer` which is supposed to be a generic API for tracing.  My understanding is that it's similar in intent to `opentracing-api`, where they're intended to provide a facade where a user can swap out the implementation.  With that said, when we were designing it, it was with zipkin / dapper-style tracing in mind, and we didn't have as many different use cases as opentracing has thought about, so my guess is that `OpenTracing` is more flexible.

So we're presented with a couple reasonable ways forward:

1.  Rewrite all tracing in finagle to be against `opentracing-api`, not against `Tracer`.  We provide an `opentracing-api` "implementation" which shells out to the `Tracer` facade in the mean time.
2.  Provide a `Tracer` “implementation” which shells out to the `opentracing-api`, that users can inject to take advantage of `opentracing-api`.

One thing to note is that all of these are preliminary steps, and choosing one of these steps doesn’t mean that we’ll never change our mind and do the other, or that one isn’t a first step in order to achieve the other.

We have a few key goals here that we have achieved with our existing tracer implementation and we want to preserve in the new world, and a few design restrictions we have for our (and our users’) sanity.

Tracing should work in your service automatically.  You might need to provide an implementation (we do this via service-loading today) and a remote location so we know where to send the spans, but it should ~just work~.  Users adding tracing individually to their service will not work for us.
Clients should be able to recursively propagate flags which can be interpreted uniformly across a company, and these should be usable to decide what is done with a span.  My understanding is that this should be easy to do with `opentracing-api` because you can send arbitrary metadata.
If a service receives a request, and because of that, sends a request, those dependent requests should be linked to the original request.  This property should be maintained even in an asynchronous context, as long as you’re using twitter’s typical asynchronous primitives.
We need to provide our users with a lightweight migration strategy.  Either this means we don’t change the protocol over the wire, or we ensure that we’re backwards compatible on the client-side and server-side.  After we’re moved over, we can get rid of the old protocol, but it would be a slow process.
Zipkin still needs to work with whatever solution we come up with.

Within these constraints, we have a decent amount of leeway, and we don’t have a strong preference one way or another.  We understand that you’re time limited, so even just moving us directionally the right way would be appreciated.

Anyway, we’re really pumped about this contribution, and look forward to working together to come up with a good solution here!

kcamenzind:
Hi Moses, thanks for the feedback and ideas! I realize that this first RFC didn't "just work" for the user, which needs to be addressed in whichever more complete solution we come up with. Some thoughts about implementation:

The core obstacle to including OpenTracing as a finagle tracer is that finagle `Tracers` pass around TraceIds with the specific `traceId-spanId-parentId` format, while OpenTracing `Tracers` rely on the span objects themselves,  which can have any implementation. This is an issue if you're trying to create a new span across the wire because OpenTracing has to have a reference to the parent span object, which it doesn't get with a `TraceId`. With that being said, I think that there are at least two reasonable implementations that seem relatively straightforward to pursue:

1. We could add another module to the `ServiceBuilder` stack that has the role `OpenTracing` and can be configured to by default do nothing, or if the service is built with an `.opentracing(tracer)`, it uses filters similar to the ones in this RFC to trace requests and send the spans across the wire. The positive of this approach is that it would be simple to implement and would have no compatibility issues with `finagle-tracing`. The downside to this approach would is that we're adding a completely separate pathway for tracing, which is inefficient and redundant.

2. Another approach that we could take would be to initialize tracers and then wrap them in a class, (let's call it `AnyTracer` for now) as follows:
 ``` scala
Service: service = new ServiceBuilder
	. ...
	.tracer(someFinagleTracer) // backwards compatible
	.build()
```
or
```scala
Service: service = new ServiceBuilder
    . ...
    .otTracer(someOpenTracingTracer)
    .build()
```
When either `tracer(someFinagleTracer)` or `otTracer(someOpenTracingTracer)` is called, finagle will then set the service's tracer to an `AnyTracer(someFinagleTracer)` or `AnyTracer(someOpenTracingTracer)`, depending on which has been called.

The `AnyTracer` class would essentially allow either the finagle or OpenTracing class to inherit its own methods, while also providing an implementation for the other tracer type's methods. This would allow the tracer to still use finagle's Annotations to record events, but will also allow for sending span objects (via inject and join) across the wire.

Roughly something like:
``` scala
class AnyTracer(ot: io.opentracing.Tracer) 
	def inject(args):
		ot.inject(args)
	def join(args):
		ot.join(args)
	def record(Record):
		// matches records to create and finish spans
		// and log other key events
    def letSpan(span: io.opentracing.Span)

class AnyTracer(ft: finagle.tracing.Tracer)
	def inject(args):
		// marshal the TraceId
	def join(args):
		// unmarshal the TraceId
	def record(Record):
		ft.record(Record)
        def letSpan(span: TraceId)
```
The only major difference regarding the filters would be to then call letSpan instead of letTraceId.

Hope this helps!

bensigelman:
A while back, @kcamenzind and I met with @mosesn in person (since we work about 4 blocks from each other) and devised a reasonable plan for this work. Unfortunately, nobody has had time to put that plan into action :)

Here’s a summary of where we ended up. Others are of course encouraged to chime in with other suggestions/opinions about a path forward as they see fit.

To review, Finagle presently has a `TraceId` concept (which, just to be super clear, is actually an object with [a traceId, parentId, spanId, and a sampled bit](https://github.com/twitter/finagle/blob/eddeb3b8552702f7fb2b5d756440552d7ad94947/finagle-core/src/main/scala/com/twitter/finagle/tracing/Id.scala#L58)). That TraceId can be stored in the Finagle context, then in various places Finagle essentially logs against that TraceId.

In OpenTracing, this would be modeled a little differently… rather than placing specific IDs into the Finagle context (via a wrapper object), a single [OpenTracing Span](http://opentracing.io/spec/#spans) would be stored in the context; and the various tracing-specific logging methods would write directly to the OpenTracing `Span` API (which has a timestamped logging facility of its own).

In a final, post-migration state, Finagle would have less code, and there would be no need for interfaces like [Record](https://github.com/twitter/finagle/blob/eddeb3b8552702f7fb2b5d756440552d7ad94947/finagle-core/src/main/scala/com/twitter/finagle/tracing/Tracer.scala#L26)... the core finagle code would just log events to the `Span` found in the context. This end state would also be even further decoupled from any specific tracing system (e.g., Finagle per se would no longer need to know about things like trace/span/parent ids, or even the sampling bit). So in this respect, the migration seems like a win both for tracing systems and for Finagle.

The tricky thing is that it’s a multi-step process assuming we aren’t willing to break existing users of Finagle’s tracing subsystem. Here’s one way that could work:

1. Support an OT Span in the Finagle context, log to it during the RPC lifecycle (e.g., at [https://github.com/twitter/finagle/blob/eddeb3b8552702f7fb2b5d756440552d7ad94947/finagle-core/src/main/scala/com/twitter/finagle/tracing/Tracer.scala#L35](these times)), and [do `Inject` and `Extract`](http://opentracing.io/spec/#inject-and-extract) at the process boundaries.
1. Create some sort of adapter to the OpenTracing `Tracer` interface that could bridge all of the various Span lifecycle events into the Finagle tracing scheme. I haven’t actually implemented it so I don’t want to overpromise, but my sense from reading the code is that this would be doable without changing existing Finagle tracing implementations.
1. Having done ^^^, remove the pre-existing Finagle tracing instrumentation (leaving the Span-based instrumentation in place)
1. Grant some reasonable period for Finagle tracing impls to migrate to something OT-based
1. "Someday," delete the bridge code from (2) above

Note that a Zipkin OpenTracing impl on one side of an RPC connection should be able to communicate with a peer running pre-OpenTracing Finagle+Zipkin without issue… i.e., there are no intended changes to the wire protocols anywhere in this proposal.

If anyone is interested in pursuing the path described above, it would be pretty fun code to write, and I'm happy to help out with any OpenTracing orientation/spin-up that's required. @kcamenzind is back at MIT for the school year and I am too swamped (and too unfamiliar with Scala) to do the brunt of the dev work myself.

virusdave:
Anyone actually working on this?  It seems far too useful to ignore.  I wish i had the spare time needed to do it myself.

kevinoliver:
@virusdave none of the core committers are working on this. but we are motivated to help get patches in if you or anyone else is interested in picking this up.

bensigelman:
@virusdave from the OpenTracing side, I also know of nobody actively working on this, though agreed that there are a lot of benefits and I would be happy to help orient anyone who wanted to take it on.

bryce-anderson:
Given that it seems decided that this doesn't belong in `finagle-http` and there hasn't been any activity on this since June 2016, I'm going to close. If there is interest in moving this forward, feel free to reopen it.

wsargent:
See https://github.com/opentracing-contrib/scala-finagle

PR #521, opened 2016-06-25
finagle-http: Remove body (and Content-Length) from 1xx, 204 and 304 responses

monkey-mas:
Problem

~~We currently have two problems. Firstly, as described in [RFC2616](https://tools.ietf.org/html/rfc2616) and
[RFC7231](https://tools.ietf.org/html/rfc7231), 204 and 304 responses must not contain a message-body. This
requirement was not enforced in finagle, i.e., a message-body can be
added. Moreover, a Content-Length header field can be added to these responses,
which is not allowed as mentioned in [RFC7230#section-3.3.2](https://tools.ietf.org/html/rfc7230#section-3.3.2).~~

```
We currently have two problems. Firstly, as described in RFC2616 and
RFC7230, 1xx, 204 and 304 responses must not contain a message-body. This
requirement was not enforced in finagle, i.e., a message-body can be
added. Moreover, a Content-Length header field can be added to 1xx and
204 responses, which is not allowed as mentioned in RFC7230#section-3.3.2.
```

Solution

~~First of all, we have to clear a message-body, which we call content in finagle,
if the response status code is either 204 or 304 to satisfy RFC7231. Secondly,
we need to make sure not to add a Content-Length header field in a 204 response
to follow RFC7230#section-3.3.2. With regard to a 304 response, a server may
send a Content-Length header as mentioned in the same section. Thus it'd be
better for us to allow users to set the field with a value of 0 if they like.
Most importantly, [we make sure that these things are achieved and the new system
is bugwards-compatible by provding a new filter](https://github.com/twitter/finagle/pull/521#issuecomment-228566007).~~

```
Firstly, we have to clear a message-body, which we call content in finagle,
if the response status code is either 1xx, 204 or 304 to satisfy RFC7230.
Then, we need to remove a Content-Length header field in 1xx and 204 responses
to follow RFC7230#section-3.3.2. With regard to a 304 response, we don't
intentionally remove it even though a Content-Length header field is set.
```

Result

We follow the specification of RFC with regard to a message-header and
message-body except that we have not added some header fields required in a
response with status code 304 to satisfy [RFC7232#section-4.1](https://tools.ietf.org/html/rfc7232#section-4.1); Any of the
following header fields must be generated: Cache-Control, Content-Location,
Date, ETag, Expires, and Vary.
~~As we achieve this by creating a new filter and set it to [Http.scala](https://github.com/monkey-mas/finagle/blob/finagle-6.35.0/finagle-http/src/main/scala/com/twitter/finagle/Http.scala#L298-L304), users
who are not willing to make use of the functionality can easily disable it
by removing the filter.~~


mosesn:
Thanks for submitting the PR!  We should make a decision about where we want to stick this behavior, and what we want finagle-http to enforce on behalf of its users.

I think that you're right that finagle should make it easy to follow the spec.  With that said, we need to make sure it's at least bugwards *migration* compatible.  For example, if your service is depending upon being able to do something that doesn't follow the spec, it would be nice to provide a clear migration to spec-compatibility for them.  This might be as simple as making validation optional.  Netty does this all the time, see [validateHeaders][0]

I think in the long-run we probably want to disallow constructing requests that are illegal, or configuring requests in an illegal way.  This will improve the user experience, since you'll fail when you do the illegal thing, rather than when you send the bad object into the client.  In the mean time, would it make sense to make request / response validation into a new filter?  This way, if folks decide that they really can't live with this behavior, they can simply remove the filter themselves.  It also separates concerns neatly, where the HttpServerDispatcher stays (mostly) just state machine stuff.

[0]: https://github.com/netty/netty/blob/3a9f47216143082bdfba62e8940160856767d672/codec-http/src/main/java/io/netty/handler/codec/http/DefaultLastHttpContent.java#L40

monkey-mas:
Thanks a lot for your quick and detailed response!

> With that said, we need to make sure it's at least bugwards migration compatible.
> ...
> In the mean time, would it make sense to make request / response validation into a new filter?

Yes, I like your idea and strongly agree with it. :)

With regard to the actual implementation, we can 1) create a new [filter](https://github.com/monkey-mas/finagle/commit/f9eee09a45679479ed559a43ef8be74792a2fffc#diff-2bb3b91aa92f68b7296ea75a97f5f610R14) and 2)add it to `Http.scala` like [this](https://github.com/monkey-mas/finagle/commit/f9eee09a45679479ed559a43ef8be74792a2fffc#diff-2d541d96a710e114d924373837e0077fR292) so that finagle follows the spec by default and if some users don't like the filter they can just remove it? I might be wrong, please let me know if this is what you meant :)

mosesn:
@monkey-mas 👍  that makes a lot of sense to me.

monkey-mas:
Hi @mosesn,

I've rebased and updated my commit! Though I think it's ready to be reviewed, one [test](https://travis-ci.org/twitter/finagle/jobs/140463912#L2777) from finagle-mysql failed... Actually it passes on my local machine.  If this is ok, please review my commits :)

finagle-http/src/main/scala/com/twitter/finagle/http/codec/HttpServerDispatcher.scala
```diff
@@ -31,6 +31,11 @@ private[finagle] class HttpServerDispatcher(
     service.close()
   }
 
+  private def mayHaveContent(status: Status): Boolean = status match {
```
mosesn: I don't *think* we need this anymore, since we have the new Filter now.  What do you think?

finagle-http/src/main/scala/com/twitter/finagle/http/filter/ModifyContentAndHeaderFieldFilter.scala
```diff
@@ -0,0 +1,55 @@
...
+
+/**
+ * Modify a content and header field filter in a 204 and 304 response.
+ * Remove a message-body and modify a Content-Length header field if necessary.
+ *
+ * @see [[https://tools.ietf.org/html/rfc2616 "Hypertext Transfer Protocol -- HTTP/1.1"]]
+ * @see [[https://tools.ietf.org/html/rfc7230 "Hypertext Transfer Protocol (HTTP/1.1): Message Syntax and Routing"]]
+ * section-3.3.2 on `Content-Length` for more information.
+ */
+private[finagle] class ModifyContentAndHeaderFieldFilter[Req <: Request] extends SimpleFilter[Req, Response] {
```
mosesn: How about `ComplianceFilter`, and that way we can continue to stick things in here if we decide to put in more changes?  For example, we could consider moving the content length setting in HttpServerDispatcher here.

finagle-http/src/main/scala/com/twitter/finagle/http/filter/ModifyContentAndHeaderFieldFilter.scala
```diff
@@ -0,0 +1,55 @@
...
+  import ModifyContentAndHeaderFieldFilter._
+
+  def apply(request: Req, service: Service[Req, Response]): Future[Response] = {
+    service(request) map { rep =>
+      if (mustNotIncludeContent(rep.status)) {
+        rep.clearContent()
```
mosesn: doing this silently makes it sort of difficult for our users.  we should probably either fail or log loudly (at least WARNING level).

finagle-http/src/main/scala/com/twitter/finagle/http/filter/ModifyContentAndHeaderFieldFilter.scala
```diff
@@ -0,0 +1,55 @@
...
+ *
+ * @see [[https://tools.ietf.org/html/rfc2616 "Hypertext Transfer Protocol -- HTTP/1.1"]]
+ * @see [[https://tools.ietf.org/html/rfc7230 "Hypertext Transfer Protocol (HTTP/1.1): Message Syntax and Routing"]]
+ * section-3.3.2 on `Content-Length` for more information.
+ */
+private[finagle] class ModifyContentAndHeaderFieldFilter[Req <: Request] extends SimpleFilter[Req, Response] {
+  import ModifyContentAndHeaderFieldFilter._
+
+  def apply(request: Req, service: Service[Req, Response]): Future[Response] = {
+    service(request) map { rep =>
```
mosesn: we prefer to avoid infix-style now, so if you could make this `service(request).map { rep =>` that would be appreciated 😀

finagle-http/src/main/scala/com/twitter/finagle/http/filter/ModifyContentAndHeaderFieldFilter.scala
```diff
@@ -0,0 +1,55 @@
...
+      if (mustNotIncludeContent(rep.status)) {
+        rep.clearContent()
+
+        if (rep.status == Status.NoContent)
+          rep.headerMap.remove(Fields.ContentLength)
```
mosesn: ditto here

finagle-http/src/main/scala/com/twitter/finagle/http/filter/ModifyContentAndHeaderFieldFilter.scala
```diff
@@ -0,0 +1,55 @@
...
+        if (rep.contentLength.isDefined)
+          rep.headerMap.set(Fields.ContentLength, "0")
+      }
+
+      rep
+    }
+  }
+}
+
+private[finagle] object ModifyContentAndHeaderFieldFilter extends HeadFilter[Request] {
```
mosesn: why extend HeadFilter here?

finagle-http/src/main/scala/com/twitter/finagle/http/filter/ModifyContentAndHeaderFieldFilter.scala
```diff
@@ -0,0 +1,55 @@
+package com.twitter.finagle.http.filter
+
+import com.twitter.finagle.http.{Fields, Request, Response, Status}
+import com.twitter.finagle._
+import com.twitter.util.Future
+
+/**
```
mosesn: could you add a test for this filter too?  end to end tests are rad, but for finding new bugs quickly, nothing beats a tightly scoped unit test =)

mosesn:
LGTM at a high level.  Can you also add a note to finagle/CHANGES?  I think this is worth calling out for our users.

finagle-http/src/main/scala/com/twitter/finagle/http/codec/HttpServerDispatcher.scala
monkey-mas: AFAIU, (Http)ServerDispatcher#[handle](https://github.com/twitter/finagle/blob/finagle-6.35.0/finagle-core/src/main/scala/com/twitter/finagle/dispatch/ServerDispatcher.scala#L63) is called after all the filters are applied? In `handle()`, we add a Content-Length header field with a value of 0 in [here](https://github.com/twitter/finagle/pull/521/files#diff-225f97996577fc88c90976a35c713e4cR102). If I understand finagle's behaviour correctly, I think we can't avoid putting the code there?
(but I have to admit this is not nice though :(

I might be wrong, please give me your idea :)

finagle-http/src/main/scala/com/twitter/finagle/http/filter/ModifyContentAndHeaderFieldFilter.scala
monkey-mas: Yeah, good name! :+1:
Fixed it!

finagle-http/src/main/scala/com/twitter/finagle/http/filter/ModifyContentAndHeaderFieldFilter.scala
monkey-mas: Sure. Though I've added a log here in a new commit, (as you suggested) it might be better to just `fail` here instead of modifying and sending a message? So that users can notice that there's something wrong happening?

finagle-http/src/main/scala/com/twitter/finagle/http/filter/ModifyContentAndHeaderFieldFilter.scala
monkey-mas: Done! :grinning:

finagle-http/src/main/scala/com/twitter/finagle/http/filter/ModifyContentAndHeaderFieldFilter.scala
monkey-mas: Oops, sorry my bad! Fixed it!

finagle-http/src/main/scala/com/twitter/finagle/http/filter/ModifyContentAndHeaderFieldFilter.scala
monkey-mas: Sure, I've added the corresponding test! :D

monkey-mas:
Thanks for the review! Except this [point](https://github.com/twitter/finagle/pull/521#discussion_r68779138), I think the fix commit should be fine. :) I'll squash additional commits into one after everything is ok.

Looking forward your feedback! :grinning:

finagle-http/src/main/scala/com/twitter/finagle/http/filter/ComplianceFilter.scala
```diff
@@ -0,0 +1,62 @@
...
+          rep.clearContent()
+          logger.warning(s"Response with a status code of ${rep.statusCode} must not have a body-message thus the content has been removed.")
+        }
+
+        if (rep.status == Status.NoContent && rep.contentLength.isDefined) {
+          rep.headerMap.remove(Fields.ContentLength)
+          logger.warning(s"Response with a status code of ${rep.statusCode} must not have a Content-Length header field thus the field has been removed.")
+        }
+
+        //Make sure a Content-Length header field is set with a value of 0 if it is explicitly sent for a 304 response
```
monkey-mas: Oops, I think I was wrong. As RFC 7230 section3.3.2 says, if we do want to send a Content-Length header field, we should return the Content-Length value of the cached content which has been returned to a client in a 200 response. I thought a value of Content-Length header field was always 0 as the content of a 304 response is always empty...
The following is the citation from the section;
`A server MAY send a Content-Length header field in a 304 (Not Modified) response to a conditional GET request (Section 4.1 of [RFC7232]); a server MUST NOT send Content-Length in such a response unless its field-value equals the decimal number of octets that would have been sent in the payload body of a 200 (OK) response to the same request.`

mosesn: Oof, this doesn't feel like it's easy to know.  I guess this is something that users must set?  I guess it's better to just not modify the content length in this case.

monkey-mas: > I guess this is something that users must set?

I think so if users really want to send the header field in a 304 response... I'm not sure how to _validate_ that they've chosen the _correct_ Content-Length value. As the RFC says, if the field-value doesn't equal the decimal number of octets that would have been sent in the payload body of a 200 (OK) response to the same request then we MUST NOT send Content-Length.

> I guess it's better to just not modify the content length in this case.

Yeah, not modifying is better I suppose so. Will fix it :)

I feel that things will be more complicated if we start to think about a Content-Length value of a 304 response as I've mentioned. Do you think we could stick to the original problems we had before and hopefully we can find a way to handle the new issue we've found now? What do you think?

monkey-mas: Fixed it :)

finagle-http/src/main/scala/com/twitter/finagle/http/codec/HttpServerDispatcher.scala
taylorleese: I have the same question as @mosesn. What's the verdict here?

monkey-mas: Hi @taylorleese  cc: @mosesn ,
I might be wrong but it seems like [handle(res: Reponse)](https://github.com/monkey-mas/finagle/blob/4002b80fe5b124459beaa8a3c8f364c3cb0762a2/finagle-http/src/main/scala/com/twitter/finagle/http/codec/HttpServerDispatcher.scala#L72) is called multiple times and it's called again after all [filters](https://github.com/monkey-mas/finagle/blob/4002b80fe5b124459beaa8a3c8f364c3cb0762a2/finagle-http/src/main/scala/com/twitter/finagle/Http.scala#L304-L308) are applied (s.t., even after `ComplianceFilter` is applied) as I mentioned before. I found this when I debug my commit with sample code like [this](https://gist.github.com/monkey-mas/eeade1dc021ab9f96a1dfa5f6198e09d).

I think this means a Content-Length is added to a 204 response for example. Suppose we're going to return a 204 response;
1)we _remove_ a Content-Length and message-body of a 204 response by using `ComplianceFilter`
2)but then `handle` is called and this [line](https://github.com/twitter/finagle/blob/develop/finagle-http/src/main/scala/com/twitter/finagle/http/codec/HttpServerDispatcher.scala#L96-L97) adds a Content-Length with a value of 0 to a 204 response as we've _already_ removed a Content-Length in 1) and `!rep.contentLength.isDefined` holds.

To achieve what we want to do, i.e., send a 204 without a Content-Length and message-body for example, we need to have this condition [here](https://github.com/monkey-mas/finagle/blob/4002b80fe5b124459beaa8a3c8f364c3cb0762a2/finagle-http/src/main/scala/com/twitter/finagle/http/codec/HttpServerDispatcher.scala#L101). The condition checks whether or not send a Content-Length.

This is why I think we still need `myHaveContent`. Anyway, I need your help to sort this out. Can you please help me? :)

mosesn:
@monkey-mas sorry for the delay, haven't had a chance to take a look this week.  but I haven't forgotten about you!

monkey-mas:
@mosesn haha, sure sure. :) Please have a look when you have time and let me know then!

finagle-http/src/main/scala/com/twitter/finagle/http/codec/HttpServerDispatcher.scala
bryce-anderson: Hi @monkey-mas, I'm new to this code so feel free to let me know if what follows is incorrect!
I don't think that `handle(res: Response)` is called multiple times: it's responsible for writing the message to the wire so if it's called multiple times we're in big trouble. I do think you're right in that its going to be called after all the filters so in its current form, its a problem.
It sounds like the problem is accommodating people sending a 204 with a body. In the case of people sending a 204 _without_ a body, the content-length header gets set to 0. Maybe what we can do is have `handle` look for a 204 response with a `content-length: 0` header, and strip those from the response? That will make 'legit' 204 responses legal by rfc 7230, and give your filter a way to signal that the header should be stripped.

To make this happen, the single `if` statement on 96 (old) might turn into:
```scala
rep.contentLength match {
  case Some(0) if rep.status == Status.NoContent =>
    rep.headers.remove(Fields.ContentLength)

  case Some(_) => // NOOP: don't perturb explicitly set content lengths other than 204 with length 0.  

  case None =>
    rep.contentLength = rep.content.length
}
```
There might be a cleaner way to implement that logic.

The filter would then set the `content-length` headers for Status(304), set it to 0 for Status(204), and strip away the bodies.

monkey-mas: Hi @bryce-anderson,
Thanks for your advice! (and sorry for my late reply!) Well I think I don't fully follow your strategy, can you give me more details please?

> I do think you're right in that its going to be called after all the filters so in its current form, its a problem.

Thanks for checking it, yeah that's what I expected :)

> It sounds like the problem is accommodating people sending a 204 with a body

Do you mean we don't provide support for removing a body-message of 204 and 304 responses if a body is explicitly sent by users?

> In the case of people sending a 204 without a body, the content-length header gets set to 0

Umm.. I think `Content-Length` header field is not set to 0 for a 204 response when there's no message-body. (But I think it's correct that the header field is added when we explicitly send a message-body, which is our problem here.)
More exactly, in the case of a 204 response with no body, `rep.contentLength` is `None` and `rep.content.length` is `0` I suppose.

Oh BTY, do you assume to use `ComplianceFilter` (or some filters) to handle the problem? It seems like you tried to solve the problem by just changing the original code at 96 in `HttpServerDispatcher`? (I thought so cos we don't have to check if a 204 response has `contentLength=Some(0)` after applying `ComplianceFitler`.)

bryce-anderson: TLDR: I've had some more time to think about it, and I'm not certain of an easy solution.

> Umm.. I think Content-Length header field is not set to 0 for a 204 response when there's no message-body. (But I think it's correct that the header field is added when we explicitly send a message-body, which is our problem here.)
> More exactly, in the case of a 204 response with no body, rep.contentLength is None and rep.content.length is 0 I suppose.

Sorry, its not clear to me if we agree. I believe at a `Response` with an empty body will get assigned a content length header of length 0. I haven't added tags to the `handle` function to check, but a raw `Response` should not be chunked and will not have a defined `contentLength`, so line 97 (on _master_, maybe this is the source of our confusion) should set the content-length to 0 on a 204 response without a body.

I don't think this can be solved _only_ in the compliance `Filter`, there needs to be some coupling because, as is, the content-length is going to be set for non-chunked bodies regardless of status code. I personally would hate to make the `HttpServerDispatcher` aware of a single `Filter`, that just doesn't seem fair to the other filters. 😄  That suggests to me (and we should get @mosesn's opinion in light of what we've found) that this isn't actually a job for a filter. This feels more like the things that are getting delegated to the netty pipeline such as [`RespondToExpectContinue`](https://github.com/twitter/finagle/blob/develop/finagle-http/src/main/scala/com/twitter/finagle/http/Codec.scala#L307).

monkey-mas: > Sorry, its not clear to me if we agree. I believe at a Response with an empty body will get assigned a content length header of length 0. I haven't added tags to the handle function to check, but a raw Response should not be chunked and will not have a defined contentLength, so line 97 (on master, maybe this is the source of our confusion) should set the content-length to 0 on a 204 response without a body.

Yes, I understand the code in the same way. Thanks for your explanation! :)  (I was thinking about `Content-Length` of `rep` where the code on line 97 is _not_ applied.)

> That suggests to me (and we should get @mosesn's opinion in light of what we've found) that this isn't actually a job for a filter. This feels more like the things that are getting delegated to the netty pipeline such as RespondToExpectContinue. 

Okay, so you mean we create a `ChannelHandler`, like `RespondToExpectContinue`, instead of using a `Filter` and give it a responsibility to handle 204 and 304 responses? Correct? Honestly, I'm not familiar with Netty's pipeline and not quite sure how it works in finagle. What sort of high-level solution (or actual implementation) do you have in your mind? (Can you please be more specific?)

bryce-anderson: I would definately wayt for @mosesn to chime in before going down the netty pipeline route. As a preview, I'll give my understanding of it. The code at https://github.com/twitter/finagle/blob/develop/finagle-http/src/main/scala/com/twitter/finagle/http/Codec.scala#L307 gives an example of how the `RespondToExpectContinue` handler works, but it is an `UpstreamHandler` (I believe that corresponds to `InboundHandler` in netty4). The `DownstreamHandler` works essentially the same but in the other direction. In my minds eye, the `DownstreamHandler` intercepts the `HttpResponse` and drops the body and requisite headers.

One thing that occurs to me is that streaming responses (chunked) don't appear to be addressed in this commit. The way streaming responses are rendered appears more complex and might make using the netty pipeline more tricky, if even possible. For example, In the case of a 304 or 204 I doubt that we want to use the `Reader`, and instead we would want to discard it. I don't think that is a possibility, at least in the place I pointed you to above.

bryce-anderson:
Okay @monkey-mas, I have discussed this with @mosesn and there may be another way to get this done in a bugwards compatible way. A lot of the functionality in the [HttpServerDispatcher.handle](https://github.com/twitter/finagle/blob/develop/finagle-http/src/main/scala/com/twitter/finagle/http/codec/HttpServerDispatcher.scala#L67) method deals with similar issues to what you're addressing, namely ensuring that the `Response` is in a conformant state. Perhaps this isn't something that the dispatcher should be concerned with and this could be moved from the HttpServerDispatcher to your `ComplianceFilter`. 
This is bugwards compatible in that those who need the older behavior could replace the filter you've made.

What do you think?

monkey-mas:
@bryce-anderson, cc: @mosesn 
Thanks for having a discussion! Yes, I think it sounds nice! :)

>  Perhaps this isn't something that the dispatcher should be concerned with and this could be moved from the HttpServerDispatcher to your ComplianceFilter.

Just to make sure, do you mean we leave `handle`'s core [functionality](https://github.com/twitter/finagle/blob/finagle-6.36.0/finagle-http/src/main/scala/com/twitter/finagle/http/codec/HttpServerDispatcher.scala#L78-L93) and move rest of the functionality to `ComplianceFilter`?

If so, I think we need to make two things clear:
  1. which functionality can we move to `ComplianceFilter`?
If I understand your thought correctly, we might move:
    1. [setKeepAlive](https://github.com/twitter/finagle/blob/finagle-6.36.0/finagle-http/src/main/scala/com/twitter/finagle/http/codec/HttpServerDispatcher.scala#L68)
    2. [others](https://github.com/twitter/finagle/blob/finagle-6.36.0/finagle-http/src/main/scala/com/twitter/finagle/http/codec/HttpServerDispatcher.scala#L69-L100): e.g., [removeContentLength](https://github.com/twitter/finagle/blob/finagle-6.36.0/finagle-http/src/main/scala/com/twitter/finagle/http/codec/HttpServerDispatcher.scala#L75), [setTransferEncoding](https://github.com/twitter/finagle/blob/finagle-6.36.0/finagle-http/src/main/scala/com/twitter/finagle/http/codec/HttpServerDispatcher.scala#L76) and [addContentLength](https://github.com/twitter/finagle/blob/finagle-6.36.0/finagle-http/src/main/scala/com/twitter/finagle/http/codec/HttpServerDispatcher.scala#L97)
    
        It could be not really good to move these [functionality](https://github.com/twitter/finagle/blob/finagle-6.36.0/finagle-http/src/main/scala/com/twitter/finagle/http/codec/HttpServerDispatcher.scala#L70-L76) if we take the comment into account though???

  2. can we _safely_ achieve it(moving functionality to the filter) and maintain the current finagle's behaviour?
I think the biggest reason why we've added the problematic logic to `handle` is to make sure we _apply_ the changes, such as removing a `Content-Length` and so on, *just* before we actually dispatch a response (after applying all filters and so forth)?
  Based on your idea, `ComplianceFilter` will take this responsibility? If we can't guarantee that `ComplianceFilter` is _used_ (just) before `handle` is called, then we might change finagle's behaviour? I might be wrong but I mean it's not _nice_ if there's a possibility that another filter or functionality, which adds a `Content-Length` for example, can be called after `ComplianceFilter` is used to filter a response? (or is this fine?)

What do you think?

mosesn:
@monkey-mas good catch with that comment.  maybe the first step is to fix netty's content compressor?  then we can make the rest of these changes more safely.

bryce-anderson:
@monkey-mas, I'm sorry we've been slow on this. We have started to move in the direction needed to incorporate your commit. See https://github.com/twitter/finagle/commit/fa8cef1c30e635366c763ae12de4133133a1f50b
If you're interested, perhaps we can get your work into the `ResponseConformanceFilter`.

bryce-anderson:
@monkey-mas, we'd love to have your contribution, you've put a lot of good work into this and I imagine you know the problem domain better than we do at this point (or else we would have this problem to begin with!).

In my minds eye, modifying validate is the right solution. @mosesn had some comments on being bugwards compatible. Personally, I don't see much of a case for being bugwards compatible here considering we are potentially committing a protocol violation by sending a body. I would suggest we log such events as errors considering it is certainly an error to generate a 204 or 304 with a body and should be fixed.

mosesn:
@bryce-anderson that seems reasonable to me, provided we advertise the change in CHANGES

CHANGES
```diff
@@ -62,6 +62,11 @@ Bug Fixes
      had a bug when toggled on. That toggle is no longer used and is superseded by
     `com.twitter.finagle.http.serverErrorsAsFailuresV2`. ``RB_ID=882151``
 
+  * finagle-http: To conform to the RFC, a message body is NO LONGER sent when 1xx, 204
+    and 304 responses are returned. Additionally, a Content-Length header field is NOT
+    sent for 1xx and 204 responses. Both rules are enforced even though users intentionally
```
bryce-anderson: Reads a little strange to me. Maybe ".. Both rules are enforced even _if_ users intentionally add body data or the header field for these responses." Maybe add something about logging violations as an error.

finagle-http/src/main/scala/com/twitter/finagle/http/codec/ResponseConformanceFilter.scala
```diff
@@ -36,13 +39,57 @@ private[codec] object ResponseConformanceFilter extends SimpleFilter[Request, Re
...
+        logger.error(
+          "Response with a status code of %d must not have a body-message thus the content has been removed.",
+          rep.statusCode)
+      }
+
+      rep.status match {
+        case Continue | SwitchingProtocols | Processing | NoContent if rep.contentLength.isDefined =>
+          rep.headerMap.remove(Fields.ContentLength)
+          logger.error(
+            "Response with a status code of %d must not have a Content-Length header field thus the field has been removed.",
```
mosesn: cols < 100 please

finagle-http/src/main/scala/com/twitter/finagle/http/codec/ResponseConformanceFilter.scala
```diff
@@ -36,13 +39,57 @@ private[codec] object ResponseConformanceFilter extends SimpleFilter[Request, Re
...
+   * "A server MAY send a Content-Length header field in a 304 (Not Modified) response to a conditional GET request
+   * (Section 4.1 of [RFC7232]); a server MUST NOT send Content-Length in such a response unless its field-value equals
+   * the decimal number of octets that would have been sent in the payload body of a 200 (OK) response to the same
+   * request."
+   */
+  private[this] def handleNoMessageResponse(rep: Response): Unit = {
+      if (rep.getContent() != ChannelBuffers.EMPTY_BUFFER) {
+        rep.clearContent()
+        logger.error(
+          "Response with a status code of %d must not have a body-message thus the content has been removed.",
```
mosesn: cols < 100 please

finagle-http/src/main/scala/com/twitter/finagle/http/codec/ResponseConformanceFilter.scala
```diff
@@ -36,13 +39,57 @@ private[codec] object ResponseConformanceFilter extends SimpleFilter[Request, Re
...
+   * "A server MUST NOT send a Content-Length header field in any response with a status code of 1xx (Informational)
+   * or 204 (No Content)."
+   *
+   * "A server MAY send a Content-Length header field in a 304 (Not Modified) response to a conditional GET request
+   * (Section 4.1 of [RFC7232]); a server MUST NOT send Content-Length in such a response unless its field-value equals
+   * the decimal number of octets that would have been sent in the payload body of a 200 (OK) response to the same
+   * request."
+   */
+  private[this] def handleNoMessageResponse(rep: Response): Unit = {
+      if (rep.getContent() != ChannelBuffers.EMPTY_BUFFER) {
```
bryce-anderson: We are trying to phase out usage of netty types in our HTTP model so could we do a length check instead? Something like `if (rep.length > 0) {` would be good. That has the added benefit of avoiding the ambiguity of equality in Java (eg, does `!=` mean reference equality here? I don't know...).

finagle-http/src/main/scala/com/twitter/finagle/http/codec/ResponseConformanceFilter.scala
```diff
@@ -36,13 +39,57 @@ private[codec] object ResponseConformanceFilter extends SimpleFilter[Request, Re
...
+        rep.clearContent()
+        logger.error(
```
bryce-anderson: ffti: might be useful to say how many bytes were in the body. Then again, maybe not. 😉

finagle-http/src/test/scala/com/twitter/finagle/http/AbstractEndToEndTest.scala
```diff
@@ -916,4 +918,138 @@ abstract class AbstractEndToEndTest extends FunSuite
...
+      val res = await(client(Request(Method.Get, "/")))
+      assert(res.status == resStatus)
+      assert(!res.httpMessage.isChunked)
+      assert(res.httpMessage.getContent == ChannelBuffers.EMPTY_BUFFER)
+      assert(res.contentLength.isEmpty)
+      await(client.close())
+      await(server.close())
+    }
+
+    List(Status.Continue, /*Status.SwitchingProtocols,*/ Status.Processing, Status.NoContent).foreach {
```
mosesn: please de-comment SwitchingProtocols

finagle-http/src/test/scala/com/twitter/finagle/http/AbstractEndToEndTest.scala
```diff
@@ -916,4 +918,138 @@ abstract class AbstractEndToEndTest extends FunSuite
...
+      val res = await(client(Request(Method.Get, "/")))
+      assert(res.status == resStatus)
+      assert(!res.httpMessage.isChunked)
+      assert(res.httpMessage.getContent == ChannelBuffers.EMPTY_BUFFER)
+      assert(res.contentLength.isEmpty)
+      await(client.close())
+      await(server.close())
+    }
+
+    List(Status.Continue, /*Status.SwitchingProtocols,*/ Status.Processing, Status.NoContent, Status.NotModified).foreach {
```
mosesn: please de-comment SwitchingProtocols

finagle-http/src/main/scala/com/twitter/finagle/http/codec/ResponseConformanceFilter.scala
```diff
@@ -36,13 +39,57 @@ private[codec] object ResponseConformanceFilter extends SimpleFilter[Request, Re
...
+          "Response with a status code of %d must not have a body-message thus the content has been removed.",
+          rep.statusCode)
+      }
+
+      rep.status match {
+        case Continue | SwitchingProtocols | Processing | NoContent if rep.contentLength.isDefined =>
```
bryce-anderson: This doesn't cover the case of 1xx status codes that are not 100-Continue. Could we be inclusive of the whole range?

finagle-http/src/test/scala/com/twitter/finagle/http/AbstractEndToEndTest.scala
```diff
@@ -916,4 +918,138 @@ abstract class AbstractEndToEndTest extends FunSuite
...
```
mosesn: please de-comment switching protocols

finagle-http/src/test/scala/com/twitter/finagle/http/AbstractEndToEndTest.scala
```diff
@@ -916,4 +918,138 @@ abstract class AbstractEndToEndTest extends FunSuite
     await(server.close())
     await(client.close())
   }
+
+  testIfImplemented(NoBodyMessage)(
```
mosesn: I have a preference for small tests, do you mind structuring this differently so that the `test` clause is *inside* of the loop?  We can get around the duplicate names by string interpolating into the name of the test.

finagle-http/src/main/scala/com/twitter/finagle/http/codec/ResponseConformanceFilter.scala
```diff
@@ -36,13 +39,57 @@ private[codec] object ResponseConformanceFilter extends SimpleFilter[Request, Re
...
+          rep.headerMap.remove(Fields.ContentLength)
+          logger.error(
+            "Response with a status code of %d must not have a Content-Length header field thus the field has been removed.",
+            rep.statusCode)
+        case _ =>
+      }
+  }
+
+  private def mustNotIncludeMessageBody(status: Status): Boolean = status match {
```
bryce-anderson: This has the same 1xx problem as above.
Maybe this is worth it, maybe its not, but it seems to me that this pattern match could be reused with a special case for `NotModified`. Eg
```scala
private def mustNotHaveLengthHeader(status: Status) = 
  status != NotModified && mostNotIncludeMessageBody(status)
```

Maybe that is getting too convoluted, so use discretion with that suggestion.

finagle-http/src/test/scala/com/twitter/finagle/http/AbstractEndToEndTest.scala
```diff
@@ -916,4 +918,138 @@ abstract class AbstractEndToEndTest extends FunSuite
...
+      await(server.close())
+    }
+
+    List(Status.Continue, /*Status.SwitchingProtocols,*/ Status.Processing, Status.NoContent).foreach {
+      check(_)
+    }
+  }
+
+
+  testIfImplemented(NoBodyMessage)(
```
mosesn: does this need the NoBodyMessage feature?  or can we make this a regular test?

finagle-http/src/test/scala/com/twitter/finagle/http/AbstractEndToEndTest.scala
```diff
@@ -916,4 +918,138 @@ abstract class AbstractEndToEndTest extends FunSuite
...
+    "response with status code 304 must not have a message body *BUT* Content-Length header field " +
```
mosesn: cols <= 100 please

finagle-http/src/test/scala/com/twitter/finagle/http/codec/ResponseConformanceFilterTest.scala
```diff
@@ -78,6 +79,69 @@ class ResponseConformanceFilterTest extends FunSuite {
     intercept[ReaderDiscarded] { Await.result(res.writer.write(Buf.Empty), 5.seconds) }
   }
 
+  test("response with status code {1xx, 204 and 304} must not have a message body nor Content-Length header field") {
```
mosesn: cols <= 100 please

finagle-http/src/test/scala/com/twitter/finagle/http/codec/ResponseConformanceFilterTest.scala
```diff
@@ -78,6 +79,69 @@ class ResponseConformanceFilterTest extends FunSuite {
...
+      assert(response.status == status)
+      assert(response.getContent.readableBytes() == 0)
+      assert(!response.isChunked)
+      assert(response.headers().get(Fields.ContentLength) == null)
+    }
+
+    List(Continue, SwitchingProtocols, Processing, NoContent, NotModified).foreach(validate(_))
+  }
+
+  test("response with status code {1xx, 204 and 304} must not have a message body nor Content-Length header field when non-empty body is returned") {
```
mosesn: cols <= 100 please

finagle-http/src/test/scala/com/twitter/finagle/http/codec/ResponseConformanceFilterTest.scala
```diff
@@ -78,6 +79,69 @@ class ResponseConformanceFilterTest extends FunSuite {
...
+      assert(response.status == status)
+      assert(response.getContent().readableBytes() == 0)
+      assert(!response.isChunked)
+      assert(response.headers().get(Fields.ContentLength) == null)
+    }
+
+    List(Continue, SwitchingProtocols, Processing, NoContent, NotModified).foreach(validate(_))
+  }
+
+  test("response with status code {1xx and 204} must not have a message body nor Content-Length header field when non-empty body with explicit Content-Length is returned") {
```
mosesn: cols <= 100 please

finagle-http/src/test/scala/com/twitter/finagle/http/codec/ResponseConformanceFilterTest.scala
```diff
@@ -78,6 +79,69 @@ class ResponseConformanceFilterTest extends FunSuite {
...
+      assert(response.status == status)
+      assert(response.getContent().readableBytes() == 0)
+      assert(!response.isChunked)
+      assert(response.headers().get(Fields.ContentLength) == null)
+    }
+
+    List(Continue, SwitchingProtocols, Processing, NoContent).foreach(validate(_))
+  }
+
+  test("response with status code 304 must not have a message body *BUT* Content-Length header field when non-empty body with explicit Content-Length is returned") {
```
mosesn: cols <= 100 please

finagle-http/src/test/scala/com/twitter/finagle/http/codec/ResponseConformanceFilterTest.scala
```diff
@@ -78,6 +79,69 @@ class ResponseConformanceFilterTest extends FunSuite {
...
 
+  test("response with status code {1xx, 204 and 304} must not have a message body nor Content-Length header field") {
+    def validate(status: Status) = {
+      val res = Response(Version.Http11, status)
+      val response = fetchResponse(res)
+
+      assert(response.status == status)
+      assert(response.getContent.readableBytes() == 0)
+      assert(!response.isChunked)
+      assert(response.headers().get(Fields.ContentLength) == null)
```
bryce-anderson: Try to use `.headerMap` to remove netty types from the testing. It's also nice not to look for `null` in scala code. There are a few more instances of this below as well.

finagle-http/src/test/scala/com/twitter/finagle/http/codec/ResponseConformanceFilterTest.scala
```diff
@@ -78,6 +79,69 @@ class ResponseConformanceFilterTest extends FunSuite {
     intercept[ReaderDiscarded] { Await.result(res.writer.write(Buf.Empty), 5.seconds) }
   }
 
+  test("response with status code {1xx, 204 and 304} must not have a message body nor Content-Length header field") {
+    def validate(status: Status) = {
+      val res = Response(Version.Http11, status)
+      val response = fetchResponse(res)
+
+      assert(response.status == status)
+      assert(response.getContent.readableBytes() == 0)
```
bryce-anderson: This could be `response.length` which is shorter and avoids netty types. A few more instances of this below.

finagle-http/src/main/scala/com/twitter/finagle/http/codec/ResponseConformanceFilter.scala
```diff
@@ -36,13 +39,57 @@ private[codec] object ResponseConformanceFilter extends SimpleFilter[Request, Re
...
+   * RFC7230 section-3.3.2: (https://tools.ietf.org/html/rfc7230#section-3.3.2)
+   * "A server MUST NOT send a Content-Length header field in any response with a status code of 1xx (Informational)
+   * or 204 (No Content)."
+   *
+   * "A server MAY send a Content-Length header field in a 304 (Not Modified) response to a conditional GET request
+   * (Section 4.1 of [RFC7232]); a server MUST NOT send Content-Length in such a response unless its field-value equals
+   * the decimal number of octets that would have been sent in the payload body of a 200 (OK) response to the same
+   * request."
+   */
+  private[this] def handleNoMessageResponse(rep: Response): Unit = {
```
bryce-anderson: I think we need a plan for dealing with chunked responses as well.

finagle-http2/src/test/scala/com/twitter/finagle/http2/ClientFailUpgradeTest.scala
```diff
@@ -25,7 +25,8 @@ class ClientFailUpgradeTest extends AbstractHttp1EndToEndTest {
     TooLongStream,
     Streaming,
     CloseStream,
-    StreamFixed
+    StreamFixed,
+    NoBodyMessage
```
mosesn: my suspicion is that this fails for ClientFailUpgradeTest because the client starts having issued an upgrade message, so it's expecting a 101 in response.  When it gets a 101, it expects upgrade headers, and netty will throw an exception if you don't have them.  I don't know how the finagle client behaves in that scenario.  Do the tests work if you remove the 101 test from ClientFailUpgradeTest?  Could we restructure the tests to use "featureImplemented" to decide whether to include 101 or not instead of just turning them off for CFUT?

finagle-http/src/test/scala/com/twitter/finagle/http/AbstractEndToEndTest.scala
```diff
@@ -916,4 +918,138 @@ abstract class AbstractEndToEndTest extends FunSuite
...
+      await(server.close())
+    }
+
+    List(Status.Continue, /*Status.SwitchingProtocols,*/ Status.Processing, Status.NoContent, Status.NotModified).foreach {
+      check(_)
+    }
+  }
+
+  testIfImplemented(NoBodyMessage)(
+    "response with status code {1xx and 204} must not have a message body nor Content-Length header field" +
```
mosesn: cols <= 100 please

finagle-http/src/test/scala/com/twitter/finagle/http/codec/ResponseConformanceFilterTest.scala
```diff
@@ -78,6 +79,69 @@ class ResponseConformanceFilterTest extends FunSuite {
     intercept[ReaderDiscarded] { Await.result(res.writer.write(Buf.Empty), 5.seconds) }
   }
 
+  test("response with status code {1xx, 204 and 304} must not have a message body nor Content-Length header field") {
+    def validate(status: Status) = {
```
mosesn: I have the same feedback as before about the structure of these tests.

finagle-http/src/main/scala/com/twitter/finagle/http/codec/ResponseConformanceFilter.scala
monkey-mas: Do you mean we need a plan to conform to [RFC7230#section-3.3.1](https://tools.ietf.org/html/rfc7230#section-3.3.1) probably?
It says:
```
   A server MUST NOT send a Transfer-Encoding header field in any
   response with a status code of 1xx (Informational) or 204 (No
   Content).  A server MUST NOT send a Transfer-Encoding header field in
   any 2xx (Successful) response to a CONNECT request (Section 4.3.6 of
   [RFC7231]).
```

Do you think it's better for us to handle the issue in this PR or another PR?

Well, my first thought to solve the problem is like this:
```
private[this] def handleNoMessageResponse(rep: Response): Unit = {
  // remove message body if necessary

  // remove Content-Length and Transfer-Encoding if necessary
  if (rep.status != 304 && mustNotIncludeMessageBody(rep.status)) {
  	if (rep.contentLength.isDefined) {
      rep.headerMap.remove(Fields.ContentLength)
      log.error(...)
  	}
  	if (rep.isChunked) {
      rep.headerMap.remove(Fields.TransferEncoding)
      log.error(...)
  	}
  }
}
```

finagle-http/src/main/scala/com/twitter/finagle/http/codec/ResponseConformanceFilter.scala
monkey-mas: Okay, I think you mean we need to validate `100 <= status code < 200` for 1xx condition check, correct? I thought it'd be enough to check Continue, SwitchingProtocols and Processing but (if you meant it) yes I think it's reasonable :)

finagle-http/src/main/scala/com/twitter/finagle/http/codec/ResponseConformanceFilter.scala
monkey-mas: Good point, thanks! I fixed the condition to check 1xx codes as well. WDYT?

finagle-http/src/test/scala/com/twitter/finagle/http/AbstractEndToEndTest.scala
monkey-mas: Yeah, you're right. It's nice to have small tests! Hope my modification is what you suggested :)

finagle-http/src/test/scala/com/twitter/finagle/http/AbstractEndToEndTest.scala
monkey-mas: It seems like I misunderstood the concept of `Feature`. I defined `NoBodyMessage` to _pass_ some tests in finagle-http2 for example (by ignoring test cases?). What do you think I should've done here? :eyes:

finagle-http2/src/test/scala/com/twitter/finagle/http2/ClientFailUpgradeTest.scala
monkey-mas: Your suspicion sounds reasonable to me. In that case, we might have _timeout_ error or exception?

> Do the tests work if you remove the 101 test from ClientFailUpgradeTest?

Yup, I think it works without 101 test. Currently, 101 test (finagle-http) also fails for `EndToEndTest` with the following error:
`Some(0) was not empty (AbstractEndToEndTest.scala:949)`
[AbstractEndToEndTest.scala:949](https://github.com/twitter/finagle/pull/521/commits/057ed65fd51669e2c36a0af12300cb7f024c6094#diff-b884af71723a791bed9b01d0b4ece1e9R949)
It looks like Content-Length has the value of 0, which shouldn't happen like we can see from `ResponseConformanceFilterTest`...

@bryce-anderson What do you think about this?

>  Could we restructure the tests to use "featureImplemented" to decide whether to include 101 or not instead of just turning them off for CFUT?

Could you show me some sample code of how to use `featureImplemented` please? It seems like there's no code using it in finagle? :eyes:

mosesn:
@monkey-mas sorry, we dropped the ball on this!  I'm going to see if I can get this over the goalpost and merge it in.  Will keep you posted!

mosesn:
this is failing in really strange ways.  I think the messiness of finagle's HTTP compliance stuff is catching up to us.  I'm going to keep on trying to make it work, but might have to give up if it takes too much time.  If you have some time to look at this again, I'd be happy to send you what I've done so far.

monkey-mas:
@mosesn 
My apologies. Sorry for not handling this PR for a long time.

> this is failing in really strange ways. I think the messiness of finagle's HTTP compliance stuff is catching up to us. I'm going to keep on trying to make it work, but might have to give up if it takes too much time. If you have some time to look at this again, I'd be happy to send you what I've done so far.

Honestly, I've got no idea why it fails. You have some clue, don't you?

I think I can have some time to fix it. Can you please give me some information you've collected so far?
If things go wrong (or if you'd think it's not good to keep doing this stuff), let's close this PR. (Well, that's my fault for keeping the PR open... Sorry again! :))

mosesn:
@monkey-mas I ended up undoing most of my changes since they didn't seem to be helping.  I'm not sure where the issue is, but I've merged against master and made a PR: https://github.com/monkey-mas/finagle/pull/1

monkey-mas:
@mosesn Thanks so much! Let me have a look at it :)

monkey-mas:
@mosesn @bryce-anderson 
Okay I kinda understand our situation. I might be wrong but give me your thoughts please :)

### Problem
The current (wierd) problem we have is our response gives us `Content-Length: 0` even if we remove it using [ResponseConformanceFilter](https://github.com/monkey-mas/finagle/blob/2dfc051e249b538050008325624f82eb1432b5ad/finagle-http/src/main/scala/com/twitter/finagle/http/codec/ResponseConformanceFilter.scala#L40-L41)
Actually end-to-end test fails like:
```
[info] - response with status code 204 must not have a message body nor Content-Length header field when non-empty body with explicit Content-Length is returned *** FAILED ***
[info]   Some(0) was not empty (AbstractEndToEndTest.scala:1472)
```

Based on the fact that the unit test `ResponseConformanceFilterTest` passes, I think the filter works properly.

### What I've observed so far

I've done some print-debugging (& stack tracing) and found our code works as follows:

1. [Body & Content-Length are set in test code](https://github.com/monkey-mas/finagle/blob/2dfc051e249b538050008325624f82eb1432b5ad/finagle-http/src/test/scala/com/twitter/finagle/http/AbstractEndToEndTest.scala#L1444-L1445)
  [Netty3HeaderMap#set](https://github.com/twitter/finagle/blob/finagle-6.44.0/finagle-base-http/src/main/scala/com/twitter/finagle/http/Netty3HeaderMap.scala#L47-L50)(Content-Length , 9) is called

2. [Body & Content-Length are both removed](https://github.com/monkey-mas/finagle/blob/2dfc051e249b538050008325624f82eb1432b5ad/finagle-http/src/main/scala/com/twitter/finagle/http/codec/ResponseConformanceFilter.scala#L71-L90) using `ResponseConformanceFilter`

3. [Bijections#fullResponseToFinagle(res)](https://github.com/twitter/finagle/blob/develop/finagle-netty4-http/src/main/scala/com/twitter/finagle/netty4/http/Bijections.scala#L90-L104) and [Netty3HeaderMap#add](https://github.com/twitter/finagle/blob/finagle-6.44.0/finagle-base-http/src/main/scala/com/twitter/finagle/http/Netty3HeaderMap.scala#L52-L55)(content-length, 0) are called where `res` is
  ```
  DefaultFullHttpResponse(decodeResult: success, version: HTTP/1.1, content: EmptyByteBufBE)
  HTTP/1.1 204 No Content
  content-length: 0)
  ```

Additionally, I roughly checked if our code adds the header somewhere, by using `git grep 'contentLength'` for example, but it seems not.

Based on this, I _think_ `Content-Length: 0` is set when `fullResponseToFinagle(...)` is called, which ends up with `contentLength == Some(0)` thus test fails. It could be possible that Netty adds `Content-Length: 0` and we have the header in `NettyHttp.FullHttpResponse` when calling `fullResponseToFinagle(...)`?


### (Naive) Solution
With that said I'm still not 100% for sure how and when `Content-Length: 0` is set. If my guess is correct, the header is added after filters are applied to a response. So we can't avoid handling the issue within `Bijections` like below?

```diff
@@ -88,15 +90,21 @@ private[finagle] object Bijections {
     }
 
     def fullResponseToFinagle(rep: NettyHttp.FullHttpResponse): FinagleHttp.Response = {
       val payload = ByteBufAsBuf(rep.content)
+      val status = rep.status
 
       val resp = FinagleHttp.Response(
         versionToFinagle(rep.protocolVersion),
-        statusToFinagle(rep.status),
+        statusToFinagle(status),
         BufReader(payload)
       )
 
       resp.setChunked(false)
+      if (status.codeClass() == HttpStatusClass.INFORMATIONAL || status == HttpResponseStatus.NO_CONTENT) {
+        rep.headers().remove("Content-Length")
+      }
       writeNettyHeadersToFinagle(rep.headers, resp.headerMap)
       resp.content = payload
```

mosesn:
Good sleuthing!  I would love to know why content length is being set to 0, but I'm OK with pulling this in before we know.  @bryce-anderson what do you think?

monkey-mas:
Finally I determined the cause of `Content-Length: 0`. I believe it's set in [HttpObjectAggregator#finishAggregation(...)](https://github.com/netty/netty/blob/netty-4.1.11.Final/codec-http/src/main/java/io/netty/handler/codec/http/HttpObjectAggregator.java#L214-L227) by netty.

The following is the rough sketch of (reversed) execution paths to encounter `Content-Length: 0`.
```
3. Bijections#fullResponseToFinagle(...)

2. UnpoolHttpHandler#channelRead(...)

1. HttpObjectAggregator#finishAggregation(...)
```

As I've mentioned before, the header's already set to 0 when `Bijections#fullResponseToFinagle(...)` is executed. Before this, [UnpoolHttpHandler#channelRead(ctx: ChannelHandlerContext, msg: Any)](https://github.com/twitter/finagle/blob/develop/finagle-netty4-http/src/main/scala/com/twitter/finagle/netty4/http/handler/UnpoolHttpHandler.scala#L21) is called where `msg` is something like

```
HttpObjectAggregator$AggregatedFullHttpResponse(decodeResult: success, version: HTTP/1.1, content: CompositeByteBuf(ridx: 0, widx: 0, cap: 0, components=0))
HTTP/1.1 204 No Content
content-length: 0
```

As `msg` indicates, we've already got `content-length: 0`. This header is actually appended when [HttpObjectAggregator#finishAggregation(FullHttpMessage aggregated)](https://github.com/netty/netty/blob/netty-4.1.11.Final/codec-http/src/main/java/io/netty/handler/codec/http/HttpObjectAggregator.java#L214-L227) _finishes_ execution. Because we've already removed `Content-Length` header using `ResponseConformanceFilter`, `aggregated(AggregatedFullHttpResponse)` doesn't have the corresponding header at the beginning of the method call. Actually `aggregated.message.headers` is `DefaultHeadersImpl[]` (empty). This leads to our unexpected result, i.e., `Content-Length` is [set to String.valueOf(aggregated.content().readableBytes()))](https://github.com/netty/netty/blob/netty-4.1.11.Final/codec-http/src/main/java/io/netty/handler/codec/http/HttpObjectAggregator.java#L223-L225) where
```
aggregated.content() == CompositeByteBuf(ridx: 0, widx: 0, cap: 0, components=0)
aggregated.content().readableBytes == (this.writerIndex - this.readerIndex) == 0
```
ref. [readableBytes](https://github.com/netty/netty/blob/netty-4.1.11.Final/buffer/src/main/java/io/netty/buffer/ByteBuf.java#L409)

Therefore, `aggregated` has `Content-Length: 0` here.

bryce-anderson:
I'll pull it in and take another look! Sounds like this is further down in the pipeline a bit, so this is more interesting than we'd hoped. :) Excellent sleuthing @monkey-mas.

bryce-anderson:
@monkey-mas, I've looked through your PR and wanted to get your take: it does look like Netty is responsible for the test failures as you've laid out above. How do you want to proceed? In my mind's eye, Netty should probably be fixed, so if you're interested in contributing to that project I think you've got a nice contribution on deck. In the meantime, we could disable the end-to-end tests that are failing due to the client issue.

What do you think?

monkey-mas:
@bryce-anderson Sorry for my late reply. 

> In the meantime, we could disable the end-to-end tests that are failing due to the client issue.

Do you mean we can just _merge_ functionality of `handleNoMessageResponse ` to `ResponseConformanceFilter` without end-to-end tests? Or should we _lock_ this PR until Netty solves the problem? Either way sounds good to me, let me know your opinion :)

> In my mind's eye, Netty should probably be fixed, so if you're interested in contributing to that project I think you've got a nice contribution on deck.

Cool! I'll open a PR related to this issue though I'm not quite sure if Netty community likes the change to follow RFC7230. :(
Anyway I'll keep you updated! :)

bryce-anderson:
@monkey-mas, no worries! I'm in the process of merging this in with the end-to-end tests disabled: this works as expected when running a simple telnet session.

I feel the Netty folks are generally quite receptive to patches that make sure _they_ don't break things, which this seems to be an instance of. They already have similar logic laying around to deal with cases where a body shouldn't exist, for example [here](https://github.com/netty/netty/blob/4.1/codec-http/src/main/java/io/netty/handler/codec/http/HttpContentEncoder.java#L209). If you ask me, that code would be better off renamed and in the HttpUtil object.

bryce-anderson:
@monkey-mas, after long last, this was merged as commit 1e4252e. Thanks a lot for both your patience and good work! Keep us updated on the status of any Netty PR you submit to address the chunk aggregation issue.

monkey-mas:
@bryce-anderson @mosesn 
Actually thanks a lot for your patience and great feedback! It wouldn't be possible without your support :)

Okay, I'll get started with the aggregation issue. The easiest solution could be to modify the logic of [finishAggregation(...)](https://github.com/netty/netty/blob/4.1/codec-http/src/main/java/io/netty/handler/codec/http/HttpObjectAggregator.java#L214-L227) as this should be the last point before an actual response is returned to a client. 
Moreover, some methods can be moved to HttpUtil and refactored like you said. Anyway let's discuss it at Netty's github :)

bryce-anderson:
@monkey-mas 👍
Please feel free to ping me in any related Netty issue/PR you file.

vkostyukov:
Hey @monkey-mas! We had to revert this temporary (see b5cbb3e2ef23c82f30dbff6479c834048992e6fe) for some internal reasons. We'll put it back as soon as we can. Stay tuned and thanks again - this is a solid piece of work!

monkey-mas:
Oops sorry to hear that :( I hope all is well at your end.
(Anyway thanks for kindly letting me know!)

vkostyukov:
Hi @monkey-mas! I'm trying to put your work back where it belongs. Will update this thread when it's merged in.

vkostyukov:
It's merged (see f8208cd202fb98b2b071d53e8882457c4b4af208)!

PR #525, opened 2016-07-11
finagle-netty4: Wait for client channels to go active when connecting

olix0r:
Problem

Netty4Transporter currently satisfies its future as soon as a connection is established, causing Finagle to write messages to the transport immediately. Some protocols require transport initialization after connection (see, for example, RFC7540 §3.5).

Solution

Do not yield Transports from the Transporter until the channel enters the active, allowing the netty pipeline to perform additional negotiation.

### DIFF
--- a/finagle-netty4/src/main/scala/com/twitter/finagle/netty4/Netty4Transporter.scala
+++ b/finagle-netty4/src/main/scala/com/twitter/finagle/netty4/Netty4Transporter.scala
@@ -80,7 +80,23 @@ private[finagle] object Netty4Transporter {
             case e: UnresolvedAddressException => e
             case NonFatal(e) => Failure.rejected(e)
           })
-          else transportP.setValue(new ChannelTransport[In, Out](channelF.channel()))
+          else {
+            // A connection has been established. Wait until the
+            // channel is active before returning a transport.
+            val ch = channelF.channel
+            ch.pipeline.addLast(new ChannelInboundHandlerAdapter {
+              override def channelActive(ctx: ChannelHandlerContext): Unit = {
+                transportP.setValue(new ChannelTransport[In, Out](ch))
+                ch.pipeline.remove(this)
+              }
+
+            // The channel never went active...
+            override def exceptionCaught(ctx: ChannelHandlerContext, exn: Throwable): Unit = {
+                transportP.setException(exn)
+                val _ = ctx.fireExceptionCaught(exn)
+              }
+            })
+          }
         }
       })
 


olix0r:
I'm unsure how to unit test this... open to ideas.

mosesn:
@olix0r I don't think we want to do this.  The problem is that the netty state machine doesn't require that writes happen after the entire connection setup has started, so it's not a safe assumption to make.  Given that, it simplifies matters not to encode that assumption into finagle, since it's valid for netty handlers not to assume this, and we want to be able to use netty handlers.

olix0r:
Fair enough -- I suppose the alternative is to have my pipeline block the connect future from firing until the channel goes active?  Or, alternatively, buffer outbound messages until the state is active? Any suggestions?

mosesn:
Yeah, we buffer outbound messages until the state is active.  See [BufferingChannelOutboundHandler][0].

[0]: https://github.com/twitter/finagle/blob/develop/finagle-netty4/src/main/scala/com/twitter/finagle/netty4/channel/BufferingChannelOutboundHandler.scala

vkostyukov:
*I think* If transport/pipeline requires to perform upgrade/handshake/whatever, it should be responsible for buffering writes on its own. For example, Netty's `SslHandler` works this way. We adopted that idea (with some changes) in HTTP proxy handshake (see `HttpConnectHandler`) / SOCKS proxy handshake / SSL handshake. Basically, we buffer writes from the handlers before and also delays the connect promise so the Finagle transport is not established. The reasons we do that are:

- want to make sure, for example, SSL handshake exception fails the connect promise, not the request one;
- want to make sure we don't send much traffic to a client that might potentially fail the session/handshake;

olix0r:
@vkostyukov thanks for the explanation. I'm unclear on one thing, though:

Suppose we use a Netty4Transporter that returns a future that is satisfied on connect; and then we use a buffering channel adapter that prevents writes on the transport from succeeding until some later state is initialized.  So, normally the following happens:

1. Connection established.
2. Write req0, buffering.
3. Session established.
4. buffered writes are flushed.

If step 3 fails, we fail all of the _requests_ that were buffered.  I assume these are considered requeuable? It's not immediately obvious to me where this is handled...

It seems that by changing a Transporter to not return a Transport until the session is active, this problem could be avoided as session establishment errors would be treated as connection failures...

mosesn:
@olix0r yeah, it's a tradeoff between clarifying where failures happen in finagle and following typical netty paradigms.  maybe we could write a generic test suite for netty handlers that we can use to demonstrate we're resilient against this kind of failure, and then add this patch back on?  idk, this is the tradeoff we've chosen right now, but it's not necessarily the right one (or immutable).

olix0r:
Cool. Thanks for helping me understand this.

vkostyukov:
Just wanted to make sure I'm not missing something here (quite likely I do).

@olix0r The scenario you're describing is exactly what [ConnectPromiseDelayListenners](https://github.com/twitter/finagle/blob/develop/finagle-netty4/src/main/scala/com/twitter/finagle/netty4/channel/ConnectPromiseDelayListeners.scala) is supposed to solve - we do not satisfy transport promise until we've performed all the handshakes (established the session). This works today for HTTP/Socks proxy and TLS/SSL.

As I understand we're talking about http2 upgrade/preface? My take is if we need to make sure http2 session is established before accepting traffic, we should do the similar thing we do for SSL/HTTP & SOCKS proxies - in an ad-hoc way delay the connect promise.

olix0r:
@vkostyukov thanks for the pointer! this is exactly what I was looking for.

It appears that `CancelledConnectionException` is raised on buffered requests -- are these requests considered to be retryable? Does the RequeueFilter handle failed connections like this?

vkostyukov:
@olix0r The idea is that if we fail those writes we also fail the connect promise, which means a channel was never active. We only buffer writes from the channel handlers before/above - not from a Finagle client. That said, all those writers are somehow part of the session handshake (initiated by a pipeline). Failing them means failing the connect promise, which will be converted into a `Failure.rejected` and will be retried by the `Retries` module.

vkostyukov:
Technically, we only need buffering to protect ourselves from other handlers that can write on `channelActive` or `handlerAdded` (eg: `SslHndler`). We know for sure that Finagle won't write until the connect promise is satisfied so we never buffer users requests.

PR #527, opened 2016-07-14
Add transition to ProbeClosed in onServiceAcquisitionFailure

### LINKED ISSUE #524: Inactive memcached node remains stuck in `ProbeOpen` state, when EjectFailedHost is configured
Inactive memcached node remains stuck in `ProbeOpen` state, when EjectFailedHost is configured.

### Expected behavior

* I am using finagle 6.35.0, I have Memcached client configured as follows:
```
Memcached.client
  .configured(FailureAccrualFactory.Param(1, () => Duration.fromSeconds(10)))
  .configured(Memcached.param.EjectFailedHost(true))
  .newRichClient("localhost:11211,localhost:11212")))
```
* Every 30 seconds, I am performing get and/or set operation with 10 different Key-Value pairs 
* Node `localhost:11212` is down the entire time (because NO server is listening at that port)
* I expect that since `EjectFailedHost` and strict `FailureAccrualPolicy` is configured, every 30 seconds, all 10 operations succeed - by going to memcached node `localhost:11211`
* I expect that memcached node `localhost:11212` stays `Dead` and out of the ring whole time.

### Actual behavior

* On first attempt say time-0, all the 10 operations succeed, all of them goto active node `localhost:11211`. Other node `localhost:11212` is marked as `Dead` (the log message print confirms this). So this is good, as per my expectation
* On 2nd attempt and successive attempts, some (say 4) operations out of 10 operations try to connect to inactive node i.e. `localhost:11212` and fail.  
* I tried configuring `.configured(FailFastFactory.FailFast(true))`, but that didn't help either

* Create Memcached client with above mentioned code (for 2 memcached nodes, 1 active, other inactive) 
* every 30 seconds try 10 set operations with different key-value pairs
* On 2nd & susccessive attempts, some operations should fail

### Debugging Observations 

* The inactive node gets ejected after 1st failure (i.e. 1st attempt)
* After markDeadFor timeout (10 seconds in this case), on startProbing (i.e. state=probeOpen), the node is added back to ring. 
* When service/probe is invoked on this node, it tries to process [underlying(conn).map](https://github.com/twitter/finagle/blob/c80cccfc575494275cfb9d9fbb6a6800b1e7409d/finagle-core/src/main/scala/com/twitter/finagle/service/FailureAccrualFactory.scala#L367) and 
hits the following exception:
```
com.twitter.finagle.ChannelWriteException: java.net.ConnectException: Connection refused: localhost/0:0:0:0:0:0:0:1:11213. Remote Info: Upstream Address: /127.0.0.1:54459, Upstream Client Id: Not Available, Downstream Address: localhost/0:0:0:0:0:0:0:1:11213, Downstream Client Id: fixedinet!localhost:11213, Trace Id: 88c6d00fd9f95341.8f24651fa01698bc<:f700f541942f60f2
```
* The exception is caught by [onServiceAcquisitionFailure](https://github.com/twitter/finagle/blob/c80cccfc575494275cfb9d9fbb6a6800b1e7409d/finagle-core/src/main/scala/com/twitter/finagle/service/FailureAccrualFactory.scala#L400) and calls didFail with state=probeOpen. But it lacks a check for state probeOpen and fails to put it back into state=Dead.
* Due to exception it skips [this code block](https://github.com/twitter/finagle/blob/c80cccfc575494275cfb9d9fbb6a6800b1e7409d/finagle-core/src/main/scala/com/twitter/finagle/service/FailureAccrualFactory.scala#L379-L386).
* Thus, the node remains stuck in the ring and causing failures

maheshkelkar:
Problem

- If a memcached node is inactive, FAF pointing to the node enters DEAD state
- After markDeadFor timeout, the FAF enters the startProbing state
- When next probing (i.e. attempt to execute the request) fails, because node
  is still down, and factory (underlying(conn)) throws an exception
- Currently, onServiceAcquisitionFailure function fails to close the probe
- The FAF remains stuck in the ProbeOpen state and thus causing continuous
  failure

Solution

- Inside onServiceAcquisitionFailure, put transition the state into
  ProbeClosed state

- Added stopProbing method
- The onServiceAcquisitionFailure calls stopProbing before calling didFail

Result

Fixes #524

### DIFF
--- a/finagle-core/src/main/scala/com/twitter/finagle/service/FailureAccrualFactory.scala
+++ b/finagle-core/src/main/scala/com/twitter/finagle/service/FailureAccrualFactory.scala
@@ -1,6 +1,5 @@
 package com.twitter.finagle.service
 
-import java.net.InetSocketAddress
 import com.twitter.conversions.time._
 import com.twitter.finagle.Stack.{Params, Role}
 import com.twitter.finagle._
@@ -311,8 +310,8 @@ class FailureAccrualFactory[Req, Rep](
   /**
     * Exit 'Probing' state (if necessary)
     *
-    * The follow-on operation (i.e. the result of first request while probing) will determine
-    * whether the factory transitions to Alive (successful) or Dead (unsuccessful).
+    * The result of the subsequent request will determine whether the factory transitions to
+    * Alive (successful) or Dead (unsuccessful).
     */
   private[this] def stopProbing() = self.synchronized {
     state match {
@@ -337,7 +336,6 @@ class FailureAccrualFactory[Req, Rep](
           // (unsuccessful).
           stopProbing()
 
-          // Invoke service
           service(request).respond { rep =>
             if (isSuccess(ReqRep(request, rep))) didSucceed()
             else didFail()
--- a/finagle-core/src/main/scala/com/twitter/finagle/service/FailureAccrualFactory.scala
+++ b/finagle-core/src/main/scala/com/twitter/finagle/service/FailureAccrualFactory.scala
@@ -1,5 +1,6 @@
 package com.twitter.finagle.service
 
+import java.net.InetSocketAddress
 import com.twitter.conversions.time._
 import com.twitter.finagle.Stack.{Params, Role}
 import com.twitter.finagle._
@@ -254,7 +255,10 @@ class FailureAccrualFactory[Req, Rep](
     }
   }
 
-  private[this] val onServiceAcquisitionFailure: Throwable => Unit = { _ => didFail() }
+  private[this] val onServiceAcquisitionFailure: Throwable => Unit = self.synchronized { _ =>
+    stopProbing()
+    didFail()
+  }
 
   protected def isSuccess(reqRep: ReqRep): Boolean =
     responseClassifier.applyOrElse(reqRep, ResponseClassifier.Default) match {
@@ -304,6 +308,21 @@ class FailureAccrualFactory[Req, Rep](
     cancelReviveTimerTask()
   }
 
+  /**
+    * Exit 'Probing' state (if necessary)
+    *
+    * The follow-on operation (i.e. the result of first request while probing) will determine
+    * whether the factory transitions to Alive (successful) or Dead (unsuccessful).
+    */
+  private[this] def stopProbing() = self.synchronized {
+    state match {
+      case ProbeOpen =>
+        probesCounter.incr()
+        state = ProbeClosed
+      case _ =>
+    }
+  }
+
   def apply(conn: ClientConnection) = {
     underlying(conn).map { service =>
       // N.B. the reason we can't simply filter the service factory is so that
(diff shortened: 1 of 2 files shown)


finagle-core/src/test/scala/com/twitter/finagle/service/FailureAccrualFactoryTest.scala
```diff
@@ -187,6 +177,12 @@ class FailureAccrualFactoryTest extends FunSuite with MockitoSugar {
         assert(statsReceiver.counters.get(List("revivals")) == None)
         assert(factory.isAvailable)
         assert(service.isAvailable)
+
+        // After another failure, the service should be unavailable
+        intercept[Exception] {
+          Await.result(service(123))
+        }
+        assert(statsReceiver.counters.get(List("probes")).getOrElse(0) >= 1)
```
roanta: Let's split this up, it makes it clearer and easier to debug:

```
val probeStat = statsReceiver.counters.get(List("probes"))
assert(probeStat.isDefined && probeStat.get >= 1)
```

finagle-core/src/main/scala/com/twitter/finagle/service/FailureAccrualFactory.scala
```diff
@@ -316,18 +335,9 @@ class FailureAccrualFactory[Req, Rep](
...
-                state match {
-                  case ProbeOpen => state = ProbeClosed
-                  case _ =>
-                }
-              }
-            case _ =>
-          }
+          stopProbing()
 
+          // Invoke service
```
roanta: This comment isn't giving the reader any additional information, I think we can drop it.

finagle-core/src/main/scala/com/twitter/finagle/service/FailureAccrualFactory.scala
```diff
@@ -304,6 +308,21 @@ class FailureAccrualFactory[Req, Rep](
     cancelReviveTimerTask()
   }
 
+  /**
+    * Exit 'Probing' state (if necessary)
+    *
+    * The follow-on operation (i.e. the result of first request while probing) will determine
```
roanta: "The result of the subsequent request will determine..."

finagle-core/src/main/scala/com/twitter/finagle/service/FailureAccrualFactory.scala
```diff
@@ -1,5 +1,6 @@
 package com.twitter.finagle.service
 
+import java.net.InetSocketAddress
```
roanta: Is this used?

jcrossley:
LGTM

maheshkelkar:
@roanta @jcrossley. Updated as per the comments; all the checks have passed. Let me know if this can be merged. 

Also, please let me know which release and approx when will this show up in the bintray

roanta:
lgtm! It will show up in the `develop` branch after we merge it in internally. Our release cadence isn't set in stone, but I would estimate that the next release will be in a month or so.

Thanks again for the contribution!

maheshkelkar:
Thank You!

maheshkelkar:
@roanta, is there a way to ensure that this change does get merged and doesn't miss the boat for release next month

roanta:
Yes, we'll make sure it gets in before the next release. Thanks!

mosesn:
Hey @maheshkelkar thanks for the contribution!  This made it to develop here: https://github.com/twitter/finagle/commit/f0d0b642c2b43f7e11cb014ab2f137ddfe459d70 ✨💞

maheshkelkar:
Thank you @mosesn @roanta @jcrossley

PR #528, opened 2016-07-15
Removes package protection for RawZipkinTracer

adriancole:
Problem

RawZipkinTracer is a base class that is extended to support other
transports like http or kafka. The current implementation requires subtypes to be in the `com.twitter` package. This isn't likely intentional. See #465

Solution

This removes the constraint that
requires such subtypes to be in the `com.twitter` package.

Result

RawZipkinTracer can be extended from other packages, such as `zipkin.finagle`

### DIFF
--- a/finagle-zipkin-core/src/main/scala/com/twitter/finagle/zipkin/core/RawZipkinTracer.scala
+++ b/finagle-zipkin-core/src/main/scala/com/twitter/finagle/zipkin/core/RawZipkinTracer.scala
@@ -15,7 +15,9 @@ import java.nio.ByteBuffer
  * @param statsReceiver We generate stats to keep track of traces sent, failures and so on
  * @param timer A Timer used for timing out spans in the [[DeadlineSpanMap]]
  */
-abstract private[twitter] class RawZipkinTracer(
+// Not private, so that this can be extended to support other transports, such as Kafka.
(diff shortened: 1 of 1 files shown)


adriancole:
cc @sveinnfannar @kevinoliver @mosesn

finagle-zipkin-core/src/main/scala/com/twitter/finagle/zipkin/core/RawZipkinTracer.scala
```diff
@@ -15,7 +15,7 @@ import java.nio.ByteBuffer
  * @param statsReceiver We generate stats to keep track of traces sent, failures and so on
  * @param timer A Timer used for timing out spans in the [[DeadlineSpanMap]]
  */
-abstract private[twitter] class RawZipkinTracer(
```
mosesn: can you add a comment somewhere here that explains why it must be public so we don't make it private again?

mosesn:
LGTM except for a comment

kevinoliver:
(echo of what @mosesn said)

adriancole:
good idea. done!

mosesn:
Thanks for the PR!  This made it in here: https://github.com/twitter/finagle/commit/cd04959b5005b01b9842c3f4d39fc1cef92ff32b

PR #530, opened 2016-07-15
finagle-http: Close server transport when request is not keepalive.

olix0r:
Problem

When a finagle-http server receives a non-keepalive request (i.e. with Connection: close), it does not close the underlying transport. Instead it waits for the client to teardown the connection. This causes undesirable behavior: client connections go into TIME_WAIT on the remote side since they do not receive a FIN from the serverside. This can cause clients to exhaust ports when not using keepalive connections.

Solution

HttpServerDispatcher should terminate connections eagerly when a client requests that the connection be closed.

<hr/>

When running tests against linkerd, we noticed a high number of connection errors from our load generator when not reusing connections.

When connection errors occur, `netstat` on the client host shows that there are 20K+ connections to linkerd in TIME_WAIT, indicating the the client has sent a FIN and is waiting to time the connection out. Our client simply exhausts ports waiting for these connections to time out.

However, when we run this test directly against a Go application, we see no such socket leak. This is because the Go web server terminates client connections eagerly.

After introducing this change, we see the client host stabilize at about 150 sockets in TIMED_WAIT.

### DIFF
--- a/finagle-http/src/main/scala/com/twitter/finagle/http/HttpTransport.scala
+++ b/finagle-http/src/main/scala/com/twitter/finagle/http/HttpTransport.scala
@@ -17,16 +17,22 @@ private[finagle] class HttpTransport[A <: Message, B <: Message](
     manager: ConnectionManager)
   extends StreamTransportProxy[A, B](self) {
 
-  private[this] val readFn: Multi[B] => Unit = { case Multi(m, onFinish) =>
-    manager.observeMessage(m, onFinish)
-    if (manager.shouldClose)
-      self.close()
-  }
+  def this(self: StreamTransport[A, B]) =
+    this(self, new ConnectionManager)
 
-  def this(self: StreamTransport[A, B]) = this(self, new ConnectionManager)
+  // Servers are don't use `status` to determine when they should
+  // close a transport, so we close the transport when the connection
+  // is ready to be closed.
+  manager.onClose.before(self.close())
+
 
   def read(): Future[Multi[B]] =
     self.read().onSuccess(readFn)
+
+  private[this] val readFn: Multi[B] => Unit = { case Multi(m, onFinish) =>
+    manager.observeMessage(m, onFinish)
+  }
+
 
   def write(m: A): Future[Unit] =
     try {
@@ -34,11 +40,11 @@ private[finagle] class HttpTransport[A <: Message, B <: Message](
       manager.observeMessage(m, p)
       val f = self.write(m)
       p.become(f)
-      if (manager.shouldClose) f.before(self.close())
-      else f
+      f
     } catch {
       case NonFatal(e) => Future.exception(e)
     }
+
 
   override def status: CoreStatus = if (manager.shouldClose) CoreStatus.Closed else self.status
 }
--- a/finagle-http/src/main/scala/com/twitter/finagle/http/HttpTransport.scala
+++ b/finagle-http/src/main/scala/com/twitter/finagle/http/HttpTransport.scala
@@ -20,7 +20,7 @@ private[finagle] class HttpTransport[A <: Message, B <: Message](
   def this(self: StreamTransport[A, B]) =
     this(self, new ConnectionManager)
 
-  // Servers are don't use `status` to determine when they should
+  // Servers don't use `status` to determine when they should
   // close a transport, so we close the transport when the connection
   // is ready to be closed.
   manager.onClose.before(self.close())
--- a/finagle-http/src/main/scala/com/twitter/finagle/http/codec/ConnectionManager.scala
+++ b/finagle-http/src/main/scala/com/twitter/finagle/http/codec/ConnectionManager.scala
@@ -1,7 +1,7 @@
 package com.twitter.finagle.http.codec
 
 import com.twitter.finagle.http.{Message, Request, Response}
-import com.twitter.util.Future
+import com.twitter.util.{Future, Promise}
 
 /**
  * The HTTP connection manager implements connection management in
@@ -12,6 +12,7 @@ private[finagle] class ConnectionManager {
   private[this] var isKeepAlive = false
   private[this] var isIdle = true
   private[this] var activeStreams = 0
+  private[this] val closeP = new Promise[Unit]
 
   def observeMessage(message: Message, onFinish: Future[Unit]): Unit = synchronized {
     message match {
@@ -42,8 +43,9 @@ private[finagle] class ConnectionManager {
       activeStreams += 1
       onFinish.ensure {
         endStream()
+        if (shouldClose) closeP.setDone()
       }
-    }
+    } else if (shouldClose) closeP.setDone()
   }
 
   private[this] def endStream(): Unit = synchronized {
@@ -52,4 +54,5 @@ private[finagle] class ConnectionManager {
   }
 
   def shouldClose(): Boolean = synchronized { isIdle && !isKeepAlive }
+  def onClose: Future[Unit] = closeP
 }
--- a/CHANGES
+++ b/CHANGES
@@ -25,6 +25,13 @@ Breaking API Changes
     "dark_traffic_filter/forwarded", "dark_traffic_filter/skipped", and
     "dark_traffic_filter/failed". ``RB_ID=852548``
 
+Runtime Behavior Changes
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+  * finagle-http: HttpTransport now eagerly closes client connection after
+    processing non-keepalive requests.
+
+
 6.36.0
 ------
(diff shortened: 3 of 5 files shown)


finagle-http/src/main/scala/com/twitter/finagle/http/exp/ServerDispatcher.scala
```diff
@@ -41,11 +41,15 @@ private[finagle] abstract class GenSerialServerDispatcher[Req, Rep, In, Out](tra
...
-  protected def handle(rep: Rep): Future[Unit]
+  protected def handle(rep: Rep, keepAlive: Boolean): Future[Unit]
+
+  protected def isReqKeepAlive(req: Out): Boolean = true
+  protected def isRepKeepAlive(rep: Rep): Boolean = true
 
   private[this] def loop(): Future[Unit] = {
     state.set(Idle)
     trans.read().flatMap { case Multi(req, eos) =>
+      val reqKeepAlive = isReqKeepAlive(req)
```
nepthar: I don't think we need to capture, it's only used once.

olix0r: This  intentional.  Requests are mutable, and so I want to capture the keepalive state before the request is dispatched to a service that may, for instance, remove the Connection header to send it somewhere else.

nepthar: Ah, gotchya!

nepthar:
LGTM

mosesn:
@olix0r we already have logic to do this in `ConnectionManager` and `HttpTransport`.  If that doesn't work, can we fix it there instead of adding it here?

olix0r:
@mosesn That's a good point, thanks. I think the problem is that HttpTransport doesn't try to close itself after the stream finishes. Patch incoming.

mosesn:
@olix0r oh, hmm.  I guess because connection pools close client dispatchers with bad statuses?  yeah, I guess server dispatchers should do this too.

mosesn:
LGTM

@nepthar please take another look 😺

finagle-http/src/main/scala/com/twitter/finagle/http/codec/ConnectionManager.scala
```diff
@@ -12,6 +12,7 @@ private[finagle] class ConnectionManager {
   private[this] var isKeepAlive = false
   private[this] var isIdle = true
   private[this] var activeStreams = 0
+  private[this] val closeP = Promise[Unit]
```
kevinoliver: `new Promise[Unit]` or `Promise[Unit]()` please

kevinoliver:
lgtm too.

finagle-http/src/main/scala/com/twitter/finagle/http/HttpTransport.scala
```diff
@@ -17,28 +17,33 @@ private[finagle] class HttpTransport[A <: Message, B <: Message](
...
-  private[this] val readFn: Multi[B] => Unit = { case Multi(m, onFinish) =>
-    manager.observeMessage(m, onFinish)
-    if (manager.shouldClose)
-      self.close()
-  }
+  def this(self: StreamTransport[A, B]) =
+    this(self, new ConnectionManager)
+
+  
+  private[this] val closer =
```
vkostyukov: Do we need this val? I don't see we use it anywhere.

vkostyukov:
🚢 🇮🇹

finagle-http/src/main/scala/com/twitter/finagle/http/codec/ConnectionManager.scala
olix0r: heh. think i copied this style from elsewhere in finagle because it surprised me. happy to standardize on new Promise ;)

olix0r:
rebased, fixed nits

nepthar:
Well, after your changes, you've ventured in to territory that I'm unfamiliar with, but LGTM! I'll merge it soon.

olix0r:
thanks, @nepthar. Let me know if there's any context i can provide to make it a clearer (this is all pretty subtle).

Basically: we (already) track whether a connection should be closed in ConnectionManager---there's a ConnectionManager for each HttpTransport.  HttpTransport can't use `onFinish` to determine when to check the connection for closure since the ConnectionManager may not be updated (there's a race between callbacks on onFinish).  So, we have to introduce a Promise on the connection manager so that HttpTransport can be informed that the connection manager views the connection as ready to be closed...  Whew! Easy, right? ;p

nepthar:
Yup, makes sense, thanks for the overview.

nepthar:
Alrighty, I've run in to some NPEs while running finagle-http's tests:

    1) terminate http/1.1 connections without content length(com.twitter.finagle.http.codec.ConnectionManagerTest)
       java.lang.NullPointerException
        at com.twitter.util.Promise.become(Promise.scala:736)
        at com.twitter.util.Promise$Transformer.k(Promise.scala:106)
        at com.twitter.util.Promise$Transformer.apply(Promise.scala:117)
        at com.twitter.util.Promise$Transformer.apply(Promise.scala:98)
        at com.twitter.util.Promise$$anon$1.run(Promise.scala:412)
        at com.twitter.concurrent.LocalScheduler$Activation.run(Scheduler.scala:201)
        at com.twitter.concurrent.LocalScheduler$Activation.submit(Scheduler.scala:159)
        at com.twitter.concurrent.LocalScheduler.submit(Scheduler.scala:239)
        at com.twitter.concurrent.Scheduler$.submit(Scheduler.scala:107)
        at com.twitter.util.Promise.runq(Promise.scala:405)
        at com.twitter.util.Promise.updateIfEmpty(Promise.scala:791)
        at com.twitter.util.Promise.update(Promise.scala:770)
        at com.twitter.util.Promise.setValue(Promise.scala:746)
        at com.twitter.finagle.http.codec.ConnectionManagerTest.perform(ConnectionManagerTest.scala:118)
        at com.twitter.finagle.http.codec.ConnectionManagerTest$$anonfun$7.apply$mcV$sp(ConnectionManagerTest.scala:142)
        at com.twitter.finagle.http.codec.ConnectionManagerTest$$anonfun$7.apply(ConnectionManagerTest.scala:142)
        at com.twitter.finagle.http.codec.ConnectionManagerTest$$anonfun$7.apply(ConnectionManagerTest.scala:142)
        ...etc

    2) terminate http/1.1 connections with Connection: close(com.twitter.finagle.http.codec.ConnectionManagerTest)
       java.lang.NullPointerException
        at com.twitter.util.Promise.become(Promise.scala:736)
        at com.twitter.util.Promise$Transformer.k(Promise.scala:106)
        at com.twitter.util.Promise$Transformer.apply(Promise.scala:117)
        at com.twitter.util.Promise$Transformer.apply(Promise.scala:98)
        at com.twitter.util.Promise$$anon$1.run(Promise.scala:412)
        at com.twitter.concurrent.LocalScheduler$Activation.run(Scheduler.scala:201)
        at com.twitter.concurrent.LocalScheduler$Activation.submit(Scheduler.scala:159)
        at com.twitter.concurrent.LocalScheduler.submit(Scheduler.scala:239)
        at com.twitter.concurrent.Scheduler$.submit(Scheduler.scala:107)
        at com.twitter.util.Promise.runq(Promise.scala:405)
        at com.twitter.util.Promise.updateIfEmpty(Promise.scala:791)
        at com.twitter.util.Promise.setDone(Promise.scala:760)
        at com.twitter.finagle.http.codec.ConnectionManagerTest.perform(ConnectionManagerTest.scala:113)
        at com.twitter.finagle.http.codec.ConnectionManagerTest$$anonfun$8.apply$mcV$sp(ConnectionManagerTest.scala:150)
        at com.twitter.finagle.http.codec.ConnectionManagerTest$$anonfun$8.apply(ConnectionManagerTest.scala:150)
        at com.twitter.finagle.http.codec.ConnectionManagerTest$$anonfun$8.apply(ConnectionManagerTest.scala:150)
        ... etc

These do seem to directly address what you're working on. Can you let me know if you see them as well?

olix0r:
MOCKIIIIIIITOOOOOOOOOOOO!!!!!

Sorry about that, working on a fix.

nepthar:
Sweet, tests pass internally! Thanks! We're gonna canary before merging and I'll let you know when it lands.

nepthar:
We canaried this internally, but had to cut it very short due to _reasons_ (which were not related to this PR). But we haven't forgotten! I'll update soon.

kevinoliver:
This was merged back in 5c6f9167723c40e24485a98039ec9cf61a258732

PR #537, opened 2016-08-04
Lower max heap size to 3GB (-Xmx3G)

bryce-anderson:
Problem

Travis build is still failing sporadically due to excessive memory usage during sbt builds. Specifically, a test in [finagle-thriftmux](https://github.com/twitter/finagle/blob/7efeb4cc2babd7c99731090fa76ad960627fce14/finagle-thriftmux/src/test/scala/com/twitter/finagle/thriftmux/EndToEndTest.scala#L937) is attempting to allocate a 2GB `Array[Byte]`.

Solution

Lower the max heap size to 3GB. This should be enough, but there is a chance it will fail due to fragmentation of the heap (if I understand the current GC settings correctly...).

### DIFF
--- a/sbt
+++ b/sbt
@@ -28,10 +28,7 @@ java -ea                          \
   $JAVA_OPTS                      \
   -Djava.net.preferIPv4Stack=true \
   -XX:+AggressiveOpts             \
-  -XX:+UseParNewGC                \
-  -XX:+UseConcMarkSweepGC         \
-  -XX:+CMSParallelRemarkEnabled   \
-  -XX:+CMSClassUnloadingEnabled   \
+  -XX:+UseG1GC                    \
   -XX:ReservedCodeCacheSize=128m  \
   -XX:SurvivorRatio=128           \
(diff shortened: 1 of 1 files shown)


mosesn:
Why are you changing the garbage collector?

bryce-anderson:
@mosesn: CMS is not compacting, so trying to allocate 2GB of contiguous space in a 3GB heap has a good chance of failing. G1, on the other hand, is a compacting collector.

mosesn:
Not being able to allocate because old gen is too fragmented makes me think it would trigger a full GC before OOM-ing.  Does it really OOM first?

bryce-anderson:
It definately OOM's, the two links are the failing project build logs before changing the GC. 
https://travis-ci.org/twitter/finagle/jobs/149860894#L6644
https://travis-ci.org/twitter/finagle/jobs/149860894#L6679
(I believe these are the same error, just reported twice)

I will admit that I blamed heap fragmentation for the OOM error which is mostly speculation. Perhaps the _structure_ of the heap is to blame?

PR #545, opened 2016-08-20
serve multiple thrift ifaces

thirstycrow:
Problem

Finagle uses an early version of libthrift, which does not support
service multiplexing.

Solution

Since 0.9.1, thrift added several classes to handle service multiplexing.
On the client side, TMultiplexedProtocol is used to decorate the
original protocol, by prefixing a user specified service name as part of
the name of request message. Finagle uses the protocols from libthrift.
All we have to do is copying TMultiplexedProtocol and related classes
from a newer version of libthrfit.

On the server side, TMultiplexedProcessor is used to inspect the
beginning of the incoming message names, and dispatch the message
to an appropriate underlying processor. Finagle does not use
processors, so we have to re-implement this.

Result

Multiple thrift services can be served through a single port, by passing
a map from service names to service implementations.

``` scala
val serviceMap = Map(
  "echo" -> new EchoService(),
  "extendedEcho" -> new ExtendedEchoService()
)

val server = Thrift.server.serveIfaces(address, serviceMap)
```

Multiple Ifaces/ServiceIfaces can be created sharing the same underlying
thrift client.

``` scala
val client = Thrift.client.multiplex(address, "client") { client =>
  new {
    val echo = client.newIface[Echo.FutureIface]("echo")
    val extendedEcho = client.newServiceIface[ExtendedEcho.ServiceIface]("extendedEcho")
  }
}

client.echo.echo("hello")
client.extendedEcho.getStatus(ExtendedEcho.GetStatus.Args())
```

### DIFF
--- a/finagle-thrift/src/main/scala/com/twitter/finagle/rich.scala
+++ b/finagle-thrift/src/main/scala/com/twitter/finagle/rich.scala
@@ -242,6 +242,7 @@ private[twitter] object ThriftUtil {
    */
   def serverFromIfaces(
     ifaces: Map[String, AnyRef],
+    defaultService: Option[String],
     protocolFactory: TProtocolFactory,
     stats: StatsReceiver,
     maxThriftBufferSize: Int,
@@ -250,7 +251,7 @@ private[twitter] object ThriftUtil {
     val services = ifaces.map { case (serviceName, impl) =>
       serviceName -> serverFromIface(impl, protocolFactory, stats, maxThriftBufferSize, serviceName)
     }
-    new MultiplexedFinagleService(services, protocolFactory, maxThriftBufferSize)
+    new MultiplexedFinagleService(services, defaultService, protocolFactory, maxThriftBufferSize)
   }
 
   /**
@@ -568,6 +569,28 @@ trait ThriftRichClient { self: Client[ThriftClientRequest, Array[Byte]] =>
  * }}}
  *
  * @define serverExampleObject ThriftMuxRichServer
+ *
+ * @define serveIfaces
+ *
+ * Serve multiple interfaces:
+ *
+ * {{{
+ * val serviceMap = Map(
+ * "echo" -> new EchoService(),
+ * "extendedEcho" -> new ExtendedEchoService()
+ * )
+ *
+ * val server = Thrift.server.serveIfaces(address, serviceMap)
+ * }}}
+ *
+ * A default service name can be specified, so we can upgrade an
+ * existing non-multiplexed server to a multiplexed one without
+ * breaking the old clients:
+ *
+ * {{{
+ * val server = Thrift.server.serveIfaces(
+ *   address, serviceMap, defaultService = Some("extendedEcho"))
+ * }}}
  */
 trait ThriftRichServer { self: Server[Array[Byte], Array[Byte]] =>
   import ThriftUtil._
@@ -595,14 +618,14 @@ trait ThriftRichServer { self: Server[Array[Byte], Array[Byte]] =>
     serve(addr, serverFromIface(iface, protocolFactory, serverStats, maxThriftBufferSize, serverLabel))
 
   /**
-   * $serveIface
+   * $serveIfaces
    */
-  def serveIfaces(addr: String, ifaces: Map[String, AnyRef]): ListeningServer =
-    serve(addr, serverFromIfaces(ifaces, protocolFactory, serverStats, maxThriftBufferSize, serverLabel))
+  def serveIfaces(addr: String, ifaces: Map[String, AnyRef], defaultService: Option[String] = None): ListeningServer =
+    serve(addr, serverFromIfaces(ifaces, defaultService, protocolFactory, serverStats, maxThriftBufferSize, serverLabel))
 
   /**
-   * $serveIface
+   * $serveIfaces
    */
-  def serveIfaces(addr: SocketAddress, ifaces: Map[String, AnyRef]): ListeningServer =
-    serve(addr, serverFromIfaces(ifaces, protocolFactory, serverStats, maxThriftBufferSize, serverLabel))
+  def serveIfaces(addr: SocketAddress, ifaces: Map[String, AnyRef], defaultService: Option[String] = None): ListeningServer =
+    serve(addr, serverFromIfaces(ifaces, defaultService, protocolFactory, serverStats, maxThriftBufferSize, serverLabel))
 }
--- a/finagle-thrift/src/main/scala/com/twitter/finagle/rich.scala
+++ b/finagle-thrift/src/main/scala/com/twitter/finagle/rich.scala
@@ -238,6 +238,22 @@ private[twitter] object ThriftUtil {
   }
 
   /**
+   * Construct a multiplexed binary [[com.twitter.finagle.Service]].
+   */
+  def serverFromIfaces(
+    ifaces: Map[String, AnyRef],
+    protocolFactory: TProtocolFactory,
+    stats: StatsReceiver,
+    maxThriftBufferSize: Int,
+    label: String
+  ): BinaryService = {
+    val services = ifaces.map { case (serviceName, impl) =>
+      serviceName -> serverFromIface(impl, protocolFactory, stats, maxThriftBufferSize, serviceName)
+    }
+    new MultiplexedFinagleService(services, protocolFactory, maxThriftBufferSize)
+  }
+
+  /**
    * Construct a binary [[com.twitter.finagle.Service]] for a given Thrift
    * interface using whichever Thrift code-generation toolchain is available.
    * (Legacy version for backward-compatibility).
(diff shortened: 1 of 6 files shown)


kevinoliver:
@thirstycrow this sounds nice and I know there have been teams here at Twitter who would like something like this (they've worked around it in a different way, but I can't recall the details).

I'll try to take a look at your change this week.

finagle-thrift/src/main/scala/com/twitter/finagle/rich.scala
```diff
@@ -380,7 +396,18 @@ trait ThriftRichClient { self: Client[ThriftClientRequest, Array[Byte]] =>
...
+
+  /**
+   * $clientUse
+   */
+  def newIface[Iface](
+    name: Name,
+    label: String,
+    cls: Class[_],
+    protocolFactory: TProtocolFactory,
+    service: Service[ThriftClientRequest, Array[Byte]]): Iface = {
```
kevinoliver: style nitpick: move the `): Iface = {` down to the next line, indented to the same as the `def`.

finagle-thrift/src/main/scala/com/twitter/finagle/rich.scala
```diff
@@ -447,6 +473,43 @@ trait ThriftRichClient { self: Client[ThriftClientRequest, Array[Byte]] =>
   def newMethodIface[ServiceIface, FutureIface](serviceIface: ServiceIface)(
     implicit builder: MethodIfaceBuilder[ServiceIface, FutureIface]
   ): FutureIface = builder.newMethodIface(serviceIface)
+
+  def multiplex[T](dest: Name, label: String)(build: MultiplexedThriftClient => T): T = {
```
kevinoliver: lets do some scaladocs for these new `multiplex` methods.

finagle-thrift/src/main/scala/com/twitter/finagle/rich.scala
```diff
@@ -447,6 +473,43 @@ trait ThriftRichClient { self: Client[ThriftClientRequest, Array[Byte]] =>
...
+    build(new MultiplexedThriftClient(dest, label))
+  }
+
+  def multiplex[T](dest: String, label: String)(build: MultiplexedThriftClient => T): T = {
+    multiplex(Resolver.eval(dest), label)(build)
+  }
+
+  def multiplex[T](dest: String)(build: MultiplexedThriftClient => T): T = {
```
kevinoliver: mind omitting this and forcing these clients to specify a label? its for their own good ;)

finagle-thrift/src/main/scala/com/twitter/finagle/rich.scala
```diff
@@ -447,6 +473,43 @@ trait ThriftRichClient { self: Client[ThriftClientRequest, Array[Byte]] =>
...
+    val (n, l) = Resolver.evalLabeled(dest)
+    multiplex(n, l)(build)
+  }
+
+  class MultiplexedThriftClient(dest: Name, label: String) {
+
+    val service = newService(dest, label)
```
kevinoliver: can you reduce the visibility?

finagle-thrift/src/main/scala/com/twitter/finagle/thrift/MultiplexedService.scala
```diff
@@ -0,0 +1,117 @@
...
+import com.twitter.util.Future
+import java.util.Arrays
+import org.apache.thrift.TApplicationException
+import org.apache.thrift.protocol._
+import org.apache.thrift.transport.TMemoryInputTransport
+import scala.collection.JavaConverters._
+import scala.util.Try
+
+class MultiplexedFinagleService(
+  services: Map[String, Service[Array[Byte], Array[Byte]]],
```
kevinoliver: style nit, format for multi line class declarations:

```
class MyName(
    arg1: Type1,
    ...)
  extends ... {
```

finagle-thrift/src/main/scala/com/twitter/finagle/thrift/MultiplexedService.scala
```diff
@@ -0,0 +1,117 @@
...
+import com.twitter.scrooge.TReusableMemoryTransport
+import com.twitter.util.Future
+import java.util.Arrays
+import org.apache.thrift.TApplicationException
+import org.apache.thrift.protocol._
+import org.apache.thrift.transport.TMemoryInputTransport
+import scala.collection.JavaConverters._
+import scala.util.Try
+
+class MultiplexedFinagleService(
```
kevinoliver: can you reduce visibility?

finagle-thrift/src/main/scala/com/twitter/finagle/thrift/MultiplexedService.scala
```diff
@@ -0,0 +1,117 @@
...
+    extends Service[Array[Byte], Array[Byte]] {
+
+  private val serviceMap = services.mapValues(getFunctionMap)
+
+  def getFunctionMap(
+    service: Service[Array[Byte], Array[Byte]]
+  ): collection.Map[String, (TProtocol, Int) => Future[Array[Byte]]] = {
+    Try(scroogeFinagleServiceFunctionMap(service))
+      .orElse(Try(thriftFinagleServiceFunctionMap(service)))
+      .getOrElse(throw new IllegalArgumentException("%s cannot be multiplexed".format(service.getClass.getName)))
```
kevinoliver: might be helpful to chain the Throw's exception in the failure case

finagle-thrift/src/main/scala/com/twitter/finagle/thrift/MultiplexedService.scala
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+  def scroogeFinagleServiceFunctionMap[T](target: AnyRef) = {
```
kevinoliver: why do we need to support both this and `thriftFinagleServiceFunctionMap`? (mind leaving a comment in the code as a response?)

finagle-thrift/src/main/scala/com/twitter/finagle/thrift/MultiplexedService.scala
```diff
@@ -0,0 +1,117 @@
...
+            val errorMsg = if (functionMap.isEmpty) {
+              s"""Invalid service name: ${serviceName}, available serivce names: [${serviceMap.keys.mkString(", ")}]"""
+            } else {
+              s"""Invalid method name: ${functionName}, available method names: [${functionMap.get.keys.mkString(", ")}]"""
+            }
+            exception(msg.name, msg.seqid, TApplicationException.UNKNOWN_METHOD, errorMsg)
+        }
+      }
+    } catch {
+      case e: Throwable => Future.exception(e)
```
kevinoliver: should probably only catch NonFatal's:

`case NonFatal(e) => Future.exception(e)`

finagle-thrift/src/main/scala/com/twitter/finagle/thrift/MultiplexedService.scala
```diff
@@ -0,0 +1,117 @@
...
+        oprot.writeMessageBegin(new TMessage(name, TMessageType.EXCEPTION, seqid))
+        x.write(oprot)
+        oprot.writeMessageEnd()
+        oprot.getTransport().flush()
+        Future.value(Arrays.copyOfRange(memoryBuffer.getArray(), 0, memoryBuffer.length()))
+      } finally {
+        resetBuffer(memoryBuffer)
+      }
+    } catch {
+      case e: Exception => Future.exception(e)
```
kevinoliver: ditto re NonFatal

finagle-thrift/src/main/scala/com/twitter/finagle/thrift/Protocols.scala
```diff
@@ -78,6 +78,17 @@ object Protocols {
     binaryFactory(statsReceiver = statsReceiver)
   }
 
+  /**
+   * Multiplex the protocolFactory with the serviceName.
+   */
+  def multiplex(serviceName: String, protocolFactory: TProtocolFactory) = {
```
kevinoliver: for public methods, annotate with the return type

finagle-thrift/src/main/scala/com/twitter/finagle/thrift/Protocols.scala
```diff
@@ -78,6 +78,17 @@ object Protocols {
     binaryFactory(statsReceiver = statsReceiver)
   }
 
+  /**
+   * Multiplex the protocolFactory with the serviceName.
```
kevinoliver: explain why would you use this? i'm guessing its when you have a client that is talking to a multiplexed backend service?

kevinoliver:
also, apologies for the delay in reviewing. it got lost in my inbox.

thirstycrow:
@kevinoliver I'm not quite familiar with the SHA thing. Could you add it before merging this pull request? The libthrift code is copied exactly from [the 0.9.1 tag](https://github.com/apache/thrift/tree/0.9.1)

I added a param to specify a default service，so the server can be upgraded to use mutiplexing, and still be compatible with the old clients.
```scala
val server = Thrift.server.serveIfaces(
  address, serviceMap, defaultService = Some("extendedEcho"))
```

taylorleese:
What (if anything) needs to be done for this to also work with ThriftMux?

kevinoliver:
> I'm not quite familiar with the SHA thing. Could you add it before merging this pull request? The libthrift code is copied exactly from the 0.9.1 tag

a link to that tag works perfectly for me. the idea is just to help future devs track down exactly where the code originated from.

kevinoliver:
the default service addition is nice. thanks.

thirstycrow:
@taylorleese I think there's nothing special needs to be done to work with ThriftMux, which does not depend on the details of how a thrift message is encoded.

kevinoliver:
this lgtm. be good to have another set of eyes on it.

mosesn:
@thirstycrow sorry for the delay, this looks good to me!  I'll see if I can merge it in.

mosesn:
@thirstycrow it's gonna be a bit longer, we need to get legal to sign off on copying the code from Apache Thrift in.

mosesn:
@thirstycrow sorry for the delay, our lawyers finally got back to us.  this was merged in! https://github.com/twitter/finagle/commit/b9cb4883ce35e4f5a88c37d0fc4eeb507ddda0a0

Issue #546, opened 2016-08-21
Documentation/Getting started improvements


asheshambasta:
Make documentation clearer and getting started easier.

### Expected behavior

Getting started guide leads to broken situations with no apparent explanation on how to approach things. 
I set up a simple Finagle project with `build.sbt` that looks like this;
```sbt
def finagleVer = "6.36.0"

resolvers += "Typesafe repository" at "http://repo.typesafe.com/typesafe/releases/"

libraryDependencies ++= Seq(
  "com.twitter" %% "finagle-http" % finagleVer,
  "com.typesafe.play" %% "play-json" % "2.3.8"
)

fork in run := true
```

And I've run into some very weird issues.
Expected:
- should be able to follow the getting started guide from start to finish with the quick setup

### Actual behavior

- 6.36.0 fails to resolve
- tried with the sbt support in IntelliJ and the CLI with no luck.

### Steps to reproduce the behavior

- set up a simple project like stated above.

This makes the process of getting started very slow and painful, imo.

mosesn:
@asheshambasta which version of scala are you running?  Please make sure you're on 2.11, we no longer support 2.10 as of 6.36.0.

asheshambasta:
@mosesn 
That was it.
It would be helpful to add that [here](http://twitter.github.io/finagle/guide/Quickstart.html)

mosesn:
@asheshambasta if you would like to make a PR, we would appreciate it!

asheshambasta:
@mosesn https://github.com/twitter/finagle/pull/547

PR #547, opened 2016-08-23
specify Scala version in quickstart for versions >= `6.36.0`.

asheshambasta:
Problem

The quickstart doesn't specify the scala version based on the version of Finagle used. See: https://github.com/twitter/finagle/issues/546#issuecomment-241270739

Solution

Simple modifications that add a line note specifying which Scala version to use as of `6.36.0`.

Result

An extra line will be added to the quickstart specifying the Scala version to use.

### DIFF
--- a/doc/src/sphinx/Quickstart.rst
+++ b/doc/src/sphinx/Quickstart.rst
@@ -22,6 +22,7 @@ Setting up SBT
 
 We'll use sbt_ to build our project. Finagle is published to Maven Central,
 so little setup is needed: see the ``build.sbt`` in the ``quickstart`` directory.
+Also ensure that you're using ``scala`` ``2.11`` as of ``6.36.0``.
 
 .. parsed-literal::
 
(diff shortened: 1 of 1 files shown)


doc/src/sphinx/Quickstart.rst
```diff
@@ -22,6 +22,7 @@ Setting up SBT
 
 We'll use sbt_ to build our project. Finagle is published to Maven Central,
 so little setup is needed: see the ``build.sbt`` in the ``quickstart`` directory.
+Also ensure that you're using ``scala`` ``2.11`` as of ``6.36.0``.
```
mosesn: It might be better to phrase this as "we no longer support 2.10" since we plan on adding 2.12 support when it comes out (which is hopefully soon).

mosesn:
LGTM except for one  nit.  Thanks!

vkostyukov:
Thanks @asheshambasta! I'm pulling this internally. Will close this PR when it's merged.

vkostyukov:
It's merged internally. Will show up on Github this Monday. Thanks @asheshambasta!

Issue #549, opened 2016-08-29
Finagle-6.37.0 fails to behave properly when `Connection: close` is set


olix0r:
Finagle-6.37.0 appears to introduce a bug so that requests issued with `Connection: close` fail with a cancelled request exception

### Expected behavior

Requests should succeed.

### Actual behavior

Requests fail with a CancelledRequestException.

### Steps to reproduce the behavior

Simplest reproduction:
```scala
  test("simple request with Connection: close") {
    val server = Http.serve(new InetSocketAddress(0), ServiceFactory.const(
      Service.mk[Request, Response] { req =>
        Future.value(req.response)
      }
    ))
    val addr = server.boundAddress.asInstanceOf[InetSocketAddress]
    val client = Http.newService(s"/$$/inet/127.1/${addr.getPort}")
    try {
      val req = Request()
      req.headerMap.set("Connection", "close")
      assert(Await.result(client(req)).status == Status.Ok)
    } finally Await.result(client.close().join(server.close()).unit)
  }
```

Fails with:
```
Aug 29, 2016 11:11:39 PM com.twitter.util.RootMonitor$ handle
SEVERE: Exception propagated to the root monitor!
Failure(request cancelled. Remote Info: Not Available, flags=0x02) with Service -> 
Caused by: com.twitter.finagle.CancelledRequestException: request cancelled. Remote Info: Not Available
...
[info] - simple request with Connection: close *** FAILED ***
[info]   com.twitter.finagle.ChannelClosedException: ChannelException at remote address: /127.0.0.1:56028 from service: /$/inet/127.1/56028. Remote Info: Upstream Address: Not Available, Upstream Client Id: Not Available, Downstream Address: /127.0.0.1:56028, Downstream Client Id: /$/inet/127.1/56028, Trace Id: fc1c7c922cd044f8.fc1c7c922cd044f8<:fc1c7c922cd044f8
```

I am guessing that this is somehow related to #530, which seems to be the only relevant change in 6.37.0.

mosesn:
Relatedly, http/1.0 requests have a hard time too.

PR #550, opened 2016-08-30
finagle-http: fix connection management for non-keepalive requests.

olix0r:
Problem

When issuing HTTP/1.0 or non-keep-alive requests (i.e. with Connection: close),
the connection may be marked as closable before a request/response has
completed.  Specifically, when finagle-http's ConnectionManager observes a
non-keepalive request, it may satisfy onClose before a response is observed.
This causes the entire connection to be torn down (and the response canceled)
before the response can be received.

Solution

Consider a connection to be closable only when an equal number of requests and
responses have been observed.

Fixes twitter/finagle#549

### DIFF
--- a/finagle-http/src/main/scala/com/twitter/finagle/http/codec/ConnectionManager.scala
+++ b/finagle-http/src/main/scala/com/twitter/finagle/http/codec/ConnectionManager.scala
@@ -9,9 +9,26 @@ import com.twitter.util.{Future, Promise}
  * codec implementations are in {Server,Client}ConnectionManager.
  */
 private[finagle] class ConnectionManager {
+
+  /** Indicates whether the connection should be closed when it becomes idle. */
   private[this] var isKeepAlive = false
+
+  /** When false, the connection is busy servicing a a request. */
   private[this] var isIdle = true
-  private[this] var activeStreams, pendingResponses = 0
+
+  /**
+   * Indicates the number of chunked messages currently being
+   * transmitted on this connection. Practically, on [0, 2].
+   */
+  private[this] var activeStreams = 0
+
+  /**
+   * Indicates the number of requests that have been issues that have
+   * not yet received a response. Practically, on [0, 1]
+   */
+  private[this] var pendingResponses = 0
+
+  /** Satisfied when the connection is ready to be torn down. */
   private[this] val closeP = new Promise[Unit]
 
   def observeMessage(message: Message, onFinish: Future[Unit]): Unit = synchronized {
--- a/finagle-http/src/main/scala/com/twitter/finagle/http/codec/ConnectionManager.scala
+++ b/finagle-http/src/main/scala/com/twitter/finagle/http/codec/ConnectionManager.scala
@@ -11,7 +11,7 @@ import com.twitter.util.{Future, Promise}
 private[finagle] class ConnectionManager {
   private[this] var isKeepAlive = false
   private[this] var isIdle = true
-  private[this] var activeStreams, requestsObserved, responsesObserved = 0
+  private[this] var activeStreams, pendingResponses = 0
   private[this] val closeP = new Promise[Unit]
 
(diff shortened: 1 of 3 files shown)


finagle-http/src/main/scala/com/twitter/finagle/http/codec/ConnectionManager.scala
```diff
@@ -11,7 +11,7 @@ import com.twitter.util.{Future, Promise}
 private[finagle] class ConnectionManager {
   private[this] var isKeepAlive = false
   private[this] var isIdle = true
-  private[this] var activeStreams = 0
+  private[this] var activeStreams, requestsObserved, responsesObserved = 0
```
olix0r: alternatively, we could track this as a single `pendingResponses` field (much like `activeStreams`).  I suppose this would be slightly better as it has no hope of overflowing

vkostyukov: FWIF, I'm voting for a single variable here (`activeRequests` or `pendingResponses`).

vkostyukov:
👍 

Thanks for looking into this, @olix0r!

finagle-http/src/main/scala/com/twitter/finagle/http/codec/ConnectionManager.scala
```diff
@@ -11,7 +11,7 @@ import com.twitter.util.{Future, Promise}
 private[finagle] class ConnectionManager {
   private[this] var isKeepAlive = false
   private[this] var isIdle = true
-  private[this] var activeStreams = 0
+  private[this] var activeStreams, pendingResponses = 0
```
mosesn: since ConnectionManager can be used for both clients and servers, this might be better phrased as `pending` and we should add a note that for servers it goes negative instead of positive.

mosesn: Can we also document the purpose of both `activeStreams` and `pending`?

olix0r: @mosesn I don't quite understand... `pendingResponses` is always incremented in `observeRequest` and always decremented in `observeResponse`.  Isn't `pendingResponses` is an accurate name? Is there a situation where a response is observed before a request?

mosesn: Ohhhhh . . . yeah, you're right.  Ignore me.

mosesn:
LGTM

vkostyukov:
I'm pooling this internally. Thanks @olix0r! Will ping you here when it's merged.

olix0r:
Thanks for merging!

It looks like this just missed the cut for 6.38.0, but #549 is a show-stopper for linkerd. It would be _hugely_ helpful (we can ship linkerd's grpc support) to get this into a stable release in the near future (hopefully days and not weeks?).  Let me know if there's someone I should particularly nice to...

samstarling:
I'm trying to track down some behaviour that changed between v6.35.0 and v6.38.0, and I think this PR might be related. The following code now throws a `com.twitter.finagle.ChannelClosedException`, rather than the HTTP response. Is there something I need to change?

/cc @olix0r @vkostyukov 

```scala
import com.twitter.finagle.builder.ClientBuilder
import com.twitter.finagle.http.{Http, Method, Request, Version}
import com.twitter.util.Await

val client = ClientBuilder()
  .codec(Http())
  .hosts("www.google.com:80")
  .hostConnectionLimit(25)
  .build()

val request = Request(Version.Http11, Method.Get, "/")
request.headerMap.add("Connection", "close")
val response = Await.result(client(request))
response.getContentString()
```

olix0r:
(repeating what I said in Gitter for the benefit of others)

@samstarling you are correct that this PR fixes the issue you observe. I am told that 6.39 will be released in the next ~week including this fix.

PR #553, opened 2016-09-20
Tolerates reads of 128 bit X-B3-TraceId

### LINKED ISSUE #552: Don't break on 128bit X-B3-TraceId by tossing high bits
We want to phase in 128bit trace ids per See https://github.com/openzipkin/b3-propagation/issues/6. The first step is not breaking if a longer trace id is found in an http header.
### Expected behavior

The first step of transitioning to 128bit `X-B3-TraceId` is tolerantly reading 32 character long ids by throwing away the high bits (any characters left of 16 characters). This allows the tracing system to more flexibly introduce 128bit trace id support in the future.

Ex. when `X-B3-TraceId: 463ac35c9f6413ad48485a3953bb6124` is received, parse the lower 64 bits (right most 16 characters ex48485a3953bb6124) as the trace id.
### Actual behavior

Right now, the http codec parses the `X-B3-TraceId` header via `SpanId.fromString`.

``` scala
  def fromString(spanId: String): Option[SpanId] =
    try {
      Some(SpanId(new RichU64String(spanId).toU64Long))
    } catch {
      case NonFatal(_) => None
    }
```

We could either change this to toss left-most characters over 16, or define a new function for parsing a trace id. For example, like this:

``` java
  /**
   * Parses a 1 to 32 character lower-hex string with no prefix into an unsigned long, tossing any
   * bits higher than 64.
   */
  public static long lowerHexToUnsignedLong(String lowerHex) {
    int length = lowerHex.length();
    if (length < 1 || length > 32) throw isntLowerHexLong(lowerHex);

    // trim off any high bits
    int i = length > 16 ? length - 16 : 0;

    long result = 0;
    for (; i < length; i++) {
      char c = lowerHex.charAt(i);
      result <<= 4;
      if (c >= '0' && c <= '9') {
        result |= c - '0';
      } else if (c >= 'a' && c <= 'f') {
        result |= c - 'a' + 10;
      } else {
        throw isntLowerHexLong(lowerHex);
      }
    }
    return result;
  }

  static NumberFormatException isntLowerHexLong(String lowerHex) {
    throw new NumberFormatException(
        lowerHex + " should be a 1 to 32 character lower-hex string with no prefix");
  }
```

Once this is in, we should be able to test by ensuring that the following parse the same value:

`X-B3-TraceId: 463ac35c9f6413ad48485a3953bb6124`
`X-B3-TraceId: 48485a3953bb6124`

adriancole:
Problem

The first step of transitioning to 128bit `X-B3-TraceId` is tolerantly reading 32 character long ids. Until a change is made, those propagating 128bit ids to finagle will have their traces restarted due to a parse failure.

Solution

I've changed the SpanId parser to leniently parse an unsigned 64bit long from a lower-hex string. This works by throwing away the high bits (any characters left of 16 characters)

Result

When a 128bit id like `X-B3-TraceId: 463ac35c9f6413ad48485a3953bb6124` is received, the lower 64 bits (right most 16 characters ex48485a3953bb6124) will become the trace id as opposed to having the trace restarted.

Fixes #552

### DIFF
--- a/finagle-core/src/main/scala/com/twitter/finagle/tracing/Id.scala
+++ b/finagle-core/src/main/scala/com/twitter/finagle/tracing/Id.scala
@@ -49,7 +49,11 @@ object SpanId {
 
   def fromString(spanId: String): Option[SpanId] =
     try {
-      Some(SpanId(new RichU64String(spanId).toU64Long))
+      // Tolerates 128 bit X-B3-TraceId by reading the right-most 16 hex
+      // characters (as opposed to overflowing a U64 and starting a new trace).
+      val length = spanId.length()
+      val lower64Bits = if (length <= 16) spanId else spanId.substring(length - 16)
+      Some(SpanId(new RichU64String(lower64Bits).toU64Long))
     } catch {
       case NonFatal(_) => None
     }
--- a/finagle-thrift/src/main/ruby/lib/finagle-thrift/trace.rb
+++ b/finagle-thrift/src/main/ruby/lib/finagle-thrift/trace.rb
@@ -100,13 +100,14 @@ module Trace
   end
 
   class SpanId
-    HEX_REGEX = /^[a-f0-9]{16}$/i
+    HEX_REGEX = /^[a-f0-9]{16,32}$/i
     MAX_SIGNED_I64 = 9223372036854775807
     MASK = (2 ** 64) - 1
 
     def self.from_value(v)
       if v.is_a?(String) && v =~ HEX_REGEX
-        new(v.hex)
+        # drops any bits higher than 64 by selecting right-most 16 characters
+        new(v.length > 16 ? v[v.length - 16, 16].hex : v.hex)
       elsif v.is_a?(Numeric)
         new(v)
       elsif v.is_a?(SpanId)
--- a/finagle-core/src/test/scala/com/twitter/finagle/tracing/SpanIdTest.scala
+++ b/finagle-core/src/test/scala/com/twitter/finagle/tracing/SpanIdTest.scala
@@ -18,6 +18,12 @@ class SpanIdTest extends FunSuite {
     assert(SpanId.fromString("7b").get.toLong == 123L)
   }
 
+  test("create a span from a 128bit hex string by dropping high bits") {
+    val hex128Bits = "463ac35c9f6413ad48485a3953bb6124"
+    val lower64Bits = "48485a3953bb6124"
+    assert(SpanId.fromString(hex128Bits).get == SpanId.fromString(lower64Bits).get)
+  }
+
   test("return None if string is not valid hex") {
     assert(SpanId.fromString("rofl") == None)
   }
--- a/finagle-thrift/src/main/ruby/test/trace_test.rb
+++ b/finagle-thrift/src/main/ruby/test/trace_test.rb
@@ -19,6 +19,11 @@ class TraceIdTest < Test::Unit::TestCase
     id2 = Trace::SpanId.from_value(id.to_s)
     assert_equal id.to_i, id2.to_i
   end
+
+  def test_span_id_128
+    from_hex_128 = Trace::SpanId.from_value('463ac35c9f6413ad48485a3953bb6124')
+    assert_equal from_hex_128.to_s, '48485a3953bb6124'
+  end
 end
 
 class TraceTest < Test::Unit::TestCase


finagle-thrift/src/main/ruby/lib/finagle-thrift/trace.rb
```diff
@@ -100,13 +100,14 @@ class Flags
   end
 
   class SpanId
-    HEX_REGEX = /^[a-f0-9]{16}$/i
+    HEX_REGEX = /^[a-f0-9]{16,32}$/i
```
adriancole: @jcarres-mdsol I think the https://github.com/openzipkin/zipkin-ruby relies on this class, still. lemme know if my adjustment is not idiomatic

finagle-core/src/main/scala/com/twitter/finagle/tracing/Id.scala
```diff
@@ -47,12 +47,33 @@ object SpanId {
...
     }
+
+    // trim off any high bits
+    val startIndex = if (length > 16) length - 16 else 0
+
+    var result = 0L;
+    for (i <- startIndex until length) {
+      val c = spanId.charAt(i)
+      result <<= 4
+      if (c >= '0' && c <= '9') {
```
mosesn: ffti: `c.isDigit`

finagle-core/src/main/scala/com/twitter/finagle/tracing/Id.scala
```diff
@@ -47,12 +47,33 @@ object SpanId {
 
   def apply(spanId: Long): SpanId = new SpanId(spanId)
 
-  def fromString(spanId: String): Option[SpanId] =
-    try {
-      Some(SpanId(new RichU64String(spanId).toU64Long))
```
mosesn: why not keep doing this, but only use the last characters?

adriancole: mainly as I noticed other performance related code in this type. retaining the last chars means a substring, which means another string. you're right, it is far less code to do it that way. wdyt?

finagle-core/src/main/scala/com/twitter/finagle/tracing/Id.scala
adriancole: again, performance related.. the code I pasted benchmarks better. you'll find that isDigit unfolds into quite many things. no problem reverting to RichU64String, if you say go!

adriancole: here's the difference in the java variant:

if (c >= '0' && c <= '9') { 

Benchmark                                   Mode  Cnt   Score   Error   Units
UtilBenchmarks.lowerHexToUnsignedLong_128  thrpt   15  53.386 ± 1.810  ops/us
UtilBenchmarks.lowerHexToUnsignedLong_64   thrpt   15  57.554 ± 2.446  ops/us

if (Character.isDigit(c)) { 

Benchmark                                   Mode  Cnt   Score   Error   Units
UtilBenchmarks.lowerHexToUnsignedLong_128  thrpt   15  33.529 ± 1.376  ops/us
UtilBenchmarks.lowerHexToUnsignedLong_64   thrpt   15  33.341 ± 1.120  ops/us

adriancole:
@mosesn here's a summary.. agree the performance stuff is distracting.. I lept to it seeing other performance work in the file.

Interestingly, there seems to be no performance impact if we just did a substring using the existing RichU64String stuff. If we keep with RichU64String, will just do substring.

The better numbers from the code I pasted are the results of attempts to get total overhead (knowing there's often 3 ids to parse, and tickers etc) in microsecond scale. In context, particularly around pulling headers off the wire, it might not be worth it here. lemme know if I should drop the custom id parsing function.

```
Benchmark                                   Mode  Cnt   Score   Error   Units
UtilBenchmarks.lowerHexToUnsignedLong_128  thrpt   15  53.340 ± 1.851  ops/us
UtilBenchmarks.lowerHexToUnsignedLong_64   thrpt   15  59.829 ± 2.306  ops/us
UtilBenchmarks.richU64StringToU64Long_128  thrpt   15   2.589 ± 0.071  ops/us
UtilBenchmarks.richU64StringToU64Long_64   thrpt   15   2.582 ± 0.156  ops/us
```

adriancole:
ps looked at the test failures and they look unrelated

finagle-thrift/src/main/ruby/lib/finagle-thrift/trace.rb
jcarres-mdsol: I think we are good and this will work properly for us. It will work as expected, will read a string (which comes from the header) and create a SpanId class from it.

mosesn:
OK, I think we should keep it simple and just use substring for now.  If we find subsequently that this is a problem, we can fix it later.

adriancole:
SGTM change on the way!

adriancole:
@mosesn ok much smaller change now

finagle-core/src/main/scala/com/twitter/finagle/tracing/Id.scala
```diff
@@ -49,7 +49,11 @@ object SpanId {
 
   def fromString(spanId: String): Option[SpanId] =
     try {
-      Some(SpanId(new RichU64String(spanId).toU64Long))
+      // Tolerates 128 bit X-B3-TraceId by reading the right-most 16 hex
+      // characters (as opposed to overflowing a U64 and starting a new trace).
+      val length = spanId.length()
+      val lower64Bits = length <= 16 ? spanId : spanId.substring(length - 16)
```
mosesn: in scala, we usually use `if (length <= 16) spanId else spanId.substring(length - 16)` as the ternary operator.

adriancole:
thanks for review! (and the tip on ternary. fixed)

mosesn:
LGTM

kevinoliver:
lgtm too. i'll get this merged.

kevinoliver:
This got merged locally and should make it out to `develop` in today's push. Thanks @adriancole.

kevinoliver:
Merged in f8e76124d1b5f344176b79697b5c911e2c8f44fc

adriancole:
Thanks for the support!

PR #557, opened 2016-09-22
Added `withHttpStats` config to Http.[Server|Client].

spockz:
Problem

com.twitter.finagle.http.filter.StatsFilter Is missing a corresponding module
so it cannot be conveniently added to a Stack.

Solution

Added the module, added the `withHttpStats` methods to Client and Server.

Result

It is now possible to effortlessly enable Http specific statistics on both
http servers and clients.

### DIFF
--- a/doc/src/sphinx/Metrics.rst
+++ b/doc/src/sphinx/Metrics.rst
@@ -195,6 +195,10 @@ These stats pertain to the HTTP protocol.
   A counter of the number of non-retryable HTTP 503 responses the Http server returns. Those
   responses are not automatically retried.
 
+These metrics are added by
+:src:`StatsFilter <com/twitter/finagle/http/filter/StatsFilter.scala>` and can be enabled by
+using `.withHttpStats` on `Http.Client` and `Http.Server`.
+
 **status/<statusCode>**
   A counter of the number of responses received, or returned for servers, that had this
   statusCode.
@@ -204,10 +208,10 @@ These stats pertain to the HTTP protocol.
   count as 5XX for this counter.
 
 **time/<statusCode>**
-  Metric on duration per Http status code.
+  A histogram on duration per Http status code.
 
 **time/<statusCategory>**
-  Metric on duration per Http status code category.
+  A histogram on duration per Http status code category.
 
 
 Mux
--- a/doc/src/sphinx/Metrics.rst
+++ b/doc/src/sphinx/Metrics.rst
@@ -195,6 +195,21 @@ These stats pertain to the HTTP protocol.
   A counter of the number of non-retryable HTTP 503 responses the Http server returns. Those
   responses are not automatically retried.
 
+**status/<statusCode>**
+  A counter of the number of responses received, or returned for servers, that had this
+  statusCode.
+
+**status/<statusClass>**
+  Same as **status/statusCode** but aggregated per category, e.g. all 500 range responses
+  count as 5XX for this counter.
+
+**time/<statusCode>**
+  Metric on duration per Http status code.
+
+**time/<statusCategory>**
+  Metric on duration per Http status code category.
+
+
 Mux
 ---
 
--- a/doc/src/sphinx/Metrics.rst
+++ b/doc/src/sphinx/Metrics.rst
@@ -189,10 +189,10 @@ These stats pertain to the HTTP protocol.
 
 **nacks**
   A counter of the number of retryable HTTP 503 responses the Http server returns. Those
-  responses are automatically retried by Finagle Http client.
+  responses are automatically retried by Finagle HTTP client.
 
 **nonretryable_nacks**
-  A counter of the number of non-retryable HTTP 503 responses the Http server returns. Those
+  A counter of the number of non-retryable HTTP 503 responses the HTTP server returns. Those
   responses are not automatically retried.
 
 These metrics are added by
@@ -208,10 +208,10 @@ using `.withHttpStats` on `Http.Client` and `Http.Server`.
   count as 5XX for this counter.
 
 **time/<statusCode>**
-  A histogram on duration per Http status code.
+  A histogram on duration in milliseconds per HTTP status code.
 
 **time/<statusCategory>**
-  A histogram on duration per Http status code category.
+  A histogram on duration in milliseconds per HTTP status code category.
 
 
 Mux
--- a/finagle-core/src/main/scala/com/twitter/finagle/Stack.scala
+++ b/finagle-core/src/main/scala/com/twitter/finagle/Stack.scala
@@ -455,6 +455,11 @@ object Stack {
       Node(this, (prms, next) => Leaf(this, make(next.make(prms))), next)
   }
 
+  class NoOpModule[T](val role: Role, val description: String) extends Module0[T] {
+    override def make(next: T): T =
+      next
+  }
+
   /** A module of 1 parameter. */
   abstract class Module1[P1: Param, T] extends Stackable[T] {
     final val parameters: Seq[Stack.Param[_]] =
--- a/finagle-core/src/main/scala/com/twitter/finagle/Stack.scala
+++ b/finagle-core/src/main/scala/com/twitter/finagle/Stack.scala
@@ -456,7 +456,7 @@ object Stack {
   }
 
   class NoOpModule[T](val role: Role, val description: String) extends Module0[T] {
-    override def make(next: T): T =
+    def make(next: T): T =
       next
   }
 
--- a/finagle-http/src/main/scala/com/twitter/finagle/Http.scala
+++ b/finagle-http/src/main/scala/com/twitter/finagle/Http.scala
@@ -174,6 +174,7 @@ object Http extends Client[Request, Response] with HttpRichClient
         .replace(TraceInitializerFilter.role, new HttpClientTraceInitializer[Request, Response])
         .prepend(http.TlsFilter.module)
         .prepend(nonChunkedPayloadSize)
+        .prepend(new Stack.NoOpModule(http.filter.StatsFilter.role, http.filter.StatsFilter.description))
 
     private def params: Stack.Params =
       StackClient.defaultParams +
@@ -272,8 +273,8 @@ object Http extends Client[Request, Response] with HttpRichClient
     /**
      * Enable the collection of HTTP specific metrics. See [[http.filter.StatsFilter]].
      */
-    def enableHttpStats(): Client =
-      withStack(http.filter.StatsFilter.module +: stack)
+    def withHttpStats: Client =
+      withStack(stack.replace(http.filter.StatsFilter.role, http.filter.StatsFilter.module[Request]))
 
     // Java-friendly forwarders
     // See https://issues.scala-lang.org/browse/SI-8905
@@ -323,6 +324,7 @@ object Http extends Client[Request, Response] with HttpRichClient
         .replace(StackServer.Role.preparer, HttpNackFilter.module)
         .prepend(nonChunkedPayloadSize)
         .prepend(ServerContextFilter.module)
+        .prepend(new Stack.NoOpModule(http.filter.StatsFilter.role, http.filter.StatsFilter.description))
 
     private def params: Stack.Params =
       StackServer.defaultParams +
@@ -414,8 +416,8 @@ object Http extends Client[Request, Response] with HttpRichClient
     /**
      * Enable the collection of HTTP specific metrics. See [[http.filter.StatsFilter]].
      */
-    def enableHttpStats(): Server =
-      withStack(http.filter.StatsFilter.module +: stack)
+    def withHttpStats: Server =
+      withStack(stack.replace(http.filter.StatsFilter.role, http.filter.StatsFilter.module[Request]))
 
     // Java-friendly forwarders
     // See https://issues.scala-lang.org/browse/SI-8905
--- a/finagle-http/src/main/scala/com/twitter/finagle/Http.scala
+++ b/finagle-http/src/main/scala/com/twitter/finagle/Http.scala
@@ -274,7 +274,7 @@ object Http extends Client[Request, Response] with HttpRichClient
      * Enable the collection of HTTP specific metrics. See [[http.filter.StatsFilter]].
      */
     def withHttpStats: Client =
-      withStack(stack.replace(http.filter.StatsFilter.role, http.filter.StatsFilter.module[Request]))
+      withStack(stack.replace(http.filter.StatsFilter.role, http.filter.StatsFilter.module))
 
     // Java-friendly forwarders
     // See https://issues.scala-lang.org/browse/SI-8905
@@ -417,7 +417,7 @@ object Http extends Client[Request, Response] with HttpRichClient
      * Enable the collection of HTTP specific metrics. See [[http.filter.StatsFilter]].
      */
     def withHttpStats: Server =
-      withStack(stack.replace(http.filter.StatsFilter.role, http.filter.StatsFilter.module[Request]))
+      withStack(stack.replace(http.filter.StatsFilter.role, http.filter.StatsFilter.module))
 
     // Java-friendly forwarders
     // See https://issues.scala-lang.org/browse/SI-8905
--- a/finagle-http/src/main/scala/com/twitter/finagle/Http.scala
+++ b/finagle-http/src/main/scala/com/twitter/finagle/Http.scala
@@ -268,6 +268,13 @@ object Http extends Client[Request, Response] with HttpRichClient
     def withCompressionLevel(level: Int): Client =
       configured(param.CompressionLevel(level))
 
+
+    /**
+     * Enable the collection of HTTP specific metrics. See [[http.filter.StatsFilter]].
+     */
+    def enableHttpStats(): Client =
+      withStack(http.filter.StatsFilter.module +: stack)
+
     // Java-friendly forwarders
     // See https://issues.scala-lang.org/browse/SI-8905
     override val withSessionPool: SessionPoolingParams[Client] =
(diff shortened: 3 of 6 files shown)


finagle-http/src/main/scala/com/twitter/finagle/Http.scala
```diff
@@ -268,6 +268,13 @@ object Http extends Client[Request, Response] with HttpRichClient
     def withCompressionLevel(level: Int): Client =
       configured(param.CompressionLevel(level))
 
+
+    /**
+     * Enable the collection of HTTP specific metrics. See [[http.filter.StatsFilter]].
+     */
+    def enableHttpStats(): Client =
```
kevinoliver: did you consider `withHttpStats` as a name? cc @vkostyukov (here and below)

finagle-http/src/main/scala/com/twitter/finagle/http/filter/StatsFilter.scala
```diff
@@ -1,9 +1,29 @@
...
-import com.twitter.finagle.{Service, SimpleFilter}
-import com.twitter.finagle.stats.StatsReceiver
+import com.twitter.finagle._
 import com.twitter.finagle.http.{Request, Response, Status}
-import com.twitter.util.{Duration, Future, Return, Stopwatch, Throw}
+import com.twitter.finagle.stats.StatsReceiver
+import com.twitter.util._
+
+object StatsFilter {
+  val role = Stack.Role("HttpStatsFilter")
```
kevinoliver: i know we haven't always done a good job in the past, but i'd like to have public methods and members have type annotations.
`val role: Stack.Role = ...`

finagle-http/src/main/scala/com/twitter/finagle/http/filter/StatsFilter.scala
```diff
@@ -1,9 +1,29 @@
...
+
+  def module[Req <: Request]: Stackable[ServiceFactory[Req, Response]] =
+    new Stack.Module1[param.Stats, ServiceFactory[Req, Response]] {
+      val role = StatsFilter.role
+      val description = "Register HTTP Stats"
```
kevinoliver: omit "Register"?

finagle-http/src/main/scala/com/twitter/finagle/http/filter/StatsFilter.scala
```diff
@@ -29,7 +49,7 @@ class StatsFilter[REQUEST <: Request](stats: StatsReceiver)
     future respond {
       case Return(response) =>
         count(elapsed(), response)
-      case Throw(_) =>
+      case Throw(_)         =>
```
kevinoliver: lets leave these hashrockets unaligned

finagle-http/src/main/scala/com/twitter/finagle/http/filter/StatsFilter.scala
```diff
@@ -1,9 +1,29 @@
...
+
+object StatsFilter {
+  val role = Stack.Role("HttpStatsFilter")
+
+  def module[Req <: Request]: Stackable[ServiceFactory[Req, Response]] =
+    new Stack.Module1[param.Stats, ServiceFactory[Req, Response]] {
+      val role = StatsFilter.role
+      val description = "Register HTTP Stats"
+
+      override def make(statsParam: param.Stats,
```
kevinoliver: omit `override`.

finagle-http/src/main/scala/com/twitter/finagle/http/filter/StatsFilter.scala
```diff
@@ -1,9 +1,29 @@
...
+                        next: ServiceFactory[Req, Response]): ServiceFactory[Req, Response] = {
```
mosesn: just two spaces please.

finagle-http/src/main/scala/com/twitter/finagle/http/filter/StatsFilter.scala
```diff
@@ -1,9 +1,29 @@
...
+        if (statsParam.statsReceiver.isNull) {
```
mosesn: we typically drop braces for single exprs.

finagle-http/src/main/scala/com/twitter/finagle/Http.scala
```diff
@@ -404,6 +411,12 @@ object Http extends Client[Request, Response] with HttpRichClient
     def withMaxInitialLineSize(size: StorageUnit): Server =
       configured(param.MaxInitialLineSize(size))
 
+    /**
+     * Enable the collection of HTTP specific metrics. See [[http.filter.StatsFilter]].
+     */
+    def enableHttpStats(): Server =
+      withStack(http.filter.StatsFilter.module +: stack)
```
mosesn: I don't think we want to have non-idempotent configuration methods on StackClients.  How about instead inserting a nop "http stats" module into the Http stack and then replacing it here?

It might also be useful to look at our design principles for configuration methods for StackClients.

http://twitter.github.io/finagle/guide/Configuration.html#design-principles

spockz: Good point. I'll also have a look at those principles.

finagle-http/src/main/scala/com/twitter/finagle/Http.scala
spockz: Fair enough. I choose enable because there are no arguments passed in. Just like noFailFast.

vkostyukov: There is `noFailFast` on `Client/Server` (it's a client builder thing). We use `with`-prefixed methods here instead. And it's totally fine to have `with`-API w/o an argument (we do that all the time). See [Design Principles for the `with`-API](http://twitter.github.io/finagle/guide/Configuration.html#design-principles) (item number 2).

vkostyukov: Also, please no `()`.

finagle-http/src/main/scala/com/twitter/finagle/http/filter/StatsFilter.scala
```diff
@@ -1,9 +1,29 @@
...
+import com.twitter.finagle._
 import com.twitter.finagle.http.{Request, Response, Status}
-import com.twitter.util.{Duration, Future, Return, Stopwatch, Throw}
+import com.twitter.finagle.stats.StatsReceiver
+import com.twitter.util._
+
+object StatsFilter {
+  val role = Stack.Role("HttpStatsFilter")
+
+  def module[Req <: Request]: Stackable[ServiceFactory[Req, Response]] =
```
vkostyukov: I honestly don't think we still need this `Req <: Request` type-param. *I think* we've done this before to support something we called "rich http" (which is now just `finagle-http`). Please, correct me if I'm wrong. /cc @mosesn @dschobel @roanta

mosesn: @vkostyukov I think you're right we don't need it here.

finagle-http/src/main/scala/com/twitter/finagle/http/filter/StatsFilter.scala
spockz: There are other descriptions of modules containing verbs. Why should it be gone here?

finagle-http/src/main/scala/com/twitter/finagle/http/filter/StatsFilter.scala
spockz: I've put the type-param in there because the filter itself has it as well and limiting it in the module seemed strange. Additionally, our applications use types that inherit from `RequestProxy` so we actually like the presence of the type-param in the filter and think the api would be less rich and flexible without it.

finagle-http/src/main/scala/com/twitter/finagle/Http.scala
spockz: Done

finagle-http/src/main/scala/com/twitter/finagle/http/filter/StatsFilter.scala
spockz: Done

finagle-http/src/main/scala/com/twitter/finagle/http/filter/StatsFilter.scala
spockz: Done

finagle-http/src/main/scala/com/twitter/finagle/http/filter/StatsFilter.scala
spockz: Done

finagle-http/src/main/scala/com/twitter/finagle/http/filter/StatsFilter.scala
spockz: Done

finagle-http/src/main/scala/com/twitter/finagle/http/filter/StatsFilter.scala
kevinoliver: i guess we aren't consistent. lets leave as is.

doc/src/sphinx/Metrics.rst
```diff
@@ -195,6 +195,21 @@ These stats pertain to the HTTP protocol.
   A counter of the number of non-retryable HTTP 503 responses the Http server returns. Those
   responses are not automatically retried.
 
+**status/<statusCode>**
```
kevinoliver: lets add a sub heading explaining that they come from this Filter.

doc/src/sphinx/Metrics.rst
```diff
@@ -195,6 +195,21 @@ These stats pertain to the HTTP protocol.
...
+  A counter of the number of responses received, or returned for servers, that had this
+  statusCode.
+
+**status/<statusClass>**
+  Same as **status/statusCode** but aggregated per category, e.g. all 500 range responses
+  count as 5XX for this counter.
+
+**time/<statusCode>**
+  Metric on duration per Http status code.
```
kevinoliver: for consistency with the rest of the file:
s/Metric/A histogram/

doc/src/sphinx/Metrics.rst
```diff
@@ -195,6 +195,21 @@ These stats pertain to the HTTP protocol.
...
+
+**time/<statusCategory>**
+  Metric on duration per Http status code category.
```
kevinoliver: ditto

finagle-core/src/main/scala/com/twitter/finagle/Stack.scala
```diff
@@ -455,6 +455,11 @@ object Stack {
       Node(this, (prms, next) => Leaf(this, make(next.make(prms))), next)
   }
 
+  class NoOpModule[T](val role: Role, val description: String) extends Module0[T] {
+    override def make(next: T): T =
```
kevinoliver: nit: we omit `override` for abstract methods

finagle-http/src/main/scala/com/twitter/finagle/Http.scala
```diff
@@ -268,6 +269,13 @@ object Http extends Client[Request, Response] with HttpRichClient
     def withCompressionLevel(level: Int): Client =
       configured(param.CompressionLevel(level))
 
+
+    /**
+     * Enable the collection of HTTP specific metrics. See [[http.filter.StatsFilter]].
+     */
+    def withHttpStats: Client =
```
kevinoliver: lets add a basic unit test that verifies these work as expected. something like a request populates at least one of the stats. same for the server side.

spockz:
So, I've added a test-case. It also shows that the stats are directly under the client-label. Should we scope it to http? This would probably be a backwards incompatible change, unless we make it so that it is only prefixed with http when using `withHttpStats` and not when newing the `StatsFilter` manually.

doc/src/sphinx/Metrics.rst
```diff
@@ -204,10 +208,10 @@ These stats pertain to the HTTP protocol.
   count as 5XX for this counter.
 
 **time/<statusCode>**
-  Metric on duration per Http status code.
+  A histogram on duration per Http status code.
```
kevinoliver: sorry, should've noticed previously, please add that these two histograms are in milliseconds.

doc/src/sphinx/Metrics.rst
```diff
@@ -195,6 +195,25 @@ These stats pertain to the HTTP protocol.
...
+**status/<statusCode>**
+  A counter of the number of responses received, or returned for servers, that had this
+  statusCode.
+
+**status/<statusClass>**
+  Same as **status/statusCode** but aggregated per category, e.g. all 500 range responses
+  count as 5XX for this counter.
+
+**time/<statusCode>**
+  A histogram on duration per Http status code.
```
vkostyukov: Please, capitalize HTTP.

doc/src/sphinx/Metrics.rst
```diff
@@ -195,6 +195,25 @@ These stats pertain to the HTTP protocol.
...
+
+**time/<statusCategory>**
+  A histogram on duration per Http status code category.
```
vkostyukov: Please, capitalize HTTP.

finagle-core/src/main/scala/com/twitter/finagle/Stack.scala
```diff
@@ -455,6 +455,11 @@ object Stack {
       Node(this, (prms, next) => Leaf(this, make(next.make(prms))), next)
   }
 
+  class NoOpModule[T](val role: Role, val description: String) extends Module0[T] {
```
vkostyukov: I think you can avoid defining a new type by just passing an `identity` function as a stack module - it should be implicitly converted. We do that in [a bunch of places](https://github.com/twitter/finagle/blob/develop/finagle-core/src/main/scala/com/twitter/finagle/client/StackClient.scala#L70).

finagle-http/src/main/scala/com/twitter/finagle/http/filter/StatsFilter.scala
```diff
@@ -1,9 +1,29 @@
...
 import com.twitter.finagle.http.{Request, Response, Status}
-import com.twitter.util.{Duration, Future, Return, Stopwatch, Throw}
+import com.twitter.finagle.stats.StatsReceiver
+import com.twitter.util._
+
+object StatsFilter {
+  val role: Stack.Role = Stack.Role("HttpStatsFilter")
+  val description: String = "HTTP Stats"
+
+  def module[Req <: Request]: Stackable[ServiceFactory[Req, Response]] =
```
vkostyukov: I still don't think you need this type-param.

Please, note that it has nothing to do with your custom HTTP types (either for request or response) since at that point they should be lifted to something Finagle can work with (i.e., `c.t.f.http.Request` and `c.t.f.http.Response`).

By the same reason, `StatsFilter` doesn't need a type param `Req`, but we can't remove it yet (an API break) so let's keep it (filter) as is, but remove the type-param on a module.

spockz: Ok. I'll remove it.

spockz:
Should we scope all the stats to start with 'http'? Otherwise we clutter the stats a bit, this will give nice grouping.

finagle-core/src/main/scala/com/twitter/finagle/Stack.scala
vkostyukov: *I think* you should be fine. There is an implicit conversion from `A => A` function to `Stackable[A]`.

vkostyukov:
@spockz I don't know about scoping. On one hand, it's reasonable and that's what we do for `mux` (having a `mux` scope). On the other hand, we don't scope for HTTP and we already have some HTTP-specific metrics. So I prefer us to be consistent and either

1. Do not scope new metrics
2. Scope new metrics and move old metrics under `http/`

It's quite clear that it's hard to go the second road here since it breaks the API (will likely break both internal and external dashboards). So I vote for 1.

kevinoliver:
@vkostyukov "It's quite clear that it's hard to go the second road here since it breaks the API (will likely break both internal and external dashboards). So I vote for 1."

This could be done for only users of the module which is new and therefore not a breaking change.

I kinda like the scoping under http, but don't feel strongly.

doc/src/sphinx/Metrics.rst
spockz: Done

doc/src/sphinx/Metrics.rst
spockz: Done

finagle-core/src/main/scala/com/twitter/finagle/Stack.scala
spockz: Done

finagle-http/src/main/scala/com/twitter/finagle/Http.scala
spockz: Done

finagle-core/src/main/scala/com/twitter/finagle/Stack.scala
spockz: I'm not seeing it. How would we assign a role to the identity function?

spockz:
I went the route @kevinoliver mentioned. The `http.filter.StatsFilter` itself doesn't add the http scope but the module does.

kevinoliver:
lgtm!

spockz:
@kevinoliver, do you have any news on this?

mosesn:
@spockz we're still working on merging it internally.  We're hoping to get it in tomorrow.  Thanks for checking in!

jcrossley:
Hey @spockz ! I tried to merge in your request but  ran into a couple small issues. Fixing those up now and will have it in by the end of the day, thanks!

jcrossley:
All done, @spockz ! This has been merged internally and will go out with the next release :)

spockz:
Great, Thank you @jcrossley!

mosesn:
made it to develop here: https://github.com/twitter/finagle/commit/2b65e4187aacc0f7a08f04692854a8d3b1e50c31.  thanks for the PR!

PR #559, opened 2016-09-24
Implemented memoisation of counters and stats in StatsFilter.

spockz:
Problem

Outstanding todo for memoisation.

Solution

Added a Map with default value that creates the Stat/Counter on first
use.

Result

Cached Counters and Stats will be using in http.filter.StatsFilter.

### DIFF
--- a/finagle-http/src/main/scala/com/twitter/finagle/http/filter/StatsFilter.scala
+++ b/finagle-http/src/main/scala/com/twitter/finagle/http/filter/StatsFilter.scala
@@ -3,7 +3,7 @@ package com.twitter.finagle.http.filter
 import com.twitter.finagle.{Service, SimpleFilter}
 import com.twitter.finagle.stats.{Counter, Stat, StatsReceiver}
 import com.twitter.finagle.http.{Request, Response, Status}
-import com.twitter.util.{Duration, Future, Return, Stopwatch, Throw}
+import com.twitter.util.{Duration, Memoize, Future, Return, Stopwatch, Throw}
 
 /**
  * Statistic filter.
@@ -23,11 +23,11 @@ class StatsFilter[REQUEST <: Request](stats: StatsReceiver)
   private[this] val timeReceiver = stats.scope("time")
   private[this] val responseSizeStat = stats.stat("response_size")
 
-  private[this] val counterCache: Map[String, Counter] =
-    Map.empty.withDefault(statusReceiver.counter(_))
+  private[this] val counterCache: String => Counter =
+    Memoize(statusReceiver.counter(_))
 
-  private[this] val statCache: Map[String, Stat] =
-    Map.empty.withDefault(timeReceiver.stat(_))
+  private[this] val statCache: String => Stat =
+    Memoize(timeReceiver.stat(_))
 
   def apply(request: REQUEST, service: Service[REQUEST, Response]): Future[Response] = {
     val elapsed = Stopwatch.start()
--- a/finagle-http/src/main/scala/com/twitter/finagle/http/filter/StatsFilter.scala
+++ b/finagle-http/src/main/scala/com/twitter/finagle/http/filter/StatsFilter.scala
@@ -1,7 +1,7 @@
 package com.twitter.finagle.http.filter
 
 import com.twitter.finagle.{Service, SimpleFilter}
-import com.twitter.finagle.stats.StatsReceiver
+import com.twitter.finagle.stats.{Counter, Stat, StatsReceiver}
 import com.twitter.finagle.http.{Request, Response, Status}
 import com.twitter.util.{Duration, Future, Return, Stopwatch, Throw}
 
@@ -23,6 +23,12 @@ class StatsFilter[REQUEST <: Request](stats: StatsReceiver)
   private[this] val timeReceiver = stats.scope("time")
   private[this] val responseSizeStat = stats.stat("response_size")
 
+  private[this] val counterCache: Map[String, Counter] =
+    Map.empty.withDefault(statusReceiver.counter(_))
+
+  private[this] val statCache: Map[String, Stat] =
+    Map.empty.withDefault(timeReceiver.stat(_))
+
   def apply(request: REQUEST, service: Service[REQUEST, Response]): Future[Response] = {
     val elapsed = Stopwatch.start()
     val future = service(request)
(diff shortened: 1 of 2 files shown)


finagle-http/src/main/scala/com/twitter/finagle/http/filter/StatsFilter.scala
```diff
@@ -23,6 +23,12 @@ class StatsFilter[REQUEST <: Request](stats: StatsReceiver)
   private[this] val timeReceiver = stats.scope("time")
   private[this] val responseSizeStat = stats.stat("response_size")
 
+  private[this] val counterCache: Map[String, Counter] =
+    Map.empty.withDefault(statusReceiver.counter(_))
```
mosesn: this doesn't memoize the counter, it will call statusReceiver every time.  note that the underlying and created maps are both immutable.

mosesn: I don't follow.  MemoizedSR decorates another SR, right?  Wouldn't that solve the problem here?  Why not just wrap the underlying SR with a MemoizedSR in the StatsFilter?

spockz: Somehow I looked at the MetricsStatsReceiver instead of the MemoizedStatsReciever. I cannot find a MemoizedStatsReceiver in Finagle. I do find a Memoized in twitter-util, which is used by Finatra. Should we copy that solution to Finagle?

mosesn: Oh geez, it might not be open source.  Let me check if I just got the name wrong or if we haven't open sourced it yet.

kevinoliver: @mosesn you got the name wrong *and* its not OSS-ed 😭

mosesn: lol double whammy.  @spockz let's go ahead with your original plan and just use a map.

spockz: I tried using `Memoize` from twitter-util, which works. I've verified that by seeing the effect of println only once. However, I'm having issues verifying from the outside of `StatsFilter` that it actually works. `counter()` is being called twice for two requests.

kevinoliver:
lgtm.

finagle-http/src/test/scala/com/twitter/finagle/http/filter/StatsFilterTest.scala
```diff
@@ -7,12 +7,13 @@ import com.twitter.util.{Await, Future, Time}
...
 import org.scalatest.FunSuite
 import org.scalatest.junit.JUnitRunner
+import org.mockito.Mockito.{spy,verify,times}
 
 @RunWith(classOf[JUnitRunner])
 class StatsFilterTest extends FunSuite {
 
   test("increment stats") {
-    val receiver = new InMemoryStatsReceiver
+    val receiver = spy(new InMemoryStatsReceiver)
```
mosesn: instead of appending to an existing test, can you add a new test?

finagle-http/src/test/scala/com/twitter/finagle/http/filter/StatsFilterTest.scala
```diff
@@ -25,13 +26,20 @@ class StatsFilterTest extends FunSuite {
...
     }
 
-    assert(receiver.counters(Seq("status", "404")) == 1)
-    assert(receiver.counters(Seq("status", "4XX")) == 1)
-    // TODO: until we can mock stopwatches
-    //      receiver.stats(Seq("time", "404"))      must_== Seq(0.0)
-    //      receiver.stats(Seq("time", "4XX"))      must_== Seq(0.0)
-    assert(receiver.stats(Seq("response_size")) == Seq(5.0))
+    // Verify that the counters and stats were only created once
+    verify(receiver, times(1)).counter("status", "404")
```
mosesn: I think if you don't mention `times(1)` it implies exactly once.

finagle-http/src/test/scala/com/twitter/finagle/http/filter/StatsFilterTest.scala
```diff
@@ -25,13 +26,20 @@ class StatsFilterTest extends FunSuite {
...
+    verify(receiver, times(1)).counter("status", "4XX")
+    verify(receiver, times(1)).stat("time", "404")
+    verify(receiver, times(1)).stat("time", "4XX")
+    verify(receiver, times(1)).stat("response_size")
+
+    assert(receiver.counters(Seq("status", "404")) === 2)
```
mosesn: please use ==, instead of ===.  scalatest has an assert macro which does the right thing here.

finagle-http/src/test/scala/com/twitter/finagle/http/filter/StatsFilterTest.scala
spockz: Sure

finagle-http/src/test/scala/com/twitter/finagle/http/filter/StatsFilterTest.scala
spockz: Sure

spockz:
@mosesn yes, the tests pass, on my machine.

finagle-http/src/test/scala/com/twitter/finagle/http/filter/StatsFilterTest.scala
```diff
@@ -12,34 +12,46 @@ import org.mockito.Mockito.{spy,verify,times}
...
-      def apply(request: Request): Future[Response] = {
-        val response = request.response
-        response.statusCode = 404
-        response.write("hello")
-        Future.value(response)
-      }
+    val filter = new StatsFilter(receiver) andThen service
+
+    Time.withCurrentTimeFrozen { _ =>
+      Await.result(filter(Request()))
```
mosesn: feel free to ignore: please add a timeout–we no longer use `Await.result` in our tests to avoid deadlocks.

spockz: Yes I was wondering about that. What is commonly used as timeout  value?

mosesn: 5 seconds is fine.

mosesn: ah, when I wrote "we no longer use Await.result" I meant unbounded Await.result.

kevinoliver:
working on getting this merged.

kevinoliver:
Alrighty this is merged and'll show up on the `develop` branch this Monday. Thanks again!

PR #560, opened 2016-09-25
finagle-http: Added default "Host" header, for http 1.1 Requests.

vazyzy:
Problem

By HTTP 1.1 RFC  all requests must contains a Host header field.
(https://tools.ietf.org/html/rfc7230#page-44). It's not clear, when you create
request with version HTTP 1.1 and it does not satisfy protocol.

Solution

Generalize creation of Request instance and add the empty Host header for HTTP
version 1.1.


kevinoliver:
@vazyzy, yeah this is a surprise for new users and a pothole in terms of ease-of-use. 

+@vkostyukov, @bryce-anderson, @mosesn who've been talking about this and some related issues.

vkostyukov:
Thanks @vazyzy!

Yeah, we're actually thinking about it at this very moment. Besides the `Host` header, there are other places where finagle-http is not completely specs-compliant. Good news - we have some ideas in mind on how to make it better (and explicit!), but that's part of a larger discussion. One way or another, this kind of concerns should be applied generically, not matter how a request was build (using the `RequestBuilder` or just `Request`).

Historically, we used HTTP dispatcher to fill the missing gaps, but now we're thinking about separating those concerns (making a separate spec-compliant stack module).

Saint1991:
@vkostyukov 

Maybe you're discussing, I got another problem when using Finagle's load balancing feature.
In order to distribute requests to multiple servers, we instantiate a service via Http.Client#newService as follows,

`val client = Http.client.newService("hostA.org:80, hostB.org:80")`

Then construct request,
```

val request = Request("/path/to/service")
request.host = ???

What should we fill ??? with?

Moreover, we're forced to pass the host addresses twice when the service creation and the building request. It's not DRY.

Are there any reasons that the Host header automatically set that given to the Service?

File: finagle-http/src/main/scala/com/twitter/finagle/http/Request.scala [hunk 1]
```diff
@@ -287,10 +287,36 @@ object Request {
...
+    buildRequest(
+      request = reqIn,
+      socketAddress = new InetSocketAddress(0)
+    )
+  }
+
+  private def buildRequest(
+    request: HttpRequest,
+    socketAddress: InetSocketAddress,
+    reader: Option[Reader] = None): Request = {
```
bryce-anderson: Could we remove the default and be explicit? Its private code so I don't see any reason to be ergonomic about it.

finagle-http/src/main/scala/com/twitter/finagle/http/Request.scala
```diff
@@ -287,10 +287,36 @@ object Request {
...
+        }
+      case None =>
+        new  Request {
+          val httpRequest: HttpRequest = request
+          lazy val remoteSocketAddress: InetSocketAddress = socketAddress
+        }
+    }
+
+    if (req.version == Version.Http11 && !req.headers.contains(HttpHeaders.Names.HOST)) {
+      req.headers.set(HttpHeaders.Names.HOST, "")
```
bryce-anderson: This is more of a question than a concern, but is an empty `Host` header meaningful and/or legal per the spec? Adding a link to any relevant RFC would be a good touch, if it exists.

bryce-anderson:
It also looks like some of the Travis failures are legitimate and need to be addressed.

finagle-http/src/main/scala/com/twitter/finagle/http/Request.scala
vazyzy: I already added link to RFC in my commit. Here a quote from it:
```
 A client MUST send a Host header field in all HTTP/1.1 request
   messages.  If the target URI includes an authority component, then a
   client MUST send a field-value for Host that is identical to that
   authority component, excluding any userinfo subcomponent and its "@"
   delimiter (Section 2.7.1).  If the authority component is missing or
   undefined for the target URI, then a client MUST send a Host header
   field with an empty field-value.
```
I think an empty Host header during initialization is ok.
 Other way is to parse a Host from URI like you already do in `RequestBuilder`. And then we must remove this logic from `RequestBuilder` to avoid duplications.  But I did not find a way to do it neatly and without changes in public interfaces.

bryce-anderson: Thanks for the clarification; this looks correct to me. By adding a link, I meant adding it as a code comment close to this logic in case others are interested in reading the spec in the future.

finagle-http/src/test/scala/com/twitter/finagle/http/RequestTest.scala
```diff
@@ -65,6 +65,18 @@ class RequestTest extends FunSuite {
     assert(actual == expected)
   }
 
+  test("Default Host header for HTTP_1.1") {
```
bryce-anderson: Could you add a test that verifies that the Host header isn't overwritten if its already present?

bryce-anderson:
Hey @vazyzy, I was in the process of merging this internally but ran into a few unexpected issues. We have a number of internal uses that look for a Host header when forming Url's etc that end up breaking when the Host header is the empty string. In theory, this could be fixed, but it's unlikely to happen in the near future.
What is the problem you're trying to solve? As @vkostyukov mentioned above, we are starting to address theses conformance issues using filters stack modules.  See the [`ResponseConformanceFilter`](https://github.com/twitter/finagle/blob/develop/finagle-http/src/main/scala/com/twitter/finagle/http/codec/ResponseConformanceFilter.scala). Would this solve the problem you're trying to address?

bryce-anderson:
I'm going to close this since, unfortunately, this approach causes internal problems at Twitter. I think there is a way forward which has been mentioned above which is making the responsibility of request conformance a part of the client and not the request.

PR #568, opened 2016-11-08
Make WorkerPool public (again)

herberteuler:
Problem

There is an infamous epoll bug in Netty (netty/netty#663), causing a service not serving any requests eats 100% CPU. Worker pools provide a method, namely rebuildSelectors, to work around this problem. This method needs to be invoked periodically. WorkerPool being package private makes calling this method on the default worker pool impossible, it would be handy if WorkerPool is public, so that the method is easily available.

Solution

Make WorkerPool public again.

### DIFF
--- a/finagle-core/src/main/scala/com/twitter/finagle/netty3/package.scala
+++ b/finagle-core/src/main/scala/com/twitter/finagle/netty3/package.scala
@@ -33,5 +33,5 @@ package object netty3 {
     Executors.newCachedThreadPool(threadFactory)
   }
 
-  private[netty3] val WorkerPool: NioWorkerPool = new NioWorkerPool(Executor, numWorkers())
+  val WorkerPool: NioWorkerPool = new NioWorkerPool(Executor, numWorkers())
 }


mosesn:
LGTM.  @vkostyukov what do you think?  it's not great that this exposes a netty3 API, but I'm also not sure if it's worth it to build a background thread into finagle to call this for netty3 on behalf of the user.  part of it depends on how much longer we expect we'll have to support netty3.

vkostyukov:
Reposting (from Gitter) my question here.

Is it still a bug though? netty/netty#663 is closed and Trustin mentions something between the lines: “It's not a bug in Netty but a bug in JDK unfortunately. “ Maybe it's fixed already in Java 7?

herberteuler:
It is indeed not a bug of Netty, my description was not precise. Unfortunately, this issue still exists in Oracle's jdk 8u77 (the version we are using).

kevinoliver:
@vkostyukov @mosesn I think we should expose this Netty 3 API. It would go away once Netty 3 goes away.

vkostyukov:
@kevinoliver Works for me!

jcrossley:
Thanks @codecov-io ! This has been merged internally and will go out in next month's release.

PR #574, opened 2016-11-30
finagle-http: Fix SpnegoAuthenticator test failing due to missing realm

jordanly:
Problem / Solution

We create a KerberosPrincipal in SpnegoAuthenticatorTest so we can test the
validity of a ticket under that principal. KerberosPrincipal has a check to see
if a realm has been provided (an @{NAME} appended to the end). If a realm is
not provided, the constructor will look on the machine running the test to see
if a default realm has been configured. Since CI does not have a default realm
configured, the test failed. We append a realm to the end of the name provided
to KerberosPrincipal in order to avoid this exception.


kevinoliver:
thanks, this was merged internally.

Issue #544, opened 2016-08-18
InetResolver uses an unbounded thread pool


reikje:
The `InetResolver` uses an unbounded threadpool under the hood. For applications that resolve a lot of `inet!....` destinations, this creates a lot of threads which can be a problem. Unfortunately this is not at all configurable. It's also not possible to extend and create a new `InetResolver` as it is package private. 
### Expected behavior

`InetResolver` can be used with a threadpool that can be configured if needed.
### Actual behavior

`InetResolver` uses `FuturePool.unbounded`
### Steps to reproduce the behavior

Create an application that creates a few hunded `Http.client` instances using `inet! ..` destinations. Do a thread dump and look for threads named `UnboundedFuturePool-xyz`.

kevinoliver:
@edma2 or @atollena may have thoughts about this.

atollena:
The number of concurrent DNS resolutions (and thus the number of threads spawned) is bound by a semaphore to a concurrency of 100:

https://github.com/twitter/finagle/blob/develop/finagle-core/src/main/scala/com/twitter/finagle/Resolver.scala#L109

I could see value in making that more configurable, but presumably this would need to be done via a global flag, since InetResolver is a singleton.

We are exploring the use Netty 4 DNS implementation (io.netty.resolver.dns) which would make this implementation obsolete, but in the mean time you could submit a patch to tune the max concurrency level.

mosesn:
@reikje do you mind telling us about your use case?  why are you doing so many DNS resolutions?

reikje:
@mosesn it's an application that listens on a event stream that a remote server provides. on the stream we receive notifications containing host:port of slave machines (nodes being added and removed). Before we start listening we receive all current nodes in one go, around 150-200 items. For each we create a finagle.Http.client using the `inet!host:port` destination, which will internally `bind` using the `InetResolver`.

[...]

PR #579, opened 2017-01-03
InetResolver can now be created with a custom FuturePool

reikje:
Problem

The `InetResolver` by default uses a unbounded threadpool under the hood. This can be a problem for applications that want to control over number of threads. see https://github.com/twitter/finagle/issues/544

Solution

Added optional `FuturePool` argument to the InetResolver factory. Applications can now create a custom `Resolver` which uses delegation and a custom `InetResolver`.

Notes

- `FixedInetResolver` is unchanged
- I tried to extend my test in `InetResolverTest` to also cover uses of the pool in `bindHostPortsToAddr` but for some reason the update timer is never executed. I think it requires a different test setup with the `Var` and `Addr` that I would need help with.

### DIFF
--- a/finagle-core/src/test/scala/com/twitter/finagle/InetResolverTest.scala
+++ b/finagle-core/src/test/scala/com/twitter/finagle/InetResolverTest.scala
@@ -108,7 +108,6 @@ class InetResolverTest extends FunSuite {
         assert(b.contains(Address("127.0.0.1", 80)))
       case _ => fail()
     }
-
     assert(latch.await(maxWaitTimeout))
   }
 }
--- a/finagle-core/src/test/scala/com/twitter/finagle/InetResolverTest.scala
+++ b/finagle-core/src/test/scala/com/twitter/finagle/InetResolverTest.scala
@@ -109,7 +109,5 @@ class InetResolverTest extends FunSuite {
     }
 
     assert(latch.await(pollInterval * 2))
-
   }
-
 }
--- a/finagle-core/src/test/scala/com/twitter/finagle/InetResolverTest.scala
+++ b/finagle-core/src/test/scala/com/twitter/finagle/InetResolverTest.scala
@@ -101,17 +101,14 @@ class InetResolverTest extends FunSuite {
     val inetResolverWithPool = new InetResolver(resolveLoopback, DefaultStatsReceiver, Some(pollInterval), resolvePool)
 
     val maxWaitTimeout = Duration(500, TimeUnit.MILLISECONDS)
-    Time.withCurrentTimeFrozen { timeControl =>
-      val addr = inetResolverWithPool.bind("127.0.0.1:80")
-      val f = addr.changes.filter(_ != Addr.Pending).toFuture
-      Await.result(f, maxWaitTimeout) match {
-        case Addr.Bound(b, meta) if meta.isEmpty =>
-          assert(b.contains(Address("127.0.0.1", 80)))
-        case _ => fail()
-      }
-
-      timeControl.advance(pollInterval)
-      assert(latch.await(maxWaitTimeout))
+    val addr = inetResolverWithPool.bind("127.0.0.1:80")
+    val f = addr.changes.filter(_ != Addr.Pending).toFuture
+    Await.result(f, maxWaitTimeout) match {
+      case Addr.Bound(b, meta) if meta.isEmpty =>
+        assert(b.contains(Address("127.0.0.1", 80)))
+      case _ => fail()
     }
+
+    assert(latch.await(maxWaitTimeout))
   }
 }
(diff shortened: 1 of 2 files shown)


finagle-core/src/test/scala/com/twitter/finagle/InetResolverTest.scala
```diff
@@ -76,4 +78,36 @@ class InetResolverTest extends FunSuite {
...
+      if (host.equals("127.0.0.1")) dnsResolverWithPool(host)
+      else Future.exception(new UnknownHostException())
+    }
+
+    val pollInterval = Duration(100, TimeUnit.MILLISECONDS)
+    val inetResolverWithPool = new InetResolver(resolveLoopback, DefaultStatsReceiver, Some(pollInterval), resolvePool)
+
+    val addr = inetResolverWithPool.bind("127.0.0.1:80")
+    val f = addr.changes.filter(_ != Addr.Pending).toFuture
+    Await.result(f) match {
```
bryce-anderson: I know this spec is inconsistent in this regard, but could you add a timeout to this `Await.result` call?

finagle-core/src/test/scala/com/twitter/finagle/InetResolverTest.scala
```diff
@@ -76,4 +78,36 @@ class InetResolverTest extends FunSuite {
...
+    val resolvePool = new TestPool(latch)
+
+    val dnsResolverWithPool = new DnsResolver(DefaultStatsReceiver, resolvePool)
+
+    def resolveLoopback(host: String): Future[Seq[InetAddress]] = {
+      if (host.equals("127.0.0.1")) dnsResolverWithPool(host)
+      else Future.exception(new UnknownHostException())
+    }
+
+    val pollInterval = Duration(100, TimeUnit.MILLISECONDS)
```
bryce-anderson: The timing here is a bit tight and might introduce some flakyness in this test case. Have you seen our `Time` api that allows you to manipulate the current time? See `Time.withCurrentTimeFrozen` etc. It's what we commonly use in finagle tests to introduce some determanistic behavior for things that deal in `Time` and seems like it should be useful here.

finagle-core/src/test/scala/com/twitter/finagle/InetResolverTest.scala
reikje: fixed

finagle-core/src/test/scala/com/twitter/finagle/InetResolverTest.scala
reikje: Yes I have seen it, good idea! However, with the current test setup, advancing time isn't really required as `inetResolverWithPool.bind` will count down the latch one time immediately. In other words `pollInterval` isn't really used in the test. 

What I tried to do, is to test this [section](https://github.com/twitter/finagle/blob/develop/finagle-core/src/main/scala/com/twitter/finagle/InetResolver.scala#L124-L135) which uses `pollInterval`. However, I didn't manage to have the `timer` execute in `bindHostPortsToAddr`. I am not an expert in `Var` and `Addr` but my suspicion is that this test setup bypasses the timer:

    val addr = inetResolverWithPool.bind("127.0.0.1:80")
    val f = addr.changes.filter(_ != Addr.Pending).toFuture

I would be happy if you could suggest a different test setup, so that the timer is executed. In that case we could have a `CountDownLatch(2)` and put `Time.withCurrentTimeFrozen` to use and as a bonus bump test coverage :)

bryce-anderson: I think the issue there is that `InetResolver` doesn't have a way to pass in a `Timer` and thus cannot use a `MockTimer`. It could be added as another constructor param, though that might not be worth it.

reikje: True, so let's skip that test then. Out of curiosity, if I set the latch to `2` (remove the timeout in `await`) and send in a `pollInterval`, why does that never poll for updates?

bryce-anderson: Since we're not actually using the magic time features, maybe the `Time.withCurrentTimeFrozen(` isn't necessary. Sorry for the misdirection.

Regarding the latch, the `addr.changes.toFuture` call causes the polling event to get canceled after the `Future` resolves, so it never ticks a second time.  Do something to the tune of
```scala
addr.changes.filter(_ != Addr.Pending).respond(addr => println(s"Address: $addr"))
```
adjust your latch to 2, and the test should complete.

reikje: Removed `Time.withCurrentTimeFrozen`. Thanks for the hint regarding `addr.changes`.

reikje: @bryce-anderson changes are made

bryce-anderson:
Good news @reikje, this has been merged internally and has now landed as commit 9df664a. Thanks for the PR!

PR #584, opened 2017-02-12
Use RetryExceptionsFilter in docs

spockz:
Problem

The docs still use the RetryFilter which can't be constructed with a
RetryPolicy[Try[Rep]] but takes a RetryPolicy[(Req, Try[Rep])] instead.

Solution

Use the RetryExceptionsFilter in the example in order to keep the
semantics the same.


mosesn:
Instead of changing RetryFilter to RetryExceptionsFilter, can you instead change the type parameter of RetryPolicy?

mosesn:
@spockz have you had a chance to take another look at this?

nepthar:
@spockz I'm going to close this for bookkeeping reasons. If you feel like working on it again, I'd be more than happy to re-open.

PR #587, opened 2017-02-22
finagle-redis: Replace PATTERN with MATCH in SCAN and HSCAN

mkhq:
Problem

HSCAN and SCAN commands take an optional argument for pattern matching.
This argument is called MATCH and not PATTERN.

Solution

Replace PATTERN with MATCH in SCAN and HSCAN.

Result

Redis clients can execute SCAN and HSCAN with an optional pattern match
argument.

### DIFF
--- a/finagle-redis/src/test/scala/com/twitter/finagle/redis/commands/key/KeyClientIntegrationSuite.scala
+++ b/finagle-redis/src/test/scala/com/twitter/finagle/redis/commands/key/KeyClientIntegrationSuite.scala
@@ -1,9 +1,10 @@
 package com.twitter.finagle.redis.integration
 
+import com.twitter.conversions.time._
 import com.twitter.finagle.redis.RedisClientTest
 import com.twitter.finagle.redis.tags.{RedisTest, ClientTest}
 import com.twitter.io.Buf
-import com.twitter.util.Await
+import com.twitter.util.{Await, Future}
 import com.twitter.finagle.redis.util.BufToString
 import java.util.Arrays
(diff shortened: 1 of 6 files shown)


finagle-redis/src/test/scala/com/twitter/finagle/redis/commands/key/KeyClientIntegrationSuite.scala
```diff
@@ -14,11 +15,13 @@ import org.scalatest.junit.JUnitRunner
 @RunWith(classOf[JUnitRunner])
 final class KeyClientIntegrationSuite extends RedisClientTest {
 
+  val TIMEOUT = 5.seconds
```
mosesn: we use CamelCase for constants, can you change this to Timeout?  an alternative would be to add an `await` method:

```scala
def await[A](a: Future[A]): A = Await.result(a, 5.seconds)
```

mkhq:
@mosesn updated and tried the await function approach, works fine.

mosesn:
Thanks @mkhq!

ryanoneill:
Thank you for the contribution @mkhq. I've put this up for review internally. Once it's submitted internally, it will get pushed back out to GitHub on the develop branch sometime soon after. At that point I'll close this PR.

ryanoneill:
Thanks again for the contribution @mkhq. This has been pushed back out to GitHub in commit https://github.com/twitter/finagle/commit/894a7b04a49c9e3698156281944747513fb3bec6

PR #589, opened 2017-03-02
ZooKeeper exists watch now ignores session state events

simontoens:
Hi all, this regarding this oom: https://groups.google.com/forum/#!topic/finaglers/_wSvPz0YKjE

This fix may be too naive, but I couldn't see any other way of handling this problem given how the existsWatch method is called.  Most of all I'd like to get the discussion going on how we can get this ApacheWatcher leak fixed.

========

Problem

When a ZooKeeper "exists" watch is triggered because the path being
watched is created or deleted, ZooKeeper removes that watch, and it has
to be re-added in order to receive further notifications about the
watched path.

However the exists watch also gets triggered for ZooKeeper session events
(connected, disconnected etc) - for those events, ZooKeeper does not remove
the watch, so it doesn't have to be re-added.

Currently the code that re-adds the watch does not distinguish between
event types and always re-adds a new watch, causing an ApacheWatcher
memory leak each time the watch is triggered for a session event type.

Solution

Filter out session event types when watching for path existence

### DIFF
--- a/finagle-serversets/src/main/scala/com/twitter/finagle/serverset2/client/apache/ApacheWatcher.scala
+++ b/finagle-serversets/src/main/scala/com/twitter/finagle/serverset2/client/apache/ApacheWatcher.scala
@@ -5,16 +5,18 @@ import com.twitter.finagle.stats.{NullStatsReceiver, StatsReceiver}
 import com.twitter.util.Var
 import org.apache.zookeeper.{Watcher, WatchedEvent}
 
-private[serverset2] class ApacheWatcher(statsIn: StatsReceiver = NullStatsReceiver)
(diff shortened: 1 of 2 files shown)


kevinoliver:
Thanks @simontoens I forwarded it along to folks who should have more knowledge in this area

simontoens:
Closing this for now until we have a more concrete plan to fix something here.

ernstae:
This issue can creep up when there's an underlying network problem. Is there any interest in rekindling this PR to avoid stale ApacheWatcher objects?

kevinoliver:
@ernstae i think so, if you can drive it forward.

PR #591, opened 2017-03-07
Rename pending-commands prop to link-pending-commands

tomas-edwardsson:
The sentinel properties do not include `pending-commands`, updated to the
actual property `link-pending-commands`.

Problem

When initializing the `SentinelClient` a invalid property exception is thrown
on the `pending-commands` property.

Solution

Renamed to `link-pending-commands`

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -109,6 +109,10 @@ Breaking API Changes
     to `ThriftMethod.SuccessType` instead of `ThriftMethod.Args` to `ThriftMethod.Result`.
     ``RB_ID=908846``
 
+  * finagle-redis: Remove pendingCommands from `c.t.f.finagle.redis.SentinelClient.Node` and
+    add linkPendingCommands for compatibility with redis 3.2 and newer.
+    ``RB_ID=??????``
+
 Runtime Behavior Changes
 ~~~~~~~~~~~~~~~~~~~~~~~~
 
@@ -151,6 +155,9 @@ Bug Fixes
 
   * finagle-thrift: Properly locate sub-classed MethodIface services to instantiate for serving
     BaseServiceIface implemented thrift services. ``RB_ID=907608``
+
+  * finagle-redis: The SentinelClient will no longer throw an NoSuchElementException when
+    initializing connections to a redis 3.2 or greater sentinel server. ``RB_ID=??????``
 
 Dependencies
 ~~~~~~~~~~~~
--- a/finagle-redis/src/main/scala/com/twitter/finagle/redis/SentinelClient.scala
+++ b/finagle-redis/src/main/scala/com/twitter/finagle/redis/SentinelClient.scala
@@ -26,7 +26,7 @@ object SentinelClient {
     val port: Int = props("port").toInt
     val runid: String = props("runid")
     val flags: Seq[String] = props("flags").split(",")
-    val pendingCommands: Int = props("pending-commands").toInt
+    val linkPendingCommands: Int = props("link-pending-commands").toInt
     val lastPingSent: Int = props("last-ping-sent").toInt
     val lastPingReply: Int = props("last-ping-reply").toInt
     val downAfterMilliseconds: Int = props("down-after-milliseconds").toInt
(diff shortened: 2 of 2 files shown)


mosesn:
Thanks for the PR!  From googling around, it looks like there might have been a change in behavior at some point in time: https://redis.io/topics/sentinel-old.  Do you know if there's a way that we can support both the old behavior and the new behavior?  Sorry, I'm not familiar with the way this API has changed, so you'll have to bear with me.

mosesn:
@thirstycrow you might be interested in this PR.

tomas-edwardsson:
Not sure if considered dirty but we could go with the following:
```
import scala.util.Try
```

```
    val pendingCommands: Int =
      Try(props("pending-commands").toInt).getOrElse(-1)
    val linkPendingCommands: Int =
      Try(props("link-pending-commands").toInt).getOrElse(-1)
```

I would prefer changing this to `Option[Int]` but that would be backwards breaking.

This way the types would match but the value would be -1. Thoughts?

mosesn:
OK, so if I understand this correctly, the server sends the client a list of properties, which used to include  "pending-commands" but now includes "link-pending-commands".  Is that right?  Maybe we should use the name "link-pending-commands" but check for properties by both names, and use whichever we find first (starting with link-pending-commands).  what do you think?  that might look like:

val linkedPendingCommands = `props.get("link-pending-commands").getOrElse(props.get("pending-commands")).toInt`

finagle-redis/src/main/scala/com/twitter/finagle/redis/SentinelClient.scala
```diff
@@ -26,7 +26,8 @@ object SentinelClient {
     val port: Int = props("port").toInt
     val runid: String = props("runid")
     val flags: Seq[String] = props("flags").split(",")
-    val linkPendingCommands: Int = props("link-pending-commands").toInt
+    val pendingCommands: Int = props.get("link-pending-commands")
```
mosesn: can you add a comment that explains why we do this?

finagle-redis/src/main/scala/com/twitter/finagle/redis/SentinelClient.scala
```diff
@@ -26,7 +26,8 @@ object SentinelClient {
     val port: Int = props("port").toInt
     val runid: String = props("runid")
     val flags: Seq[String] = props("flags").split(",")
-    val pendingCommands: Int = props("pending-commands").toInt
+    val pendingCommands: Int = props.get("link-pending-commands")
```
mosesn: also, maybe we should call this `linkPendingCommands`  since that's its name going forward?  should be OK to break the API if you add a note to the CHANGES file.

mosesn:
This looks good, I think we can merge this soon.  I wonder what other clients do.  Do you know?

tomas-edwardsson:
Updated CHANGES file, not sure where the RB_IDs come from so left them as `??????`. I agree renaming to `linkPendingCommands` to better reflect the actual value returned by redis.

CHANGES
```diff
@@ -152,6 +156,9 @@ Bug Fixes
   * finagle-thrift: Properly locate sub-classed MethodIface services to instantiate for serving
     BaseServiceIface implemented thrift services. ``RB_ID=907608``
 
+  * finagle-redis: The SentinelClient will no longer throw an NoSuchElementException when
```
mosesn: can you also add a note that this is a breaking API change?  we've renamed pendingCommands to linkPendingCommands.

tomas-edwardsson: Already noted above in Line 112 to 115 under Breaking Changes

mosesn:
ah, good point!  works for me.

CHANGES
mosesn: ah, good point!  works for me.

vkostyukov:
This was merged in (https://github.com/twitter/finagle/commit/e89eacba5471456a86db3177a2540c85424c08bc). Thanks @tomas-edwardsson!

PR #592, opened 2017-03-17
Make peer certificate available to service factory

koshelev:
Problem

Transport's peer certificate is not accessible from a service factory. In some cases a certificate or information if client has presented a certificate can be useful at service construction time.

Solution

Change StdStackService to set current TLS peer certificate context if peer certificate is available in transport when calling a service factory.


kevinoliver:
lgtm too. thanks for the patch, @koshelev

bryce-anderson:
This has been merged as 9fa0dafb59080f5e600e0d00f3012a8ac66a5216. Thanks for the PR @koshelev!

koshelev:
@bryce-anderson @kevinoliver Thanks for reviewing!

Issue #594, opened 2017-03-22
pending Metric in StatsFilter can be incorrect


reikje:
The `pending` metric in the `StatsFilter` can be incorrect.

### Expected behavior

The metric should always contain the correct number of outstanding requests.

### Actual behavior

If `service(request)` is throwing an exception, the count is never decremented.

### Steps to reproduce the behavior

Create a `Server` that is throwing an `Exception` without wrapping it in a `Future`.

mosesn:
Thanks for the bug report!  My guess would be that we should wrap `service(request)` in a try/catch block in https://github.com/twitter/finagle/blob/develop/finagle-core/src/main/scala/com/twitter/finagle/service/StatsFilter.scala.  @reikje do you want to take a stab at this?

reikje:
Yes will do @mosesn :)

reikje:
Created a pull request: https://github.com/twitter/finagle/pull/595 and also added some notes.

vkostyukov:
Fixed by #595.

PR #595, opened 2017-03-23
StatsFilter decrements pending metric also on exceptions

reikje:
Problem

I have seen in our graphs that the `pending` metric in the `StatsFilter` slowly but steadily increases over time. When `service(request)` in `StatsFilter.apply` throws an uncaught exception, the gauge isn't decremented. 

see https://github.com/twitter/finagle/issues/594

Solution

Wrap the `service(request)` call in a `try-catch` and decrement before re-throwing the original exception.

Result

`pending` gauge is properly decremented. 

**Note** if you compose the `StatsFilter` with a `Service` using `andThen` (which `StatsFilterTest` does), this will use `Service.rescue` under the hood - wrapping all non fatal exceptions in a `Future.exception`. This is the reason the unit test throws a fatal exception to verify the fix. However, if I hit this problem in my application, this must mean that either my service is not wrapped  using `Service.rescue` or I am hitting a fatal exception that doesn't bring my JVM down?! I am using Finagle 6.42.0 and this is how the service is created:

```
val thriftService = new FooService()

ThriftMux.server
      .configured(param.Label("thrift"))
      .serveIface(new InetSocketAddress(10200), thriftService)
```

### DIFF
--- a/finagle-core/src/main/scala/com/twitter/finagle/service/StatsFilter.scala
+++ b/finagle-core/src/main/scala/com/twitter/finagle/service/StatsFilter.scala
@@ -1,13 +1,11 @@
 package com.twitter.finagle.service
 
-import java.util.concurrent.atomic.LongAdder
-import java.util.concurrent.TimeUnit
-
 import com.twitter.finagle.Filter.TypeAgnostic
 import com.twitter.finagle._
 import com.twitter.finagle.stats.{ExceptionStatsHandler, MultiCategorizingExceptionStatsHandler, StatsReceiver}
 import com.twitter.util._
-
+import java.util.concurrent.atomic.LongAdder
+import java.util.concurrent.TimeUnit
 import scala.util.control.NonFatal
 
 object StatsFilter {
--- a/finagle-core/src/main/scala/com/twitter/finagle/service/StatsFilter.scala
+++ b/finagle-core/src/main/scala/com/twitter/finagle/service/StatsFilter.scala
@@ -8,6 +8,8 @@ import com.twitter.finagle._
 import com.twitter.finagle.stats.{ExceptionStatsHandler, MultiCategorizingExceptionStatsHandler, StatsReceiver}
 import com.twitter.util._
 
+import scala.util.control.NonFatal
+
 object StatsFilter {
   val role: Stack.Role = Stack.Role("RequestStats")
 
@@ -160,33 +162,33 @@ class StatsFilter[Req, Rep](
 
     outstandingRequestCount.increment()
 
-    try {
-      service(request).respond { response =>
-        outstandingRequestCount.decrement()
-        if (!isBlackholeResponse(response)) {
-          dispatchCount.incr()
-          responseClassifier.applyOrElse(
-            ReqRep(request, response),
-            ResponseClassifier.Default
-          ) match {
-            case ResponseClass.Failed(_) =>
-              latencyStat.add(elapsed().inUnit(timeUnit))
-              response match {
-                case Throw(e) =>
-                  exceptionStatsHandler.record(statsReceiver, e)
-                case _ =>
-                  exceptionStatsHandler.record(statsReceiver, SyntheticException)
-              }
-            case ResponseClass.Successful(_) =>
-              successCount.incr()
-              latencyStat.add(elapsed().inUnit(timeUnit))
-          }
+    val result = try {
+      service(request)
+    } catch { case NonFatal(e) =>
+      Future.exception(e)
+    }
+
+    result.respond { response =>
+      outstandingRequestCount.decrement()
+      if (!isBlackholeResponse(response)) {
+        dispatchCount.incr()
+        responseClassifier.applyOrElse(
+          ReqRep(request, response),
+          ResponseClassifier.Default
+        ) match {
+          case ResponseClass.Failed(_) =>
+            latencyStat.add(elapsed().inUnit(timeUnit))
+            response match {
+              case Throw(e) =>
+                exceptionStatsHandler.record(statsReceiver, e)
+              case _ =>
+                exceptionStatsHandler.record(statsReceiver, SyntheticException)
+            }
+          case ResponseClass.Successful(_) =>
+            successCount.incr()
+            latencyStat.add(elapsed().inUnit(timeUnit))
         }
       }
-    } catch {
-      case e: Throwable =>
-        outstandingRequestCount.decrement()
-        throw e
     }
   }
 }
--- a/finagle-core/src/main/scala/com/twitter/finagle/service/StatsFilter.scala
+++ b/finagle-core/src/main/scala/com/twitter/finagle/service/StatsFilter.scala
@@ -1,7 +1,7 @@
 package com.twitter.finagle.service
 
-import java.util.concurrent.TimeUnit
 import java.util.concurrent.atomic.LongAdder
+import java.util.concurrent.TimeUnit
 
 import com.twitter.finagle.Filter.TypeAgnostic
 import com.twitter.finagle._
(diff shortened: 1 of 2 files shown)


finagle-core/src/main/scala/com/twitter/finagle/service/StatsFilter.scala
```diff
@@ -1,11 +1,12 @@
 package com.twitter.finagle.service
 
+import java.util.concurrent.TimeUnit
+import java.util.concurrent.atomic.LongAdder
```
mosesn: please keep all of the imports sorted alphabetically

finagle-core/src/main/scala/com/twitter/finagle/service/StatsFilter.scala
```diff
@@ -158,27 +159,34 @@ class StatsFilter[Req, Rep](
...
-              case Throw(e) =>
-                exceptionStatsHandler.record(statsReceiver, e)
-              case _ =>
-                exceptionStatsHandler.record(statsReceiver, SyntheticException)
-            }
-          case ResponseClass.Successful(_) =>
-            successCount.incr()
-            latencyStat.add(elapsed().inUnit(timeUnit))
+
+    try {
```
mosesn: let's scope the try/catch block more narrowly.  I think we want to be able to reuse the behavior from the rest of the method, so how about we convert a `service(request)` that throws an exception into one that returns a failed future?  we could make this instead be

```scala
val result = try {
  service(request)
}  catch { case NonFatal(e) => 
  Future.exception(e)
}

result.respond { ...
```

what do you think?

finagle-core/src/test/scala/com/twitter/finagle/service/StatsFilterTest.scala
```diff
@@ -5,22 +5,37 @@ import com.twitter.finagle.stats.{CategorizingExceptionStatsHandler, ExceptionSt
...
 import org.scalatest.FunSuite
 import org.scalatest.junit.JUnitRunner
 
+import scala.util.control.ControlThrowable
+
 @RunWith(classOf[JUnitRunner])
 class StatsFilterTest extends FunSuite {
   val BasicExceptions = new CategorizingExceptionStatsHandler(_ => None, _ => None, rollup = false)
 
+  sealed trait ServiceBehaviour
```
mosesn: instead of adding all of this extra machinery to the `getService` method, we should just inline a new service.

finagle-core/src/test/scala/com/twitter/finagle/service/StatsFilterTest.scala
```diff
@@ -141,6 +156,17 @@ class StatsFilterTest extends FunSuite {
     assert(receiver.gauges(Seq("pending"))() == 0.0)
   }
 
+  test("don't report pending requests after uncaught exceptions") {
+    val (_, receiver, statsService) = getService(behaviour = Fail)
+    assert(receiver.gauges(Seq("pending"))() == 0.0)
+
+    intercept[ControlThrowable] {
+      statsService("foo")
```
mosesn: it would be nice to validate that pending actually increments.  how about adding a new filter in between StatsFilter and the service that throws the exception that checks that pending is 1.0?

finagle-core/src/main/scala/com/twitter/finagle/service/StatsFilter.scala
```diff
@@ -1,11 +1,14 @@
 package com.twitter.finagle.service
 
+import java.util.concurrent.atomic.LongAdder
+import java.util.concurrent.TimeUnit
```
mosesn: sorry, do you mind sorting these with everything else?  as in, com comes before java, java comes before scala, etc.

finagle-core/src/test/scala/com/twitter/finagle/service/StatsFilterTest.scala
```diff
@@ -141,6 +144,21 @@ class StatsFilterTest extends FunSuite {
     assert(receiver.gauges(Seq("pending"))() == 0.0)
   }
 
+  test("don't report pending requests after uncaught exceptions") {
+    val receiver = new InMemoryStatsReceiver()
+    val service = new Service[String, String] {
+      def apply(request: String): Future[String] = throw new Exception("broken")
+    }
+    val statsFilter = new StatsFilter[String, String](receiver, BasicExceptions)
+    val chain = new Service[String, String] {
```
mosesn: instead of chain, simply: `statsFilter.andThen(service)`

finagle-core/src/test/scala/com/twitter/finagle/service/StatsFilterTest.scala
```diff
@@ -5,10 +5,13 @@ import com.twitter.finagle.stats.{CategorizingExceptionStatsHandler, ExceptionSt
 import com.twitter.finagle._
 import com.twitter.util._
 import java.util.concurrent.TimeUnit
+
 import org.junit.runner.RunWith
 import org.scalatest.FunSuite
 import org.scalatest.junit.JUnitRunner
 
+import scala.util.control.ControlThrowable
```
mosesn: this is unused

finagle-core/src/test/scala/com/twitter/finagle/service/StatsFilterTest.scala
```diff
@@ -5,10 +5,13 @@ import com.twitter.finagle.stats.{CategorizingExceptionStatsHandler, ExceptionSt
 import com.twitter.finagle._
 import com.twitter.util._
 import java.util.concurrent.TimeUnit
+
```
mosesn: rm newline please

finagle-core/src/test/scala/com/twitter/finagle/service/StatsFilterTest.scala
```diff
@@ -141,6 +144,21 @@ class StatsFilterTest extends FunSuite {
...
+    val service = new Service[String, String] {
+      def apply(request: String): Future[String] = throw new Exception("broken")
+    }
+    val statsFilter = new StatsFilter[String, String](receiver, BasicExceptions)
+    val chain = new Service[String, String] {
+      def apply(request: String): Future[String] = statsFilter.apply(request, service)
+    }
+
+    assert(receiver.gauges(Seq("pending"))() == 0.0)
+    chain("foo")
```
mosesn: can we also assert that this is failed?

finagle-core/src/test/scala/com/twitter/finagle/service/StatsFilterTest.scala
reikje: yes sounds good. gonna add this tomorrow.

finagle-core/src/main/scala/com/twitter/finagle/service/StatsFilter.scala
reikje: no problem, will fix

finagle-core/src/test/scala/com/twitter/finagle/service/StatsFilterTest.scala
reikje: will do

finagle-core/src/test/scala/com/twitter/finagle/service/StatsFilterTest.scala
reikje: This is on purpose. If I use `andThen` then this is already [wrapping](https://github.com/twitter/finagle/blob/develop/finagle-core/src/main/scala/com/twitter/finagle/Filter.scala#L77) `NonFatal`'s. Which means that the test wouldn't trigger the problem we are trying to fix.

finagle-core/src/test/scala/com/twitter/finagle/service/StatsFilterTest.scala
reikje: yes, will add it

reikje:
Changes are made @mosesn

vkostyukov:
Thanks @reikje! This has been merged in (6e3306d9c2d2e0649b1c8a6572775d689f41e488).

PR #596, opened 2017-03-24
finagle-toggle: Remove unused params

zaneli:
Problem

`loadJsonConfig` method receives `libraryName` and `configName` params, but finally the latter only seems to be used.

Solution

Remove `libraryName` from `loadJsonConfig` and `loadJsonConfigWithEnv` method.


mosesn:
LGTM, do you mind editing your commit message to include the Problem / Solution you have in your description?

zaneli:
Sure.
I did it and force push.
Would this be OK?

vkostyukov:
This was merged in (cf4c0f5a42f32da057b04211895f519ffc2f3abc). Thanks @zaneli!

PR #602, opened 2017-04-05
finagle-core: Fix BufChannelBuffer.getLong()

zaneli:
Problem

In the case of specifying LITTLE_ENDIAN, due to lack of `|` operator `BufChannelBuffer.getLong()` returns incorrect value.

Solution

Fix `BufChannelBuffer.getLong()` method by adding `|` operator.

Result

Even in the case of specifying LITTLE_ENDIAN, `BufChannelBuffer.getLong()` returns correct value.


vkostyukov:
Omg. Pulling this internally.

vkostyukov:
This has been merged (1e0b86593bd6b66ee5a92155e22759ebd9fd63be). Thanks @zaneli!

Issue #600, opened 2017-04-03
No peer certificate in context when using netty4.


koshelev:
When using `finagle-netty4` from the `develop` branch, peer certificate is not available in the context after successful TLS handshake.

### Expected behavior

If server is configured to use mutual TLS authentication and TLS handshake was successful, than service can query client's certificate using `com.twitter.finagle.transport.Transport.peerCertificate`.

### Actual behavior

`Transport.peerCertificate` returns `None`. 
[ChannelTransport.peerCertificate](https://github.com/twitter/finagle/blob/173a090a26745afaddbb686f0f04553b3cd1ebb5/finagle-netty4/src/main/scala/com/twitter/finagle/netty4/transport/ChannelTransport.scala#L193) is initialized before handshake is finished.

### Steps to reproduce the behavior

```scala
package com.twitter.finagle.http

import java.net.InetSocketAddress

import com.twitter.finagle.{Address, Service, Http => FinagleHttp}
import com.twitter.finagle.ssl.client.SslClientConfiguration
import com.twitter.finagle.ssl.{ClientAuth, Engine}
import com.twitter.finagle.ssl.server.SslServerConfiguration
import com.twitter.finagle.stats.NullStatsReceiver
import com.twitter.finagle.transport.Transport
import com.twitter.util.{Await, Future}
import org.scalatest.FunSuite
import io.netty.handler.ssl.SslContextBuilder
import io.netty.handler.ssl.util.SelfSignedCertificate

class SslHandshakeEndToEndTest extends FunSuite {

  test("Peer certificate is available") {

    val service = new Service[Request, Response] {
      def apply(request: Request): Future[Response] = {
        val response = Response( Transport.peerCertificate.fold(Status.NotFound)(_ => Status.Ok) )
        Future.value(response)
      }
    }

    val serverCert = new SelfSignedCertificate()
    val clientCert = new SelfSignedCertificate()
    val allocator = io.netty.buffer.UnpooledByteBufAllocator.DEFAULT
    val server = FinagleHttp.server
      .configured(FinagleHttp.Netty4Impl)
      .withTransport.tls(SslServerConfiguration(clientAuth = ClientAuth.Needed), (config: SslServerConfiguration) => {
      val context = SslContextBuilder.forServer(serverCert.key(), serverCert.cert()).trustManager(clientCert.cert()).build()
      val engine = context.newEngine(allocator)
      engine.setNeedClientAuth(true)
      new Engine(engine)
[...]

PR #603, opened 2017-04-05
Delay service instantiation till TLS handshake finished

koshelev:
Problem

https://github.com/twitter/finagle/issues/600

Solution

- changing `ChannelTransport.peerCertificate` from `val` to `def`
- call `serveTransport` either after successful handshake, if a `SslHandler` is installed into pipeline or immediately, if none.

### DIFF
--- a/finagle-netty4/src/main/scala/com/twitter/finagle/netty4/channel/ServerBridge.scala
+++ b/finagle-netty4/src/main/scala/com/twitter/finagle/netty4/channel/ServerBridge.scala
@@ -1,7 +1,7 @@
 package com.twitter.finagle.netty4.channel
 
 import com.twitter.finagle.transport.Transport
-import io.netty.channel.{ChannelInitializer, Channel}
+import io.netty.channel.{Channel, ChannelHandlerContext, ChannelInboundHandlerAdapter, ChannelInitializer}
 import io.netty.channel.ChannelHandler.Sharable
 
 /**
@@ -11,10 +11,10 @@ import io.netty.channel.ChannelHandler.Sharable
 private[netty4] class ServerBridge[In, Out](
     transportFac: Channel => Transport[In, Out],
     serveTransport: Transport[In, Out] => Unit)
-  extends ChannelInitializer[Channel] {
+  extends ChannelInboundHandlerAdapter {
 
-  def initChannel(ch: Channel): Unit = {
-    val transport: Transport[In, Out] = transportFac(ch)
+  override def channelActive(ctx: ChannelHandlerContext): Unit = {
+    val transport: Transport[In, Out] = transportFac(ctx.channel())
     serveTransport(transport)
   }
 }
--- a/finagle-netty4/src/main/scala/com/twitter/finagle/netty4/ssl/server/Netty4ServerSslHandler.scala
+++ b/finagle-netty4/src/main/scala/com/twitter/finagle/netty4/ssl/server/Netty4ServerSslHandler.scala
@@ -53,8 +53,13 @@ private[netty4] class Netty4ServerSslHandler(
     // create an `SslHandler`.
     new SslHandler(engine.self)
 
-  private[this] def addHandlerToPipeline(pipeline: ChannelPipeline, sslHandler: SslHandler): Unit =
+  private[this] def addHandlersToPipeline(pipeline: ChannelPipeline,
+    sslHandler: SslHandler,
+    sslConnectHandler: SslServerConnectHandler
+  ): Unit = {
+    pipeline.addFirst("sslConnect", sslConnectHandler)
     pipeline.addFirst("ssl", sslHandler)
+  }
 
   /**
    * In this method, an `Engine` is created by an `SslServerEngineFactory` via
@@ -69,7 +74,7 @@ private[netty4] class Netty4ServerSslHandler(
       val combined: SslServerConfiguration = combineApplicationProtocols(config)
       val engine: Engine = factory(combined)
       val sslHandler: SslHandler = createSslHandler(engine)
-      addHandlerToPipeline(ch.pipeline(), sslHandler)
+      addHandlersToPipeline(ch.pipeline(), sslHandler, new SslServerConnectHandler(sslHandler))
     }
   }
 }
--- a/finagle-netty4/src/main/scala/com/twitter/finagle/netty4/transport/ChannelTransport.scala
+++ b/finagle-netty4/src/main/scala/com/twitter/finagle/netty4/transport/ChannelTransport.scala
@@ -139,7 +139,7 @@ private[finagle] class ChannelTransport(ch: Channel) extends Transport[Any, Any]
     closed.unit
   }
 
-  val peerCertificate: Option[Certificate] = ch.pipeline.get(classOf[SslHandler]) match {
+  def peerCertificate: Option[Certificate] = ch.pipeline.get(classOf[SslHandler]) match {
     case null => None
     case handler =>
       try {
--- a/finagle-netty4/src/test/scala/com/twitter/finagle/netty4/Netty4SslTest.scala
+++ b/finagle-netty4/src/test/scala/com/twitter/finagle/netty4/Netty4SslTest.scala
@@ -1,0 +1,124 @@
+package com.twitter.finagle.netty4
+
+import com.twitter.conversions.time._
+import com.twitter.finagle
+import com.twitter.finagle.Stack.Params
+import com.twitter.finagle._
+import com.twitter.finagle.client.{StackClient, StdStackClient, Transporter}
+import com.twitter.finagle.dispatch.{SerialClientDispatcher, SerialServerDispatcher}
+import com.twitter.finagle.param.Label
+import com.twitter.finagle.ssl.client.{SslClientConfiguration, SslClientEngineFactory}
+import com.twitter.finagle.ssl.server.{SslServerConfiguration, SslServerEngineFactory}
+import com.twitter.finagle.ssl.{ClientAuth, Engine, KeyCredentials, TrustCredentials}
+import com.twitter.finagle.transport.Transport
+import com.twitter.finagle.transport.Transport.ServerSsl
+import com.twitter.util.{Await, Future}
+import io.netty.channel.ChannelPipeline
+import io.netty.handler.codec.string.{StringDecoder, StringEncoder}
+import io.netty.handler.codec.{DelimiterBasedFrameDecoder, Delimiters}
+import io.netty.handler.ssl.SslContextBuilder
+import io.netty.handler.ssl.util.SelfSignedCertificate
+import java.net.{InetAddress, InetSocketAddress, SocketAddress}
+import java.nio.charset.StandardCharsets.UTF_8
+
+import org.scalatest.Outcome
+import org.scalatest.fixture.FunSuite
+import org.scalatest.concurrent.{Eventually, IntegrationPatience}
+
+object Netty4SslTest {
+  case class Client(clientCert: SelfSignedCertificate, serverCert: SelfSignedCertificate,
+      params: Params = Params.empty,
+      stack: Stack[ServiceFactory[String, String]] = StackClient.newStack)
+    extends StdStackClient[String, String, Client] {
+
+    override protected type In = String
+    override protected type Out = String
+
+    override protected def newTransporter(addr: SocketAddress): Transporter[String, String] =
+      Netty4Transporter.raw[String, String](pipeline => {
+        pipeline.addLast(new DelimiterBasedFrameDecoder(8192, Delimiters.lineDelimiter(): _*))
+        pipeline.addLast(new StringDecoder())
+        pipeline.addLast(new StringEncoder())
+      }, addr, params)
+
+    override protected def newDispatcher(transport: Transport[String, String]): Service[String, String] =
+      new SerialClientDispatcher(transport)
+
+    override protected def copy1(stack: Stack[ServiceFactory[String, String]], params: Params) =
+      copy(clientCert, serverCert, params, stack)
+  }
+}
+
+class Netty4SslTest extends FunSuite with Eventually with IntegrationPatience {
+
+  class Ctx {
+    val serverCert = new SelfSignedCertificate("example.server.com")
+    val clientCert = new SelfSignedCertificate("example.client.com")
+    val allocator = io.netty.buffer.UnpooledByteBufAllocator.DEFAULT
+
+    private object StringServerInit extends (ChannelPipeline => Unit) {
+      def apply(pipeline: ChannelPipeline): Unit = {
+        pipeline.addLast("line", new DelimiterBasedFrameDecoder(100, Delimiters.lineDelimiter(): _*))
+        pipeline.addLast("stringDecoder", new StringDecoder(UTF_8))
+        pipeline.addLast("stringEncoder", new StringEncoder(UTF_8))
+      }
+    }
+
+    val server = {
+      val service =
+        new Service[String, String] {
+          override def apply(request: String): Future[String] = {
+            Future.value(
+              Transport.peerCertificate match {
+                case Some(_) => "OK\n"
+                case None => "ERROR\n"
+              }
+            )
+          }
+        }
+
+      val serverConfig = SslServerConfiguration(
+        keyCredentials = KeyCredentials.CertAndKey(serverCert.certificate(), serverCert.privateKey()),
+        trustCredentials = TrustCredentials.CertCollection(clientCert.certificate()),
+        clientAuth = ClientAuth.Needed)
+
+      val p = Params.empty +
+        ServerSsl(Some(serverConfig)) +
+        Label("test")
+      val listener = Netty4Listener[String, String](StringServerInit, p)
+      val serveTransport = (t: Transport[String, String]) => {
+        if (t.peerCertificate.isEmpty) throw new IllegalStateException("No peer certificate in transport")
+        new SerialServerDispatcher(t, service)
+      }
+      listener.listen(new InetSocketAddress(InetAddress.getLoopbackAddress, 0))(serveTransport(_))
+    }
+
+    val clientConfig = SslClientConfiguration(
+      keyCredentials = KeyCredentials.CertAndKey(clientCert.certificate(), clientCert.privateKey()),
+      trustCredentials = TrustCredentials.CertCollection(serverCert.certificate()))
+    val client = {
+      val addr = server.boundAddress.asInstanceOf[InetSocketAddress]
+      new finagle.netty4.Netty4SslTest.Client(clientCert, serverCert)
+        .withTransport.tls(clientConfig)
+        .newService(s"${addr.getHostName}:${addr.getPort}", "client")
+    }
+
+    def close() = Future.collect(Seq(client.close(), server.close()))
+  }
+
+  override type FixtureParam = Ctx
+
+  override def withFixture(test: OneArgTest): Outcome = {
+    val ctx = new Ctx
+    try {
+      withFixture(test.toNoArgTest(ctx))
+    } finally {
+      Await.ready(ctx.close(), 3.seconds)
+    }
+  }
+
+  test("Peer certificate is available to service") { ctx =>
+    val response = Await.result(ctx.client("security is overrated!\n"), 3.seconds)
+    assert(response == "OK")
+  }
+}
--- a/finagle-netty4/src/main/scala/com/twitter/finagle/netty4/channel/ConnectPromiseDelayListeners.scala
+++ b/finagle-netty4/src/main/scala/com/twitter/finagle/netty4/channel/ConnectPromiseDelayListeners.scala
@@ -1,7 +1,7 @@
 package com.twitter.finagle.netty4.channel
 
 import io.netty.channel._
-import io.netty.util.concurrent.{Future => NettyFuture, GenericFutureListener}
+import io.netty.util.concurrent.{GenericFutureListener, Future => NettyFuture}
 
 /**
  * This trait provides [[GenericFutureListener]]s that are useful for implementing handlers
@@ -10,7 +10,7 @@ import io.netty.util.concurrent.{Future => NettyFuture, GenericFutureListener}
  *
  * For example, this is used by:
  *
- * - [[com.twitter.finagle.netty4.ssl.SslConnectHandler]]
+ * - [[com.twitter.finagle.netty4.ssl.client.SslClientConnectHandler]
  * - [[com.twitter.finagle.netty4.proxy.HttpProxyConnectHandler]]
  * - [[com.twitter.finagle.netty4.proxy.Netty4ProxyConnectHandler]]
  */
--- a/finagle-netty4/src/main/scala/com/twitter/finagle/netty4/ssl/client/Netty4ClientSslHandler.scala
+++ b/finagle-netty4/src/main/scala/com/twitter/finagle/netty4/ssl/client/Netty4ClientSslHandler.scala
@@ -1,7 +1,7 @@
 package com.twitter.finagle.netty4.ssl.client
 
 import com.twitter.finagle.client.Transporter
-import com.twitter.finagle.netty4.ssl.{Alpn, SslConnectHandler}
+import com.twitter.finagle.netty4.ssl.Alpn
 import com.twitter.finagle.ssl.{ApplicationProtocols, Engine, SessionVerifier}
 import com.twitter.finagle.ssl.client.{SslClientConfiguration, SslClientEngineFactory}
 import com.twitter.finagle.transport.Transport
@@ -81,18 +81,18 @@ private[netty4] class Netty4ClientSslHandler(
   private[this] def createSslConnectHandler(
     sslHandler: SslHandler,
     config: SslClientConfiguration
-  ): SslConnectHandler = {
+  ): SslClientConnectHandler = {
     val sessionValidation = config.hostname
       .map(SessionVerifier.hostname)
       .getOrElse(SessionVerifier.AlwaysValid)
 
-    new SslConnectHandler(sslHandler, sessionValidation)
+    new SslClientConnectHandler(sslHandler, sessionValidation)
   }
 
   private[this] def addHandlersToPipeline(
     pipeline: ChannelPipeline,
     sslHandler: SslHandler,
-    sslConnectHandler: SslConnectHandler
+    sslConnectHandler: SslClientConnectHandler
   ): Unit = {
     pipeline.addFirst("sslConnect", sslConnectHandler)
     pipeline.addFirst("ssl", sslHandler)
@@ -113,7 +113,7 @@ private[netty4] class Netty4ClientSslHandler(
       val combined: SslClientConfiguration = combineApplicationProtocols(config)
       val engine: Engine = factory(address, combined)
       val sslHandler: SslHandler = createSslHandler(engine)
-      val sslConnectHandler: SslConnectHandler = createSslConnectHandler(sslHandler, combined)
+      val sslConnectHandler: SslClientConnectHandler = createSslConnectHandler(sslHandler, combined)
       addHandlersToPipeline(ch.pipeline(), sslHandler, sslConnectHandler)
     }
   }
--- a/finagle-netty4/src/main/scala/com/twitter/finagle/netty4/ssl/client/SslClientConnectHandler.scala
+++ b/finagle-netty4/src/main/scala/com/twitter/finagle/netty4/ssl/client/SslClientConnectHandler.scala
@@ -1,9 +1,9 @@
-package com.twitter.finagle.netty4.ssl
+package com.twitter.finagle.netty4.ssl.client
 
-import com.twitter.finagle.netty4.channel.{ConnectPromiseDelayListeners, BufferingChannelOutboundHandler}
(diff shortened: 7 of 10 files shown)


mosesn:
@koshelev do you mind changing the commit message to the

```
finagle-core: Delay service instantiation till TLS handshake finished

Problem

....

Solution

....

Result

....
```

style?

koshelev:
@mosesn Done

finagle-netty4/src/main/scala/com/twitter/finagle/netty4/channel/ServerBridge.scala
```diff
@@ -15,6 +17,17 @@ private[netty4] class ServerBridge[In, Out](
 
   def initChannel(ch: Channel): Unit = {
     val transport: Transport[In, Out] = transportFac(ch)
-    serveTransport(transport)
+    val sslHandler = ch.pipeline().get(classOf[SslHandler])
+    if(sslHandler != null ) {
```
mosesn: spacing here should be:

`if (sslHandler != null) {`

finagle-netty4/src/main/scala/com/twitter/finagle/netty4/channel/ServerBridge.scala
```diff
@@ -15,6 +17,17 @@ private[netty4] class ServerBridge[In, Out](
...
+      sslHandler.handshakeFuture().addListener( new GenericFutureListener[NettyFuture[Channel]] {
+        override def operationComplete(future: NettyFuture[Channel]): Unit = {
+          if(future.isSuccess) {
```
mosesn: should be:

`if (future.isSuccess) {`

mosesn: we should make sure we handle non-success cases (interrupt, failure) gracefully here.

also, is it possible this will hang?  if I remember correctly, the sslhandler has a default timeout of 10 seconds?  is that right?

finagle-netty4/src/main/scala/com/twitter/finagle/netty4/channel/ServerBridge.scala
```diff
@@ -15,6 +17,17 @@ private[netty4] class ServerBridge[In, Out](
...
+            serveTransport(transport)
+          }
+        }
+      } )
```
mosesn: no space between `})`.

finagle-netty4/src/main/scala/com/twitter/finagle/netty4/channel/ServerBridge.scala
```diff
@@ -15,6 +17,17 @@ private[netty4] class ServerBridge[In, Out](
 
   def initChannel(ch: Channel): Unit = {
     val transport: Transport[In, Out] = transportFac(ch)
-    serveTransport(transport)
+    val sslHandler = ch.pipeline().get(classOf[SslHandler])
+    if(sslHandler != null ) {
+      sslHandler.handshakeFuture().addListener( new GenericFutureListener[NettyFuture[Channel]] {
```
mosesn: no space between `(new GenericFutureListener[NettyFuture[Channel]] {`

finagle-netty4/src/test/scala/com/twitter/finagle/netty4/Netty4SslTest.scala
```diff
@@ -0,0 +1,111 @@
+package com.twitter.finagle.netty4
+
+import java.net.{InetAddress, InetSocketAddress, SocketAddress}
```
mosesn: please sort the java.* with the rest of the imports.  we don't treat any imports specially.

finagle-netty4/src/test/scala/com/twitter/finagle/netty4/Netty4SslTest.scala
```diff
@@ -0,0 +1,111 @@
...
+import io.netty.handler.ssl.SslContextBuilder
+import io.netty.handler.ssl.util.SelfSignedCertificate
+import org.junit.runner.RunWith
+import org.scalatest.FunSuite
+import org.scalatest.concurrent.{Eventually, IntegrationPatience}
+import org.scalatest.junit.JUnitRunner
+
+object Netty4SslTest {
+  case class Client(clientCert: SelfSignedCertificate, serverCert: SelfSignedCertificate,
+                    params: Params = Params.empty,
```
mosesn: indentation for classes should be like

```
case class ClassName(
    arg1: Arg1,
    arg2: Arg2)
```

so that each argument is indented by four spaces.

finagle-netty4/src/test/scala/com/twitter/finagle/netty4/Netty4SslTest.scala
```diff
@@ -0,0 +1,111 @@
...
+  val serverCert = new SelfSignedCertificate()
+  val clientCert = new SelfSignedCertificate()
+  val allocator = io.netty.buffer.UnpooledByteBufAllocator.DEFAULT
+
+  val server = {
+    val service =
+      new Service[String, String] {
+        override def apply(request: String): Future[String] = {
+          Future.value(
+            Transport.peerCertificate.fold("ERROR")(_ => "OK\n")
```
mosesn: do you mind using a match instead of a fold here?  I think it's a bit more legible.

finagle-netty4/src/test/scala/com/twitter/finagle/netty4/Netty4SslTest.scala
```diff
@@ -0,0 +1,111 @@
...
+      ServerSsl(Some(SslServerConfiguration(clientAuth = ClientAuth.Needed))) +
+      SslServerEngineFactory.Param( new SslServerEngineFactory {
+        override def apply(config: SslServerConfiguration): Engine = {
+          val context = SslContextBuilder.forServer(serverCert.key(), serverCert.cert()).trustManager(clientCert.cert()).build()
+          val engine = context.newEngine(allocator)
+          engine.setNeedClientAuth(true)
+          new Engine(engine)
+        }
+      })
+    Label("test")
```
mosesn: I think we're missing a `+` here, so I don't think the label will be applied.

finagle-netty4/src/test/scala/com/twitter/finagle/netty4/Netty4SslTest.scala
```diff
@@ -0,0 +1,111 @@
...
+    val client = new finagle.netty4.Netty4SslTest.Client(clientCert, serverCert)
+      .withTransport.tls(SslClientConfiguration(), new SslClientEngineFactory {
+        override def apply(address: Address, config: SslClientConfiguration): Engine = {
+          val ctx = SslContextBuilder.forClient.keyManager(clientCert.key(), clientCert.cert()).trustManager(serverCert.cert()).build()
+          new Engine(ctx.newEngine(allocator))
+        }
+      })
+      .newService(s"${addr.getHostName}:${addr.getPort}", "client")
+
+    val response = Await.result(client("security is overrated!\n"))
```
mosesn: please add a timeout to the Await, so that if we introduce a regression the test doesn't hang.

finagle-netty4/src/test/scala/com/twitter/finagle/netty4/Netty4SslTest.scala
```diff
@@ -0,0 +1,111 @@
...
+    val listener = Netty4Listener[String, String](StringServerInit, p)
+    val serveTransport = (t: Transport[String, String]) => {
+      if( t.peerCertificate.isEmpty ) throw new IllegalStateException("No peer certificate in transport")
+      new SerialServerDispatcher(t, service)
+    }
+    listener.listen(new InetSocketAddress(InetAddress.getLoopbackAddress, 0))(serveTransport(_))
+  }
+
+
+  test("Peer certificate is available to service") {
```
mosesn: can we test what happens when the ssl handshake doesn't finish successfully?

finagle-netty4/src/test/scala/com/twitter/finagle/netty4/Netty4SslTest.scala
```diff
@@ -0,0 +1,111 @@
...
+      pipeline.addLast("stringDecoder", new StringDecoder(UTF_8))
+      pipeline.addLast("stringEncoder", new StringEncoder(UTF_8))
+    }
+  }
+
+  val serverCert = new SelfSignedCertificate()
+  val clientCert = new SelfSignedCertificate()
+  val allocator = io.netty.buffer.UnpooledByteBufAllocator.DEFAULT
+
+  val server = {
```
mosesn: how about making a `class Ctx` and shoving this server in there?  that way we can write our tests like:

```
test("some test stuff") {
  val ctx = new Ctx()
  import ctx._
  server.blahblahblah()
  ....
}
```

and each test will have its own server, but we can write it as if the server was some shared resource.

finagle-netty4/src/main/scala/com/twitter/finagle/netty4/channel/ServerBridge.scala
```diff
@@ -15,6 +17,17 @@ private[netty4] class ServerBridge[In, Out](
 
   def initChannel(ch: Channel): Unit = {
     val transport: Transport[In, Out] = transportFac(ch)
-    serveTransport(transport)
```
vkostyukov: Is there a way to make it a separate channel handler? I guess my question is can we "delay" channel init?

koshelev: Not really, if I understand your question correctly. `initChannel` will be called if `channelRegistered` event is triggered or a `ChannelInitializer` is added to a pipeline of a registered channel.

vkostyukov: Oh I think we can delay `channelRegistered` then given that `channelInit` is [just a fancy version of it](https://github.com/netty/netty/blob/4.1/transport/src/main/java/io/netty/channel/ChannelInitializer.java#L73-L84). What do you think?

finagle-netty4/src/main/scala/com/twitter/finagle/netty4/channel/ServerBridge.scala
koshelev: If I understand [netty code](https://github.com/netty/netty/blob/1419f5b601ec7f50938a5e745ec2312f881a580f/handler/src/main/java/io/netty/handler/ssl/SslHandler.java#L1397) correctly, it is not necessary for netty4. `SslHandler` will shutdown `SSLEngine` and context (and it will close the underlying channel)

koshelev: Yes, default timeout is 10 seconds.

finagle-netty4/src/main/scala/com/twitter/finagle/netty4/transport/ChannelTransport.scala
```diff
@@ -139,7 +139,7 @@ private[finagle] class ChannelTransport(ch: Channel) extends Transport[Any, Any]
     closed.unit
   }
 
-  val peerCertificate: Option[Certificate] = ch.pipeline.get(classOf[SslHandler]) match {
+  def peerCertificate: Option[Certificate] = ch.pipeline.get(classOf[SslHandler]) match {
```
vkostyukov: I guess when we'll actually delay the `ChannelTransport` creation, the peer certificate will be available at the moment this constructor fires and yet I still think it should be `def`, not `val`.  I like this more by two reasons.

1. I'd imagine that sessions could be renegitoated so having `def` means we could get the latest peer certificate if that happens.
2. This would reduce the memory footprint for each channel-transport (each connection).

finagle-netty4/src/main/scala/com/twitter/finagle/netty4/channel/ServerBridge.scala
koshelev: Ok, I believe now I finally understand what do you mean :). So you are suggesting to implement a server `SslConnectHandler` and delay `channelRegistered` in it? It should be possible, but we also will have to change `ServerBridge` to be a handler.

vkostyukov: Exactly! Although, I don't understand why we'd need to touch ``ServerBridge`. It's a ChannelInitialize right now that [essentially an InboundHanlder](https://github.com/netty/netty/blob/4.1/transport/src/main/java/io/netty/channel/ChannelInitializer.java#L53) that will receive `channelRegistered` (or `channelInit` in the `ChannelInitializer` case) as any other handler.

finagle-netty4/src/test/scala/com/twitter/finagle/netty4/Netty4SslTest.scala
koshelev: There is not much we can test in case of failed handshake - netty handles everything in case of failure.

finagle-netty4/src/main/scala/com/twitter/finagle/netty4/channel/ServerBridge.scala
koshelev: Yes, but `channelInit` also will be called when we add `ServerBridge` to the pipeline in our case: https://github.com/netty/netty/blob/4.1/transport/src/main/java/io/netty/channel/ChannelInitializer.java#L99

vkostyukov: Oh interesting. It looks like [we change the `isRegistered` flag on a channel before firing the `channelRegistered` event](https://github.com/netty/netty/blob/4.1/transport/src/main/java/io/netty/channel/AbstractChannel.java#L502-L509). You're totally right. We should probably change server bridge to be just `ChannelInboundHandler`. Is that what you had in mind?

koshelev: Yes, making server bridge `ChannelInboundHandler` and moving code from `channelInit` to `channelRegistered`. 
Could you help me understand why if I change `ServerBridge` like this:
```scala
  def initChannel(ch: Channel): Unit = {
    val sslHandler = ch.pipeline().get(classOf[SslHandler])
    if (sslHandler != null) {
      sslHandler.handshakeFuture().addListener(new GenericFutureListener[NettyFuture[Channel]] {
        override def operationComplete(future: NettyFuture[Channel]): Unit = {
          if (future.isSuccess) {
            val transport: Transport[In, Out] = transportFac(ch)
            serveTransport(transport)
          }
        }
      })
    } else {
      val transport: Transport[In, Out] = transportFac(ch)
      serveTransport(transport)
    }
  }
```
handshake does not happen?

vkostyukov: How do you know it never happens? This looks pretty reasonable to me. As far as I remember, the handshake is initiated on either `channelActive` or `handlerAdded`. Here you're attaching the handshake listener on `channelRegistered`, but it shouldn't matter and the listener should be fired when the handshake is done anyway.

Maybe just make it a pure ChannelHandler and override (and delay) `channelActive` instead?

koshelev: handshake times out when I'm running the test from this PR.  Just to be explicit - the only change I did is also delaying `transportFac(ch)`.
Yes, client initiates handshake on `channelActive`

koshelev: @vkostyukov I think I figured it out - we are using backpressure, so channel's auto read is set to `false`.  Means `transportFac(ch)` cannot be delayed

koshelev: Here is an [alternate implementation](https://github.com/koshelev/finagle/commit/b86a94c669ef91148283435be00e95e13b48811a), with a handler deferring the `channelRegistered` event. As we still need ChannelTransport, it will be added in `handlerAdded`. As consequence `ServerBridge` is not sharable any more. What do you think?

vkostyukov: Interesting. I'm pretty sure SslHandler should work just fine with auto-read is disabled. Maybe it's b/c we're not issuing an initial `read` given that `ChannelTransport` is not instantiated? If so I think it's a bug in Netty and we should file a ticket about that. As a workaround in your "delay" handler you can issue a `read` manually.

koshelev: Do you mean `SslHandler` should issue an initial read?

vkostyukov: Oh yeah, absolutely. And if it doesn't it's a bug.

koshelev: I'm not so convinced :). Why should it, if we explicitly configured channel do not read any data  unless we explicitly tell it to do so. And `SslHandler` has no idea, if we want to serve the channel or not.
But anyway, the workaround you suggested seems to work: https://github.com/koshelev/finagle/commit/282ca2c5f281fabcdaa22495a07d1cfd850628ad

vkostyukov: Sweet! Just let us know once you update this PR so we can have another look.

koshelev:
PR updated

finagle-netty4/src/main/scala/com/twitter/finagle/netty4/ssl/SslConnectHandler.scala
```diff
@@ -7,6 +7,28 @@ import io.netty.util.concurrent.{Future => NettyFuture, GenericFutureListener}
 import java.net.SocketAddress
 import javax.net.ssl.SSLSession
 
+private[netty4] class SslServerConnectHandler(ssl: SslHandler) extends ChannelInboundHandlerAdapter { self =>
```
vkostyukov: Maybe put it in its own file? Also rename `SslConnectHandler` to `SslClientConectHandler`?

finagle-netty4/src/main/scala/com/twitter/finagle/netty4/ssl/SslConnectHandler.scala
```diff
@@ -7,6 +7,28 @@ import io.netty.util.concurrent.{Future => NettyFuture, GenericFutureListener}
...
+
+  override def channelActive(ctx: ChannelHandlerContext): Unit = {
+    val channel = ctx.channel()
+    if (!channel.config().isAutoRead) {
+      channel.read()
```
vkostyukov: I guess we can't issue this read on `channelRegistred`?

finagle-netty4/src/main/scala/com/twitter/finagle/netty4/ssl/SslConnectHandler.scala
```diff
@@ -7,6 +7,28 @@ import io.netty.util.concurrent.{Future => NettyFuture, GenericFutureListener}
...
+    }
+    super.channelActive(ctx)
+  }
+
+  override def channelRegistered(ctx: ChannelHandlerContext): Unit = {
+    ssl.handshakeFuture().addListener(new GenericFutureListener[NettyFuture[Channel]] {
+        override def operationComplete(future: NettyFuture[Channel]): Unit = {
+          if (future.isSuccess) {
+            ctx.pipeline().remove(self)
+            SslServerConnectHandler.super.channelRegistered(ctx)
```
vkostyukov: Let's just do `ctx.channelRegistered`.

finagle-netty4/src/main/scala/com/twitter/finagle/netty4/channel/ServerBridge.scala
```diff
@@ -13,6 +13,8 @@ private[netty4] class ServerBridge[In, Out](
     serveTransport: Transport[In, Out] => Unit)
   extends ChannelInitializer[Channel] {
 
+  override def handlerAdded(ctx: ChannelHandlerContext): Unit = {}
```
vkostyukov: Why we need this? If that's to prevent init to be called maybe we should just rebrand it as a plain `InboundChannelHandler` to be explicit about what we want?

finagle-netty4/src/main/scala/com/twitter/finagle/netty4/ssl/SslConnectHandler.scala
vkostyukov: Also, please do `ctx.read()` instead.

finagle-netty4/src/main/scala/com/twitter/finagle/netty4/channel/ServerBridge.scala
koshelev: Yes, it is to prevent init to be called.
I also was considering to rebrand it, but after closer look at the sources of `ChannelInitializer` it seems like we will have to duplicate a lot of its code. So I decided the overriding `handlerAdded` is lesser of two evils.

finagle-netty4/src/main/scala/com/twitter/finagle/netty4/channel/ServerBridge.scala
```diff
@@ -11,7 +11,9 @@ import io.netty.channel.ChannelHandler.Sharable
 private[netty4] class ServerBridge[In, Out](
     transportFac: Channel => Transport[In, Out],
     serveTransport: Transport[In, Out] => Unit)
-  extends ChannelInitializer[Channel] {
+  extends ChannelInboundHandlerAdapter {
+
+  override def handlerAdded(ctx: ChannelHandlerContext): Unit = {}
```
vkostyukov: I'm still not sure why we need this. Also, how does `channelInit` work given that it's now `ChannelInboundHandlerAdapter` and not `ChannelInitializer`?

How do you feel about creating a channel transport on `channelActive` (in this server bridge) and delaying the `channelActive` event instead, in your new `SslServerConnectHandler`?

koshelev: OMG, have no idea how it is landed into PR.

I think it is a good idea! Updated PR

vkostyukov:
Thanks a lot @koshelev! I appreciate your patience going through the review and I really like your final solution. Really nice work!

Let me get somebody else from the team to take a look.

koshelev:
@vkostyukov thanks a lot for taking time to review the PR and your feedback! Hope to submit next PR next week (in case you are interested in server side SNI support :) )

finagle-netty4/src/test/scala/com/twitter/finagle/netty4/Netty4SslTest.scala
```diff
@@ -0,0 +1,131 @@
...
+    override protected def newDispatcher(transport: Transport[String, String]): Service[String, String] =
+      new SerialClientDispatcher(transport)
+
+    override protected def copy1(stack: Stack[ServiceFactory[String, String]], params: Params) =
+      copy(clientCert, serverCert, params, stack)
+  }
+}
+
+@RunWith(classOf[JUnitRunner])
+class Netty4SslTest extends fixture.FunSuite with Eventually with IntegrationPatience {
```
nepthar: nit: Import & use FunSuite directly

finagle-netty4/src/test/scala/com/twitter/finagle/netty4/Netty4SslTest.scala
```diff
@@ -0,0 +1,131 @@
...
+
+    override protected def newDispatcher(transport: Transport[String, String]): Service[String, String] =
+      new SerialClientDispatcher(transport)
+
+    override protected def copy1(stack: Stack[ServiceFactory[String, String]], params: Params) =
+      copy(clientCert, serverCert, params, stack)
+  }
+}
+
+@RunWith(classOf[JUnitRunner])
```
nepthar: This annotation isn't necessary anymore and can be removed.

finagle-netty4/src/test/scala/com/twitter/finagle/netty4/Netty4SslTest.scala
```diff
@@ -0,0 +1,131 @@
...
+    try {
+      withFixture(test.toNoArgTest(ctx))
+    } finally {
+      import com.twitter.conversions.time._
+      Await.ready(ctx.close(), 3.seconds)
+    }
+  }
+
+  test("Peer certificate is available to service") { ctx =>
+    val response = Await.result(ctx.client("security is overrated!\n"))
```
nepthar: Can you also add a timeout here?

finagle-netty4/src/test/scala/com/twitter/finagle/netty4/Netty4SslTest.scala
```diff
@@ -0,0 +1,131 @@
...
+import com.twitter.finagle.transport.Transport.ServerSsl
+import com.twitter.util.{Await, Future}
+import io.netty.channel.ChannelPipeline
+import io.netty.handler.codec.string.{StringDecoder, StringEncoder}
+import io.netty.handler.codec.{DelimiterBasedFrameDecoder, Delimiters}
+import io.netty.handler.ssl.SslContextBuilder
+import io.netty.handler.ssl.util.SelfSignedCertificate
+import java.net.{InetAddress, InetSocketAddress, SocketAddress}
+import java.nio.charset.StandardCharsets.UTF_8
+
```
nepthar: clump all of the imports together, they can be friends

finagle-netty4/src/main/scala/com/twitter/finagle/netty4/ssl/server/Netty4ServerSslHandler.scala
```diff
@@ -53,8 +53,11 @@ private[netty4] class Netty4ServerSslHandler(
     // create an `SslHandler`.
     new SslHandler(engine.self)
 
-  private[this] def addHandlerToPipeline(pipeline: ChannelPipeline, sslHandler: SslHandler): Unit =
+  private[this] def addHandlersToPipeline(pipeline: ChannelPipeline, sslHandler: SslHandler,
```
nepthar: we use a slightly different style here:

    private[this] def addHandlersToPipeline(
      pipeline: ChannelPipeline,
      sslHandler: SslHandler,
      sslConnectHandler: SslServerConnectHandler
    ): Unit = {

koshelev:
@nepthar Thanks for the feedback! Updated PR

finagle-netty4/src/test/scala/com/twitter/finagle/netty4/Netty4SslTest.scala
```diff
@@ -0,0 +1,127 @@
...
+              Transport.peerCertificate match {
+                case Some(_) => "OK\n"
+                case None => "ERROR\n"
+              }
+            )
+          }
+        }
+
+      val p = Params.empty +
+        ServerSsl(Some(SslServerConfiguration(clientAuth = ClientAuth.Needed))) +
```
ryanoneill: Please change to use the Netty4ServerEngineFactory instead. Something along the lines of

val config = SslServerConfiguration(
  keyCredentials = KeyCredentials.CertAndKey(new File(serverCert.cert()), new File(serverCert.key()),
  trustCredentials = TrustCredentials.CertCollection(new File(clientCert.cert()),
  clientAuth = ClientAuth.Needed)

...

finagle-netty4/src/test/scala/com/twitter/finagle/netty4/Netty4SslTest.scala
```diff
@@ -0,0 +1,127 @@
...
+        new SerialServerDispatcher(t, service)
+      }
+      listener.listen(new InetSocketAddress(InetAddress.getLoopbackAddress, 0))(serveTransport(_))
+    }
+
+    val client = {
+      val addr = server.boundAddress.asInstanceOf[InetSocketAddress]
+      new finagle.netty4.Netty4SslTest.Client(clientCert, serverCert)
+        .withTransport.tls(SslClientConfiguration(), new SslClientEngineFactory {
+        override def apply(address: Address, config: SslClientConfiguration): Engine = {
```
ryanoneill: You should be able to just use the Client Configuration here. Something like ...

.withTransport.tls(SslClientConfiguration(
  keyCredentials = KeyCredentials.CertAndKey(new File(clientCert.cert()), new File(clientCert.key())),
  trustCredentials = TrustCredentials.CertCollection(new File(serverCert.cert())))

and you could optionally include the Netty4SslClientEngineFactory as the second parameter.

koshelev:
@ryanoneill Thanks for your feedback! Updated PR

vkostyukov:
@mosesn Do you mind taking a final look? It seems like your approval is the last one needed.

yufangong:
thanks @koshelev , this change has been merged in 23d13d66d2414cc936d71ccd03d2e897f247b91f

finagle-netty4/src/main/scala/com/twitter/finagle/netty4/transport/ChannelTransport.scala
iPlessmann: hello,

was this merge? i don't see it on 18.1.0 nor on develop

PR #604, opened 2017-04-06
finagle-core: Add tailrec annotation

zaneli:
Problem

Some tail recursive methods have no `@tailrec` annotation.

Solution

Added `@tailrec` annotation.

### DIFF
--- a/finagle-core/src/main/scala/com/twitter/finagle/NameTree.scala
+++ b/finagle-core/src/main/scala/com/twitter/finagle/NameTree.scala
@@ -200,6 +200,7 @@ object NameTree {
   /**
    * A string parseable by [[com.twitter.finagle.NameTree.read NameTree.read]].
    */
+  @tailrec
   def show[T: Showable](tree: NameTree[T]): String = tree match {
     case Union(Weighted(_, tree)) => show(tree)
     case Alt(tree) => show(tree)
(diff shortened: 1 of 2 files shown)


finagle-core/src/main/scala/com/twitter/finagle/Dtab.scala
```diff
@@ -261,6 +262,7 @@ object Dentry {
       lazy val show: String = Path.showElem(buf)
     }
 
+    @tailrec
```
mosesn: shouldn't this be on the apply method, not the object?

zaneli: Oh, exactly.
I've tried to fix it, and got compile error.

```
[error] /Users/zaneli/ws/finagle/finagle-core/src/main/scala/com/twitter/finagle/Dtab.scala:267: @tailrec annotated method contains no recursive calls
[error]       def apply(s: String): Label = Label(Buf.Utf8(s))
```

This addition seems to be a my mistake.
I've reverted it.

mosesn:
looks great, thanks for the PR!

vkostyukov:
I'm pulling this internally. Will reply here when it's merged.

vkostyukov:
This has been merged (96f38130d432c683b561e6ddca0347c61c7bcb24). Thanks @zaneli!

PR #608, opened 2017-04-12
fix to force English to HTTP date format

torao:
**Problem**

In `finagle-base-http`, `Message.httpDateFormat()` uses JavaVM-default locale to format the specified date. For this reason, it generate incorrect HTTP date headers in non-English locales and some HTTP clients cannot recognize the value as date.

```
scala> System.getProperty("user.language")
res2: String = ja
scala> com.twitter.finagle.http.Message.httpDateFormat(new Date(0L))
res3: String = 木, 01 1 1970 00:00:00 GMT
```

 In Japanese, the `木` means `Thu` but it violates _wkday_ for [HTTP Full Date Spec](https://www.w3.org/Protocols/rfc2616/rfc2616-sec3.html#sec3.3.1). The `Message.httpDateFormat()` used by `date_=()`, `expires_=()` and `lastModified_=()` in class `Message`.

**Solution**

I added `Locale.ENGLISH` parameter to `FastDateFormat.getInstance()` to force _wkday_ in English. It seems to work very well in Japanese environment.

The other workaround is to use `headerMap.add(String,Date)` instead of these. This method uses another date-format implimentation in `HeaderMap.format(Date)'.

### DIFF
--- a/finagle-base-http/src/main/scala/com/twitter/finagle/http/Message.scala
+++ b/finagle-base-http/src/main/scala/com/twitter/finagle/http/Message.scala
@@ -7,7 +7,7 @@ import com.twitter.util.{Closable, Duration, Future}
 import java.io._
 import java.util.{Iterator => JIterator}
 import java.nio.charset.Charset
-import java.util.{Date, TimeZone}
+import java.util.{Date, Locale, TimeZone}
 import org.apache.commons.lang.StringUtils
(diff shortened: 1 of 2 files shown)


kevinoliver:
merged in 154b3dc830adaf06a4e0e8bee847db5c6ad31769

torao:
You are welcome. Thank you for developing this awesome OSS.

PR #609, opened 2017-04-19
finagle-core: Fix Scaladoc param name

zaneli:
Problem

There is some difference between Scaladoc and real param name.

Solution

Fix Scaladoc param name.


kevinoliver:
this got merged in 477476dcc5dd36a5eebe51a307b5b7f9e67be8c5

thanks again!

PR #610, opened 2017-04-23
finagle-{core, memcached, ostrich4}: Use collection.breakOut

zaneli:
Problem

use `map(...)(breakOut)` is slightly more efficient than `map(...).toSeq` and so on.

Solution

replace `map(...).toSeq`, `toList`, `toMap` with `map(...)(breakOut)`.


yufangong:
Thanks @zaneli, your change has been merged here: 8fbabf3761c71e0f055315f6ed4fcbdc67bbff6e

Issue #617, opened 2017-05-11
Tolerate X-B3-Sampled: 1


adriancole:
`X-B3-Sampled: 1` is not accepted while `X-B3-Sampled: true` is

### Expected behavior

While a lot of instrumentation leniently read "true" or "false", "1" and "0" are the most often used
ways of indicating tracing is on or off.

https://github.com/openzipkin/b3-propagation#sampled

### Actual behavior

In the http Codec.scala, we use toBoolean, which may have at some point worked with truthy values, but currently does not:

```scala
try Some(request.headerMap(Header.Sampled).toBoolean)
catch { case _: IllegalArgumentException =>
  None
}
```

This causes finagle to suspend traces from other apps, as it interprets "1" as unsampled.

### Steps to reproduce the behavior

You either write a unit test, or put a breakpoint at the catch to notice "1" is not accepted.

adriancole:
@kevinoliver @mosesn I was just about to brag on how I have brave working sharing finagle's context in process. Then I realized that finagle is breaking processes between processes :P

While a workaround is to make an interceptor to rewrite the sampled header when we know downstream is finagle, it would be great to leniently read, or even better switch to 1|0 since most instrumentation use this convention.

adriancole:
cheapest fix is to special case and look for the character '1' as anything else ends up '0' anyway

mosesn:
For now let's leniently read, we can switch to writing 1|0 in a couple releases.  We need to give folks time to upgrade so that we don't break tracing for finagle services that aren't redeployed frequently.

adriancole:
Perfectly sensible. I will raise a PR for 0|1 tolerant reads

adriancole:
https://github.com/twitter/finagle/pull/619

mosesn:
This was fixed here: https://github.com/twitter/finagle/commit/a6de338c0ffb0e7b6924f326cd0feac0bc9dad0a thanks!

adriancole:
thanks for championing it

PR #619, opened 2017-05-12
Tolerates X-B3-Sampled: 1

adriancole:
Problem

Currently, when the http codec reads `X-B3-Sampled: true`, a trace is
continued. Otherwise, it is stopped (as opposed to restarting), as other
values are interpreted as false. Most Zipkin instrumentation send
`X-B3-Sampled: 1` even if they tolerantly read `X-B3-Sampled: true`.

Solution

This change tolerantly reads '1', and also skips allocating an exception
for other values. In a future version of Finagle, we should start
writing `1`, IMHO.

Result

This change makes Finagle tolerantly read `X-B3-Sampled: 1`, so that
traces made by other libraries don't stop when they hit a Finagle
service. For example, this makes Finagle compatible with Brave and zipkin-js

See #617

### DIFF
--- a/finagle-http/src/main/scala/com/twitter/finagle/http/Codec.scala
+++ b/finagle-http/src/main/scala/com/twitter/finagle/http/Codec.scala
@@ -13,7 +13,7 @@ import com.twitter.finagle.stats.{NullStatsReceiver, ServerStatsReceiver, StatsR
 import com.twitter.finagle.tracing.{Flags, SpanId, Trace, TraceId, TraceInitializerFilter}
 import com.twitter.finagle.transport.Transport
 import com.twitter.util.{Closable, StorageUnit}
-import java.lang.{Long => JLong}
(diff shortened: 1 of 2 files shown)


mosesn:
LGTM, thanks for the PR!

mosesn:
@adriancole this was merged in here: https://github.com/twitter/util/commit/93a88444c597841e0a514538b2381b14ddbcf687 thanks for the PR!

adriancole:
thanks for the help getting this in.

adriancole:
wrong commit link I think, though

ryanoneill:
You are correct @adriancole. Think Moses meant to post this one. https://github.com/twitter/finagle/commit/a6de338c0ffb0e7b6924f326cd0feac0bc9dad0a

mosesn:
Whoops, yeah!  My bad, sorry.

PR #624, opened 2017-06-14
Pipelining dispatcher improvements

plaflamme:
Problem

`PipeliningDispatcher` has the following limitations:

* same type signature as the underlying `Transport`;
* hardcoded timeout for detecting a stalled pipeline;
* does not allow tracking more than the dispatched `Promise` in the pipeline.

Solution

Abstract over the dispatcher / transport types, add another type parameter for what is being tracked in the pipeline.

Add a new `Param` to allow configuring the stall timeout.

### DIFF
--- a/finagle-core/src/main/scala/com/twitter/finagle/dispatch/PipeliningDispatcher.scala
+++ b/finagle-core/src/main/scala/com/twitter/finagle/dispatch/PipeliningDispatcher.scala
@@ -6,7 +6,7 @@ import com.twitter.finagle.Failure
 import com.twitter.finagle.stats.StatsReceiver
 import com.twitter.finagle.transport.Transport
 import com.twitter.logging.Logger
-import com.twitter.util.{Future, Promise, Timer, Time}
+import com.twitter.util.{Future, Promise, Time, Timer, Try}
 
 /**
  * A generic pipelining dispatcher, which assumes that servers will
@@ -24,28 +24,28 @@ import com.twitter.util.{Future, Promise, Timer, Time}
  *
  * @param statsReceiver typically scoped to `clientName/dispatcher`
  */
-private[finagle] class PipeliningDispatcher[Req, Rep](
-    trans: Transport[Req, Rep],
+abstract class GenPipeliningDispatcher[Req, Rep, In, Out, Q](
+    trans: Transport[In, Out],
     statsReceiver: StatsReceiver,
     timer: Timer)
-  extends GenSerialClientDispatcher[Req, Rep, Req, Rep](
+  extends GenSerialClientDispatcher[Req, Rep, In, Out](
     trans,
     statsReceiver) { self =>
-  import PipeliningDispatcher._
+  import GenPipeliningDispatcher._
 
   // thread-safety provided by synchronization on this
   private[this] var stalled = false
-  private[this] val q = new AsyncQueue[Promise[Rep]]
+  private[this] val q = new AsyncQueue[Q]
 
   private[this] val queueSize =
     statsReceiver.scope("pipelining").addGauge("pending") {
       q.size
     }
 
-  private[this] val transRead: Promise[Rep] => Unit =
+  private[this] val transRead: Q => Unit =
     p =>
-      trans.read().respond { res =>
-        try p.updateIfEmpty(res)
+      trans.read().respond { out =>
+        try respond(p, out)
         finally loop()
       }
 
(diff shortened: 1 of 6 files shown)


dschobel:
thanks for the pr @plaflamme, lgtm big picture. I'm curious what your use case is if you don't mind saying.

finagle-core/src/main/scala/com/twitter/finagle/dispatch/PipeliningDispatcher.scala
```diff
@@ -54,10 +55,14 @@ private[finagle] class PipeliningDispatcher[Req, Rep](
 
   loop()
 
+  protected def respond(q: Q, out: Try[Out]): Unit
+
+  protected def pipeline(req: Req, p: Promise[Rep]): Future[Q]
+
   // Dispatch serialization is guaranteed by GenSerialClientDispatcher so we
   // leverage that property to sequence `q` offers.
   protected def dispatch(req: Req, p: Promise[Rep]): Future[Unit] =
```
plaflamme: I hesitated to do that and didn't similar to how `apply` is not final in `GenSerialClientDispatcher`. I figured that most use-cases would only need to override `pipeline`, but potentially more complex ones would need to override `dispatch` as well. I don't have a strong opinion, either way is fine by me.

plaflamme:
The use case is a `finagle-kafka` client which we may be able to contribute at some point. The reason why it needs these changes is mostly because Kafka has a `correlation id` associated with each request (basically a unique id) which is sent back to clients in the response. This is a dispatching concern, so I wanted to encapsulate that in there.

The `correlation id` is put into the queue along with the promise and is checked when the response comes from the server. It is required to match head of the queue and then simply thrown away.

In its current state, it also uses different types between the dispatcher and the transport, but that's probably not strictly necessary, just a nice to have.

The configuration of the stall timeout is strictly hygiene (I figured it would be better if it were configurable). So we can drop that commit if it's preferable to minimize the change.

plaflamme:
@dschobel Sorry, I had to rebase the original PR because it wasn't off `develop`.

I added an additional commit which I hope addresses your concerns.

It also enforces correctness of handling the `Promise[Rep]` which was originally left to extending classes. I think this is an improvement, but it came at the cost of instantiating an extra type in the simple case.

Also, an argument to leave `dispatch` non-final is that Kafka actually has an edge case where one particular request does not get a server response. So it would actually need to override `dispatch` to avoid enqueuing an item in the queue for those requests.

An alternative is to return an `Option[T]` from `pipeline` (or some other union type) that `dispatch` would know not to enqueue (say when it's `None`), but it seemed like too much of an edge case to add formal support for. Leaving `dispatch` non-final seemed simpler.

finagle-core/src/main/scala/com/twitter/finagle/dispatch/PipeliningDispatcher.scala
```diff
@@ -85,14 +109,14 @@ private[finagle] class PipeliningDispatcher[Req, Rep](
   }
 }
 
-private object PipeliningDispatcher {
+object GenPipeliningDispatcher {
   val log = Logger.get(getClass.getName)
```
dschobel: `log` can stay private I believe

finagle-core/src/main/scala/com/twitter/finagle/dispatch/PipeliningDispatcher.scala
```diff
@@ -85,14 +109,14 @@ private[finagle] class PipeliningDispatcher[Req, Rep](
...
 
-  val TimeToWaitForStalledPipeline = 10.seconds
+  private case class Pending[T, Rep](value: T, promise: Promise[Rep])
 
-  val StalledPipelineException =
+  def stalledPipelineException(timeout: Duration) =
```
dschobel: I think this one can be private as well

finagle-core/src/main/scala/com/twitter/finagle/dispatch/PipeliningDispatcher.scala
```diff
@@ -103,3 +127,23 @@ private object PipeliningDispatcher {
     }
   }
 }
+
+case class StalledPipelineTimeout(timeout: Duration) {
```
dschobel: docs please

finagle-memcached/src/main/scala/com/twitter/finagle/Memcached.scala
```diff
@@ -1,11 +1,12 @@
 package com.twitter.finagle
 
 import _root_.java.net.SocketAddress
+
```
dschobel: rm space please

finagle-memcached/src/main/scala/com/twitter/finagle/Memcached.scala
```diff
@@ -29,6 +30,7 @@ import com.twitter.hashing
 import com.twitter.io.Buf
 import com.twitter.util.registry.GlobalRegistry
 import com.twitter.util.{Closable, Duration, Monitor}
+
```
dschobel: ditto, no space

plaflamme:
@dschobel I've added a commit with the requested changes. I'll sign the CLA ASAP.

plaflamme:
@dschobel @kevinoliver any guidance on how to solve the test failure? Is it a flake?

dschobel:
@plaflamme looks like it was a flaky run. I'm merging the pr internally and I'll report back if it reappears.

dschobel:
merged in affced891e62428e156f31e2c14dd32121404582, thanks again for the pr @plaflamme

PR #627, opened 2017-06-30
Add Sigma to ADOPTERS.md

omerzach:


omerzach:
Our whole API server is built on Finagle through Finch and we use Finagle clients to call out to our own Node microservices via HTTP/REST and to third-party REST APIs. We're also switching some of those microservices to gRPC, so will be getting that set up pretty soon.

mosesn:
@omerzach oh, interesting, what kind of grpc?  I don't think anyone has finagle speaking the grpc protocol yet except for linkerd right now.  Are you planning on using our h2 clients and interpreting the protobuf/thrift directly?  Would you be interesting in helping to contribute some grpc-compat stuff if you run into any issues?

kevinoliver:
Thanks @omerzach this is merged in https://github.com/twitter/finagle/commit/8ea728af2857ff2c2149aad4b0d595c27c8c1cf7

PR #628, opened 2017-07-11
finagle-redis: Add support for Redis Geo commands

Mura-Mi:
### Problem

finagle-redis does not support geo index operation

### Solution

Implement GEOADD, GEOPOS,  GEOHASH, GEODIST, GEORADIUS and
GEORADIUSBYMEMBER operation in finagle-redis.

### DIFF
--- a/finagle-redis/src/main/scala/com/twitter/finagle/redis/GeoCommands.scala
+++ b/finagle-redis/src/main/scala/com/twitter/finagle/redis/GeoCommands.scala
@@ -1,0 +1,187 @@
+package com.twitter.finagle.redis
+
+import com.twitter.finagle.redis.protocol._
+import com.twitter.io.Buf
+import com.twitter.io.Buf.Utf8
+import com.twitter.util.Future
+
+private[redis] trait GeoCommands {
+  self: BaseClient =>
+
+  /**
+    * Adds the specified geospatial items (latitude, longitude, name) to the specified `key`.
+    *
+    * Redis 3.2.0 or later supports this command.
+    * For detailed API spec, see:
+    *   [[https://redis.io/commands/geoadd]]
+    *
+    * @return Future of the number of elements added to the sorted set
+    */
+  def geoAdd(key: Buf, values: GeoElement*): Future[Long] =
+    doRequest(GeoAdd(key, values: _*)) {
+      case IntegerReply(n) => Future.value(n)
+    }
+
+  /**
+    * Get valid Geohash strings representing the position of elements.
+    *
+    * Redis 3.2.0 or later supports this command.
+    * For detailed API spec, see:
+    *   [[https://redis.io/commands/geohash]]
+    *
+    * @return Future of an array where each element is the Geohash corresponding
+    *         to each member, or `None` if member is missing.
+    */
+  def geoHash(key: Buf, members: Buf*): Future[Seq[Option[Buf]]] =
+    doRequest(GeoHash(key, members: _*)) {
+      case EmptyMBulkReply => Future.Nil
+      case MBulkReply(hashes) => Future.value(hashes.map {
+        case EmptyBulkReply => None
+        case BulkReply(hashBuf) => Some(hashBuf)
+      })
+    }
+
+  /**
+    * Get the positions (longitude,latitude) of all the specified members
+    * of the geospatial index represented by the sorted set at `key`.
+    *
+    * Redis 3.2.0 or later supports this command.
+    * For detailed API spec, see:
+    *   [[https://redis.io/commands/geopos]]
+    *
+    * @return Future of an array where each element is a tuple representing
+    *         longitude and latitude (x,y) of each member, or `None` if member is missing.
+    */
+  def geoPos(key: Buf, members: Buf*): Future[Seq[Option[(Double, Double)]]] = {
+    doRequest(GeoPos(key, members: _*)) {
+      case EmptyMBulkReply => Future.Nil
+      case MBulkReply(positions) => Future.value(positions.map {
+        case MBulkReply(Seq(longitude, latitude)) => (longitude, latitude) match {
+          case (BulkReply(lon), BulkReply(lat)) => unapplyCoordinate(lon,lat)
+        }
+        case _ => None
+      })
+    }
+  }
+
+  private val geoDistResultHandler: PartialFunction[Reply, Future[Option[Double]]] = {
+    case EmptyBulkReply => Future.None
+    case BulkReply(buf) => Future.value(Utf8.unapply(buf).map(_.toDouble))
+  }
+
+  /**
+    * Get the distance between two members in the geospatial index represented by the sorted set.
+    *
+    * Redis 3.2.0 or later supports this command.
+    * For detailed API spec, see:
+    *   [[https://redis.io/commands/geodist]]
+    *
+    * @return Future of the distance in the specified unit,
+    *         or `None` if one or both the elements are missing.
+    */
+  def geoDist(key: Buf, member1: Buf, member2: Buf, unit: Option[GeoUnit] = None): Future[Option[Double]] = {
+    doRequest(GeoDist(key, member1, member2, unit))(geoDistResultHandler)
+  }
+
+  /**
+    * Get the members of a sorted set which are within the borders of the area specified
+    * with the center location * and the maximum distance from the center (the radius).
+    *
+    * Redis 3.2.0 or later supports this command.
+    * For detailed API spec, see:
+    *   [[https://redis.io/commands/georadius]]
+    *
+    * @return Future of an array where each member represents element.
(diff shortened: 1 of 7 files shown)


kevinoliver:
@Mura-Mi thanks for the patch/feature! gonna try to find someone who knows finagle-redis better than i to take a look at this.

finagle-redis/src/main/scala/com/twitter/finagle/redis/protocol/commands/Geo.scala
```diff
@@ -0,0 +1,130 @@
+package com.twitter.finagle.redis.protocol
+
+import com.twitter.io.Buf
+import com.twitter.io.Buf.Utf8
+
+case class GeoAdd(key: Buf, values: (Double, Double, Buf)*)
```
nepthar: Make a case class for (Double, Double, Buf) instead of a tuple for clarity

finagle-redis/src/main/scala/com/twitter/finagle/redis/GeoCommands.scala
```diff
@@ -0,0 +1,163 @@
...
+    */
+  def geoRadius(key: Buf,
+                longitude: JDouble,
+                latitude: JDouble,
+                radius: JDouble,
+                unit: GeoUnit,
+                withCoord: JBoolean = false,
+                withDist: JBoolean = false,
+                withHash: JBoolean = false,
+                count: JInt = null,
```
nepthar: How about we avoid nulls here and just use Option directly?

finagle-redis/src/main/scala/com/twitter/finagle/redis/GeoCommands.scala
```diff
@@ -0,0 +1,163 @@
...
+
+private[redis] trait GeoCommands {
+  self: BaseClient =>
+
+  /**
+    * Adds the specified geospatial items (latitude, longitude, name) to the specified `key`
+    *
+    * @return Future of the number of elements added to the sorted set
+    */
+  def geoAdd(key: Buf, values: (Double, Double, Buf)*): Future[JLong] =
```
vkostyukov: Why this has to be Java `Long`?

finagle-redis/src/main/scala/com/twitter/finagle/redis/GeoCommands.scala
```diff
@@ -0,0 +1,163 @@
...
+    }
+
+  /**
+    * Get the positions (longitude,latitude) of all the specified members
+    * of the geospatial index represented by the sorted set at `key`
+    *
+    * @return Future of an array where each element is a tuple representing
+    *         longitude and latitude (x,y) of each member, or `None` if member is missing.
+    */
+  def geoPos(key: Buf, members: Buf*): Future[Seq[Option[(JDouble, JDouble)]]] = {
```
vkostyukov: Why using Java `Double`s?

finagle-redis/src/main/scala/com/twitter/finagle/redis/GeoCommands.scala
```diff
@@ -0,0 +1,163 @@
...
+    doRequest(GeoPos(key, members: _*)) {
+      case EmptyMBulkReply => Future.Nil
+      case MBulkReply(positions) => Future.value(positions.map {
+        case MBulkReply(Seq(longitude, latitude)) => (longitude, latitude) match {
+          case (BulkReply(lon), BulkReply(lat)) =>
```
vkostyukov: You can inline `Buf.Utf8(lon)` in this pattern matching to get the string out of `BulkReply(lon)`. This way, you won't need to call  `BufToString` later.

finagle-redis/src/main/scala/com/twitter/finagle/redis/GeoCommands.scala
```diff
@@ -0,0 +1,163 @@
...
+            Option((double2Double(BufToString(lon).toDouble), double2Double(BufToString(lat).toDouble)))
+        }
+        case _ => None
+      })
+    }
+  }
+
+  private lazy val geoDistResultHandler: PartialFunction[Reply, Future[Option[JDouble]]] = {
```
vkostyukov: Similar question about `JDouble`.

finagle-redis/src/main/scala/com/twitter/finagle/redis/GeoCommands.scala
```diff
@@ -0,0 +1,163 @@
...
+    case EmptyBulkReply => Future.None
+    case BulkReply(buf) => Future.value(Utf8.unapply(buf).map(_.toDouble))
```
vkostyukov: Please inline `Buf.Utf8()` extractor here.

finagle-redis/src/main/scala/com/twitter/finagle/redis/GeoCommands.scala
```diff
@@ -0,0 +1,163 @@
...
+  }
+
+  /**
+    * Get the distance between two members in the geospatial index represented by the sorted set.
+    *
+    * @return Future of the distance in the specified unit,
+    *         or `None` if one or both the elements are missing.
+    */
+  def geoDist(key: Buf, member1: Buf, member2: Buf, unit: GeoUnit = null): Future[Option[JDouble]] = {
```
vkostyukov: I'd prefer this w/o nulls. Also why `JDouble`?

finagle-redis/src/main/scala/com/twitter/finagle/redis/GeoCommands.scala
```diff
@@ -0,0 +1,163 @@
...
+    case EmptyBulkReply => None
+    case BulkReply(member) => Some(GeoRadiusResult(member))
+    case MBulkReply(attrs) => attrs match {
+      case BulkReply(member) :: rest =>
+        Some((rest map {
+          case IntegerReply(hash) => (r: GeoRadiusResult) => r.copy(hash = Some(hash.toInt))
+          case BulkReply(dist) => (r: GeoRadiusResult) => r.copy(dist = Some(BufToString(dist).toDouble))
+          case MBulkReply(BulkReply(lon) :: BulkReply(lat) :: Nil) =>
+            (r: GeoRadiusResult) =>
+              r.copy(coord = Some((BufToString(lon).toDouble, BufToString(lat).toDouble)))
```
vkostyukov: Please, inline `Buf.Utf8()` extractor instead.

finagle-redis/src/main/scala/com/twitter/finagle/redis/protocol/commands/Geo.scala
```diff
@@ -0,0 +1,130 @@
...
+  * Redis' geospatial commands.
+  */
+sealed abstract class GeoUnit(str: String) {
+  lazy val toBuf: Buf = Utf8(str)
+}
+
+object GeoUnit {
+
+  /** Meter */
+  case object m extends GeoUnit("m")
```
vkostyukov: What do you think about `GeoUnit.Meter`, `GeoUnit.Mile`, etc?

Mura-Mi: It's good. I've named like these as described in Redis document, but I'll modify

nepthar:
@Mura-Mi this looks good to me once @vkostyukov's comments about nulls/java types are addressed. Thanks!

vkostyukov:
@Mura-Mi Thanks for working on this! I still think we don't need to use Java types directly. Also, please, don't use `BufToString`. Try `Buf.Utf8()` extractor instead. Other than that, this looks good to me.

bryce-anderson:
@Mura-Mi, are you still interested in getting this merged? This looks pretty close, but there appears to be a few outstanding comments.

Mura-Mi:
Thank you for reviewing my pull request. I've removed the usage of Java primitive numbers and `BufToString` converter.

In TravisCI, latest redis-server is not available so the Geo API Redis integration test fails. I'm trying to install `redis-server` via apt-package. Is this acceptable?

mkhq:
@Mura-Mi would it be possible to move the integration tests into finagle-redis/src/it? All redis integration tests were moved there recently. Run in sbt with `it:test`.

yannick-polius:
@mkhq @Mura-Mi I had an internal reason to implement a similar PR. My version took a bit from this PR and is on the internal code review. I moved the integration test into finagle-redis/src/it.

Do we want to decide if we want to use this one or the other code review?

mkhq:
@yannick-polius Is this PR replaced by https://github.com/twitter/finagle/commit/d32d123665fab5bada54a653029d60a88a29b1a6 ?

yannick-polius:
@mkhq  please see [269e181](https://github.com/twitter/finagle/commit/269e18150f776c640745d5cc27b16211f6629678) for attribution

PR #629, opened 2017-07-12
Scope dispatcher stats under client label

adleong:
Problem

Dispatcher stats are not scoped under the client label.  This is inconsistent
with all other client stats.

Solution

Modify the endpointer module to use the StatReceiver from the Stack.Params
passed into the module rather than the Stack.Params set on the StackClient.
These are mostly identicial except that `newClient` scopes the StatsReceiver
with the label and passes it into the Stack but does not update the
Stack.Params on the StackClient.

Result

Dispatcher stats are scoped under the client label.

### DIFF
--- a/finagle-http/src/main/scala/com/twitter/finagle/Http.scala
+++ b/finagle-http/src/main/scala/com/twitter/finagle/Http.scala
@@ -193,7 +193,7 @@ object Http extends Client[Request, Response] with HttpRichClient
 
     protected def endpointer: Stackable[ServiceFactory[Request, Response]] =
       new EndpointerModule[Request, Response](
-        Seq(implicitly[Stack.Param[HttpImpl]]),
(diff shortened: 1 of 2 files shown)


vkostyukov:
I think this is correct. Although this changes the metric location people might be altering on / monitoring (I honestly doubt that though). That said, I think it's fine to merge it but let's wait what others think about that.

nepthar:
@adleong in investigating this internally, we found that most (but not all) of our clients do, in fact, have the dispatcher metric in the right place. Before I go digging, do you have any insight as to why this might happen?

gpevnev:
Hi, why wouldn't you finally merge it into open-source version? As I am guessing from labels this is already merged internally more than year ago, why would't you do it on github too?

adleong:
https://github.com/twitter/finagle/commit/3e3ab1926845890687556129133e3540c219c2c1

gpevnev:
Yes, seems it is just strange concurrency bug, and dispatcher metric is just occasional victim.

PR #631, opened 2017-07-17
fix sbt deprecated warnings

matsu-chara:
Problem

Some sbt operators are deprecated

Solution

fix warnings.

Result

no warn (without Build trait deprecation)


yufangong:
@matsu-chara Thank you! Merged https://github.com/twitter/finagle/commit/dffc7d3a58b340480ab6dce03023a318c4da5bf9

PR #633, opened 2017-07-22
Migrate to build sbt

matsu-chara:
Problem:

Build trait was deprecated

Solution:

- step1: move project/Build.scala to build.sbt
- step2.1: remove Build trait
- step2.2: remove unnecessary import(sbt, Keys, and some plugins)
- step3: remove unnecessary settings
  - Defaults.coreDefaultSettings which is loaded by CorePlugin
  - JmhPlugin.projectSettings which is loaded by JmhPlugin

Result:
  Migration is completed


mosesn:
wow, this is seriously cool!  thanks so much for tackling this.

vkostyukov:
I think this is awesome! Thanks @matsu-chara!

mosesn:
@matsu-chara this was merged internally, but because of an internal breaking API change, it's not safe to publish right now, so I'm going to leave this ticket open.  Thanks for your contribution!

mosesn:
@matsu-chara this finally made it to github, thanks!  https://github.com/twitter/finagle/commit/0fd63b01b646a0a04147dedc4e7568ff1e0c9fb4

PR #638, opened 2017-08-08
Initial support for Curator Service Discovery records

joeyb:
I suspect a final implementation of this will look a bit different, but the code I've included here can be treated as a minimal straw man for solving the problem.

At Salesforce we have a mixed service discovery environment, with some services announcing via [Curator's Discovery Service](http://curator.apache.org/curator-x-discovery/) and others using Finagle's serversets. Currently we're maintaining an internal Finagle fork with the changes necessary to consume those records, but it would be great if we could get those changes merged back into mainline.


ryanoneill:
Hi @joeyb, thank you for the pull request. I'm going to have someone who is a little more familiar with the code than I am review it.

Issue #637, opened 2017-08-08
finagle-mysql: CLIENT_FOUND_ROWS


wpK:
Right now it doesn't seem possible to change [Capability.FoundRows](https://github.com/twitter/finagle/blob/develop/finagle-mysql/src/main/scala/com/twitter/finagle/mysql/Capability.scala#L5).

Is there any plans on the future to allow modifying the client capabilities when creating a client?

More information for those that are not aware of what _CLIENT_FOUND_ROWS_ changes:

For [INSERT ... ON DUPLICATE KEY UPDATE](https://dev.mysql.com/doc/refman/5.7/en/insert-on-duplicate.html) statements, the affected-rows value per row is 1 if the row is inserted as a new row, 2 if an existing row is updated, and 0 if an existing row is set to its current values. If you specify the _CLIENT_FOUND_ROWS_ flag, the affected-rows value is 1 (not 0) if an existing row is set to its current values.

ryanoneill:
Hi @wpK, 

I don't believe we have any current plans here, but thanks for letting us know that you would like to use this feature. Is this something that you would be interested in working on and adding to finagle-mysql via a pull request?

Thanks, Ryan

wpK:
Sure. I'll send over a pull request shortly.

nepthar:
Fixed w/ PR #640

PR #640, opened 2017-08-09
finagle-mysql: ability to toggle CLIENT_FOUND_ROWS flag

wpK:
**Summary:** Added `withAffectedRows()` function to the MySQL client which when used, will disable the `CLIENT_FOUND_ROWS` flag.

**Note: This change was made on the master branch since there was a dependency issue with the scrooge-sbt-plugin.**

**Problem**

Original issue: #637

Unable to change the `CLIENT_FOUND_ROWS` flag. The flag (which `finagle-mysql` enables by default) changes the behavior of affected-rows for `UPDATE` and `INSERT ... ON DUPLICATE KEY UPDATE` statements.

From the [manual](https://dev.mysql.com/doc/refman/5.7/en/information-functions.html#function_found-rows) (v5.7):

For [UPDATE](https://dev.mysql.com/doc/refman/5.7/en/update.html) statements, the affected-rows value by default is the number of rows actually changed. If you specify the `CLIENT_FOUND_ROWS` flag to [mysql_real_connect()](https://dev.mysql.com/doc/refman/5.7/en/mysql-real-connect.html) when connecting to [mysqld](https://dev.mysql.com/doc/refman/5.7/en/mysqld.html), the affected-rows value is the number of rows “found”; that is, matched by the `WHERE` clause.

For [INSERT ... ON DUPLICATE KEY UPDATE](https://dev.mysql.com/doc/refman/5.7/en/insert-on-duplicate.html) statements, the affected-rows value per row is 1 if the row is inserted as a new row, 2 if an existing row is updated, and 0 if an existing row is set to its current values. If you specify the `CLIENT_FOUND_ROWS` flag, the affected-rows value is 1 (not 0) if an existing row is set to its current values.

**Solution**

Added `withAffectedRows()` on `Mysql.client` which when used removes the `FoundRows` capability when connecting to the server.

My implementation adds another argument to the `Handshake` case class to follow how `ConnectWithDB` is handled. I've also considered checking for the param in `Handshake.apply` and if added passing it with `Capability.baseCap - Capability.FoundRows`.

### DIFF
--- a/finagle-mysql/src/main/scala/com/twitter/finagle/Mysql.scala
+++ b/finagle-mysql/src/main/scala/com/twitter/finagle/Mysql.scala
@@ -182,6 +182,14 @@ object Mysql extends com.twitter.finagle.Client[Request, Result] with MysqlRichC
     def withCharset(charset: Short): Client =
       configured(Handshake.Charset(charset))
 
+    /**
+      * Don't set the CLIENT_FOUND_ROWS flag when connected to the
+      * server. This will make "INSERT ... ON DUPLICATE KEY UPDATE"
+      * statements return the "correct" update count.
+      */
+    def withAffectedRows(): Client =
+      configured(Handshake.FoundRows(false))
+
     // Java-friendly forwarders
     // See https://issues.scala-lang.org/browse/SI-8905
     override val withSessionPool: SessionPoolingParams[Client] =
--- a/finagle-mysql/src/main/scala/com/twitter/finagle/Mysql.scala
+++ b/finagle-mysql/src/main/scala/com/twitter/finagle/Mysql.scala
@@ -183,9 +183,11 @@ object Mysql extends com.twitter.finagle.Client[Request, Result] with MysqlRichC
       configured(Handshake.Charset(charset))
 
     /**
-      * Don't set the CLIENT_FOUND_ROWS flag when connected to the
-      * server. This will make "INSERT ... ON DUPLICATE KEY UPDATE"
+      * Don't set the CLIENT_FOUND_ROWS flag when establishing a new
+      * session. This will make "INSERT ... ON DUPLICATE KEY UPDATE"
       * statements return the "correct" update count.
+      *
+      * See https://dev.mysql.com/doc/refman/5.7/en/information-functions.html#function_row-count
       */
     def withAffectedRows(): Client =
       configured(Handshake.FoundRows(false))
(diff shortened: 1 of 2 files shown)


finagle-mysql/src/main/scala/com/twitter/finagle/Mysql.scala
```diff
@@ -182,6 +182,14 @@ object Mysql extends com.twitter.finagle.Client[Request, Result] with MysqlRichC
     def withCharset(charset: Short): Client =
       configured(Handshake.Charset(charset))
 
+    /**
+      * Don't set the CLIENT_FOUND_ROWS flag when connected to the
+      * server. This will make "INSERT ... ON DUPLICATE KEY UPDATE"
+      * statements return the "correct" update count.
+      */
```
nepthar: Add a link to relevant mysql docs please.

finagle-mysql/src/main/scala/com/twitter/finagle/mysql/Handshake.scala
```diff
@@ -89,6 +100,9 @@ object Handshake {
  *
  * @param charset default character established with the server.
  *
+ * @param enableFoundRows if the server should return the number
+ * of found (matched) rows, not the number of changed rows.
```
nepthar: add something like "for UPDATE and INSERT ... ON DUPLICATE KEY UPDATE type statements"

wpK:
@nepthar Updated the comments with your suggestions.

> Where will this option be translated to the appropriate connection flag?

Not sure what you mean. Can you elaborate?

The flag is changed [here](https://github.com/twitter/finagle/pull/640/commits/0470f79d790ae670dbee54808ec8bd46ac0c772a#diff-f75d661e18911280d578af7ccf8ff0d3R133) and the client can disable the flag like so:

```
Mysql.client
  .withCredentials("test", "test")
  .withDatabase("test")
  .withAffectedRows()
  .newRichClient("localhost:3306")
```
I originally had `enableFoundRows(enabled: Boolean)` instead of `withAffectedRows()`. I went with `withAffectedRows`() which is closer to the the JDBC as seen [here](https://dev.mysql.com/doc/connector-j/5.1/en/connector-j-reference-configuration-properties.html) (you'll need to search for `useAffectedRows`.

kevinoliver:
Thanks again @wpK 

This got merged in https://github.com/twitter/finagle/commit/fae2e69d81faeef7cf0dee0340cb4ec27d2bcf10

Issue #360, opened 2015-04-06
Set finagle-redis tests integration tests up to run as integration tests in SBT and Travis CI


travisbrown:
We're currently excluding the finagle-redis integration tests in the SBT build. We should move them to the appropriate location, configure SBT to treat them as integration tests, and set up the Travis CI build to have a Redis server available to run them. (If this has a large impact on the Travis CI testing times we might need to turn them off there, but they currently run pretty quickly on my local machine.)

smlance:
The integration tests have a Pants target as of https://github.com/twitter/finagle/commit/0f96f6ca3b6d0ad7dba4233af43b80a9abe19ce3 but they don't yet run in Travis.

PR #642, opened 2017-08-11
finagle-redis: Enable running integration tests in finagle-redis.

mkhq:
Problem

Unit tests and integration tests are mixed in the finagle-redis project
under the test path. In addition, the SBT configuration is only running
unit tests. This solves issue #360.

Solution

Added configuration for IntegrationTest in the finagle-redis project
and moved integration tests to the correct path. Redis process shutdown
always includes all processes.

Result

Running 'test' only executes unit tests. Running 'it:test' executes the
integration tests.


mkhq:
Note that some of the integration tests rely on twitter internal modifications to redis and will fail if started with the public redis server.

cacoco:
@mkhq LGTM. We can work on getting this merged.

bryce-anderson:
This has been merged as commit 7d5f269. Thanks for the PR, and we'll try to get the follow-ups reviewed soon.

PR #645, opened 2017-08-17
finagle-redis: Redis cluster management

mkhq:
Problem

The current finagle-redis client does not have support for the CLUSTER commands introduced in Redis 3.0.0.

Solution

This adds support for a number of new cluster commands that are necessary to setup and manage a redis cluster.

- CLUSTER ADDSLOT, assigns a storage slot to a server
- CLUSTER SETSLOT, updates the state of a slot
- CLUSTER SLOTS, returns a list of slots known to the server
- CLUSTER GETKEYSINSLOT, returns keys stored in a slot
- CLUSTER REPLICATE, changes a node to become a replica of a server
- CLUSTER MEET, introduces cluster nodes to each other
- CLUSTER NODES, returns the list of cluster nodes known to the server
- CLUSTER INFO, returns the cluster specific info
- MIGRATE, used to transfer keys to another node

Result

Using the introduced commands it is possible to setup a redis cluster for testing and live re-sharding. This is the first step in implementing the redis cluster client protocol. The next step is to handle operation redirects, i.e. when the server returns ASK or MOVED.


mkhq:
Note that this PR includes #642

bryce-anderson:
Looks good to me, but this is honestly the first time I've looked at redis technology so I'm far from an expert. If we can get another set of eyes or two, I think it's ready to go. Ping @vkostyukov, who I believe knows more than I do about this protocol implementation.

finagle-redis/src/main/scala/com/twitter/finagle/redis/ClusterCommands.scala
```diff
@@ -0,0 +1,230 @@
...
+      reader.remainingUntil('\r') match {
+        case -1 => acc
+        case n =>
+          val newAcc = reader.readString(n, UTF_8).split(":") match {
+            case Array(key, value) => acc + (key -> value)
+            case _ => acc
+          }
+
+          // jump past the '\r\n'
+          reader.skip(2)
```
vkostyukov: How do we know `\n` follows `\r`? I don't see a check for that.

mkhq: Its an assumption from the protocol: https://redis.io/topics/protocol#bulk-string-reply

mkhq:
@vkostyukov Thanks for the review! I'll fix the visibility of ClusterClient.

In the meantime I've been working on another PR for transparently handling cluster slot redirect and management. Will try to submit soonish.

bryce-anderson:
@mkhq, perhaps you're already going to do this, but it would be helpful if you could rebase+squash the PR when you push the commit for visibility. (It makes it easier to merge internally, but its not a show stopper)

mkhq:
@bryce-anderson Hey, I already merged in master in the branch so rebase+squash would be a bit of a mess. Should I create a new branch/PR?

mkhq:
Ping @mosesn

mosesn:
@mkhq yeah, a new branch/PR would be great, if you can.  Sorry for the delay!

mkhq:
@mosesn Replaced by https://github.com/twitter/finagle/pull/668

bryce-anderson:
@mkhq, I'm really sorry I didn't see your previous ping: a lot of things got lost in my inbox during that time period.

mkhq:
@bryce-anderson no worries :) I didnt have time to fix it until now

PR #648, opened 2017-08-24
finagle-redis: Redis cluster client with redirect detection

mkhq:
Problem

The current redis cluster client does not handle redis server redirect errors, ASK, TRYAGAIN and MOVED.

Solution

This introduces a ClusterClient that extends the RedisBaseClient by returning exceptions when a Redis cluster server responds with a redirection error. 

Result

The ClusterClient now handles redis cluster redirection errors. In addition, the integration test suite has been extended in order to test all cases resulting in redirection errors. The new version of the client can be used to implement transparent handling of redirects.


mkhq:
This depends on #642 and #645.

mkhq:
Closing this and will create a new PR with redirect client support.

PR #662, opened 2017-11-04
Scalatest 3.0.0 --> 3.0.4

sullis:


mosesn:
Unfortunately, we need to keep scalatest parity with the version of scalatest that twitter uses, so I don't think we can accept this PR.  Scala 2.12.x and sbt are OK because we don't run either of those internally (scala 2.12.x is a work in progress).  However, thanks for the reviews!  Keep 'em coming.

PR #663, opened 2017-11-20
Ensures 128bit trace IDs are propagated in outbound http requests

adriancole:
Problem

When `com.twitter.finagle.tracing.traceId128Bit=true` outbound requests don't include the entire trace ID, rather only the lower 64 bits. This can result in broken traces, depending on if Zipkin is set to operate tolerantly.

Solution

This was due to a missed change where we missed setting the trace ID header according to `traceIdHigh`. This fixes that and backfills tests

Result

`X-B3-TraceId` values should be 32 lowerhex length when `com.twitter.finagle.tracing.traceId128Bit=true` and a new trace is sent downstream

See #651

### DIFF
--- a/finagle-base-http/src/main/scala/com/twitter/finagle/http/TraceInfo.scala
+++ b/finagle-base-http/src/main/scala/com/twitter/finagle/http/TraceInfo.scala
@@ -67,7 +67,12 @@ private object TraceInfo {
     removeAllHeaders(request.headerMap)
 
     val traceId = Trace.id
-    request.headerMap.add(Header.TraceId, traceId.traceId.toString)
+    val traceIdString = if (traceId.traceIdHigh.isEmpty) {
+      traceId.traceId.toString
+    } else {
+      traceId.traceIdHigh.get.toString + traceId.traceId.toString
+    }
+    request.headerMap.add(Header.TraceId, traceIdString)
     request.headerMap.add(Header.SpanId, traceId.spanId.toString)
     // no parent id set means this is the root span
     traceId._parentId match {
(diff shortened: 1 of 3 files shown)


finagle-base-http/src/main/scala/com/twitter/finagle/http/TraceInfo.scala
```diff
@@ -81,7 +86,9 @@ private object TraceInfo {
         request.headerMap.add(Header.Sampled, sampled.toString)
       case None => ()
     }
-    request.headerMap.add(Header.Flags, JLong.toString(traceId.flags.toLong))
+    if (traceId.flags.toLong != 0L) {
```
adriancole: ps I added this check because in the file and noticed we were burning some headers for no benefit.

adriancole:
ps I tested this manually with zipkin-finagle and this change (which enables 128-bit there). Once something like this PR is in, I'll help show folks how to use 128-bit stuff.
https://github.com/openzipkin/zipkin-finagle/pull/39

adriancole:
Any chance this could make it into a release soon? Would be nice considering there are a lot of events happening in the next two weeks.

kevinoliver:
@adriancole yeah i think there's a good chance. we can try to get it merged in and then it'll go out in the december release.

isabelmartin:
Thank you @adriancole! This has been merged in [76b6c9a](https://github.com/twitter/finagle/commit/76b6c9a90e4e08b2b97941266f02786fdb13c9a9).

adriancole:
great. thanks Isabel!

PR #666, opened 2017-12-05
sbt 1.0.4

sullis:


kevinoliver:
cc @mosesn

mosesn:
@sullis is there an sbt 1.0.4 feature you'd like to use?  we prefer to keep our sbt versions in sync, so if there isn't anything you need, we might not want to upgrade this immediately.

sullis:
SBT 1.0.4 fixes a performance issue in SBT 1.0.3
https://developer.lightbend.com/blog/2017-11-27-sbt-1-0-4-hotfix-and-performance-fixes/

cacoco:
@sullis sbt 1.0.4 introduces a scala compilation regression that fails for one of our other projects ([Scrooge](https://github.com/twitter/scrooge)). The regression was patched in 1.0.3 but looks like they didn't apply it to 1.0.4 before releasing. 

This is the issue: https://github.com/sbt/zinc/issues/127

sullis:
obsolete

PR #670, opened 2018-01-13
Don't calculate the hashcode of Stream in Budget

### LINKED ISSUE #669: Hashcode calculation of http.client never terminates
Calculating hashcode of the Http.Client never terminates when having a retrywithbackoff (infinite stream) set. 

### Expected behavior

```scala
val client = Http.client
println(s"$client with hashcode ${client.hashCode}")

val client2 = client.withRetryBackoff(Backoff.const(10.seconds))
println(s"$client2 with hashcode ${client2.hashCode}")
```
the second client with Backoff stream, will have a valid hashcode in finite time.

### Actual behavior

Calling hashcode on the second client, with backoff stream, will evaluate the infinite stream and thus never terminates.

See snippet above.

DieBauer:
# Problem

Calculating the hashcode of an instance of Retries.Budget could end up
in an infinite computation.
Since the case class gave us a default implementation including all
fields, including the requeueBackoffs field with a possibly infinite
Stream[Duration], we need to override this behaviour.

# Solution

This is done by ignoring the fact that there is a field with a Stream.
The equality of the class is implemented with
http://www.artima.com/lejava/articles/equality.html in mind.

# Result

```scala
val client2 = Http.client.withRetryBackoff(Backoff.const(10.seconds))
println(s"$client2 with hashcode ${client2.hashCode}")
```
successfully returns a hashcode, fixing the 'bug' behaviour.
Closes #669.

### DIFF
--- a/finagle-core/src/main/scala/com/twitter/finagle/service/Retries.scala
+++ b/finagle-core/src/main/scala/com/twitter/finagle/service/Retries.scala
@@ -60,6 +60,28 @@ object Retries {
 
     def mk(): (Budget, Stack.Param[Budget]) =
       (this, Budget)
+
+    def canEqual(other: Any): Boolean = other.isInstanceOf[Budget]
+
+    /**
+      * The equals method only compares the retryBudget field of this instance,
+      * since the requeueBackoffs Stream[Duration] is possibly infinite,
+      * resulting in an infinite computation to compare for equality.
+      */
+    override def equals(other: Any): Boolean = other match {
+      case that: Budget =>
+        (that canEqual this) &&
+          retryBudget == that.retryBudget
+      case _ => false
+    }
+
+    /**
+      * For the hashCode of this class we only take into account the hashCode of the RetryBudget field.
+      * The requeueBackoffs Stream[Duration] is possibly infinite,
+      * resulting in an infinite computation to get to the hashCode.
+      */
+    override def hashCode(): Int =
+      retryBudget.##
   }
 
   object Budget extends Stack.Param[Budget] {
--- a/finagle-core/src/test/scala/com/twitter/finagle/service/RetriesTest.scala
+++ b/finagle-core/src/test/scala/com/twitter/finagle/service/RetriesTest.scala
@@ -423,4 +423,18 @@ class RetriesTest extends FunSuite {
 
     factory()
   }
+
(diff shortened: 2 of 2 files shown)


DieBauer:
@bryce-anderson I feel the same regarding dropping the field in an equality, that is part of the unique instance. 
However, comparing infinite things doesn't make sense, it has no meaning. 

As you mention this is indeed to only cater for the issue I raised. A use-case could be putting clients in a HashMap (requiring a valid/unique hashCode). 
Having a valid equals/hashCode on the Client does make sense. And would solve (circumvent) this issue. 
What would you suggest is part of the hashCode then? Is ignoring the certain Params a better option than partly checking the equality of 'deeper' params (ie. Budget)?

You could as an alternative decide to drop the hashCode/equals class specific implementation (given by the case class) and delegate to `super` -> `java.lang.Object` for `Budget` or the `Client`, automatically giving you the instance equality.

bryce-anderson:
It seems I've gotten my terminology a bit twisted. `Http.client` isn't so much a client as a client factory, and it does kind of make sense to have `equality` and `hashCode` there since it is arguably just configuration data, and configurations seem reasonable to compare.

I was arguing for defining equality and hashcode as you've laid out at the end, but I may walk it back. I still think that the equality of `Budget` should probably consider the backoffs, but that would entail trying to detect cycles in the `Stream` which is not great.

I'm actually okay with this as is but more feedback would be helpful.

And thanks for the PR!

ryanoneill:
I've been chatting with @bryce-anderson about this offline. 

From my perspective, it's important that client.hashCode terminates, especially in the case of a properly constructed Finagle client / server. We do use `hashCode` internally and I'm surprised we haven't run into this before.

However, I'm not sold on comparing the equality of clients though, considering we have `StackParams` which take arguments that can have non-terminating `hashCode` values, as in this example. I do believe this is an appropriate fix too given the situation.

ryanoneill:
Thanks for the PR. I will work on getting this pulled in internally and reviewed.

DieBauer:
@ryanoneill alright! Let me know if I need to do something on this PR :)

ryanoneill:
Hi @DieBauer, 

Thanks again for the pull request. This has been submitted internally and pushed back out to GitHub as commit https://github.com/twitter/finagle/commit/88e7bea9bf29bc5fed3207782bf82b25bf869c4a

PR #671, opened 2018-01-25
add tradeshift to the list of adopters

leonmaia:
Add Tradeshift as an adopter.


bryce-anderson:
Closed via a376dc91f7cbec1276ee5114cd89b9e9a50bac69.

leonmaia:
@bryce-anderson sure :) We are breaking down a huge monolith into new services to be used both internally and externally (can't give more context than that, unfortunately). I basically chose it because of the great experience that I had before at Globo and other companies :) Thanks for the great work you guys are doing.

PR #678, opened 2018-02-21
Update netty versions

### LINKED ISSUE #665: finagle is potentionally vulnerable to CVE-2015-2156
### Expected behavior

There are no CVEs in `finagle`'s dependencies.

### Actual behavior

There is (at least) one CVE in `finagle`'s dependencies. [CVE-2015-2156](https://nvd.nist.gov/vuln/detail/CVE-2015-2156)

Compile and use the latest version of `finagle`.

### Steps to fix

Update the 3.x and 4.x versions`netty` to their latest.

### Note

I am not particularly familiar with `finagle`, so this feature of `netty` might not be used anywhere, and I don't have the bandwidth to investigate, so I'm passing it along to y'all for triage.

heartsucker:
## Problem

Finagle is potentially vulnerable to CVE-2015-2156.

## Solution

Bump the `netty` versions to versions that are unaffected by the CVE.

Modify the code where there were incompatibilities with class inheritance.

Fixes #665

## Result

Finagle will no longer be affected by the CVE.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -6,7 +6,7 @@ val releaseVersion = "18.3.0-SNAPSHOT"
 
 val libthriftVersion = "0.5.0-7"
 
-val netty4Version = "4.1.16.Final"
+val netty4Version = "4.1.21.Final"
 
 // zkVersion should be kept in sync with the 'util-zk' dependency version
 val zkVersion = "3.5.0-alpha"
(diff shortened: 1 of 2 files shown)


mosesn:
@heartsucker thanks for the PR!  Unfortunately, we've tried updating the netty 3 version, and we ran into some issues around the way that cookie behavior changed.  We're in the process of moving finagle to a point where we'll have parity in behavior in netty 4, but it will still take a bit more time.  cc @jcrossley who knows more about this stuff.

In terms of updating the netty 4 version, we're nervous about changing the netty 4 version while we're gaining confidence in our http/2 implementation, since codec-http2 is still in flux.  So we've frozen the netty 4 version for the next few weeks, until we have more time to really nail it and get a decent number of services running http/2 in a production environment.

I'm sorry the CVE-2015-2156 mitigation has taken so long, we're chugging away at it though!

heartsucker:
@mosesn Thanks for the quick response. If this collides with current development, feel free to let it dangle until it can be merged in to someone supersedes it. :)

mosesn:
@heartsucker sounds like a plan.  Thanks for bearing with us!

adleong:
Hey @mosesn, is the netty4 version still frozen?  We'd really like to pick up a bug fix from Netty 4.1.17

dschobel:
https://github.com/twitter/finagle/commit/5f34657bff3e6a8ab9bf2f4f48b8cca734a039c6 moved finagle-netty4 to 4.1.28.

finagle-netty3 remains vulnerable to the CVE but I think it'll be a race between our killing it off entirely and that version getting bumped again.

PR #685, opened 2018-03-24
serversets: Preserve host/ip provided in serveIface

### LINKED ISSUE #684: Wrong Host in Zookeeper when using Kubernetes
I'm deploying Finagle ThriftMux with Zookeeper ServerSets inside Kubernetes and pod hostname is being set as the host in Zookeeper instead of the IP of the pod which is what I'm binding to when initing the ThrifMux server.
 
### Expected behavior
The host I use to bind (ie. 10.x.x.x) should be the host that appears in Zookeeper

### Actual behavior
I bind to 10.x.x.x.x when starting ThriftMux Server and the Pod's Host Name is placed in Zookeeper which is not resolvable by other pods in the cluster

Looks like its caused because of calling `getHostName` on the bound inet address instead of `getHostString` which returns the desired IP. https://github.com/twitter/finagle/blob/246be7e560e5089396335462b0585214e2f75983/finagle-serversets/src/main/java/com/twitter/finagle/common/zookeeper/ServerSets.java#L81

GEverding:
Problem

I'm deploying Finagle ThriftMux with Zookeeper ServerSets inside Kubernetes and pod hostname is being set as the host in Zookeeper instead of the IP of the pod which is what I'm binding to when initing the ThrifMux server.

Solution

In k8s `getHostName` returns the name of the pod which is not 
accessible from other pods in the cluster. `getHostString` 
returns the IP of the pod.

Fixes #684

### DIFF
--- a/finagle-serversets/src/main/java/com/twitter/finagle/common/zookeeper/ServerSets.java
+++ b/finagle-serversets/src/main/java/com/twitter/finagle/common/zookeeper/ServerSets.java
@@ -78,6 +78,6 @@ public final class ServerSets {
    * @param address the target address to create the endpoint for
    */
   public static Endpoint toEndpoint(InetSocketAddress address) {
-    return new Endpoint(address.getHostName(), address.getPort());
+    return new Endpoint(address.getHostString(), address.getPort());
   }
 }


finagle-serversets/src/main/java/com/twitter/finagle/common/zookeeper/ServerSets.java
```diff
@@ -78,6 +78,6 @@ public static ServiceInstance deserializeServiceInstance(
    * @param address the target address to create the endpoint for
    */
   public static Endpoint toEndpoint(InetSocketAddress address) {
-    return new Endpoint(address.getHostName(), address.getPort());
+    return new Endpoint(address.getHostString(), address.getPort());
```
kevinoliver: seems good, especially about avoiding the reverse dns lookup.

i'll see about getting this merged.

kevinoliver:
merged in https://github.com/twitter/finagle/commit/2c17f19bf5965926e9fe94eaf790855726564ab0

thanks for the patch @GEverding

PR #690, opened 2018-05-02
finagle-stats: fix MetricsBucketedHistogram latch period

edio:
Problem

`MetricsBucketedHistogram` allows specifying custom
latchPeriod to control size of the window for snapshot
calculation. However `snapshot` would set the end of the
window to the beginning of next minute if it is the first
snapshot ever or if previous snapshot is older than
`2 * latchPeriod`

Solution

Replace beginning of the next minute in
`MetricsBucketedHistogram.snapshot` with `Time.now + latchPeriod`


edio:
I made this fix as small and as isolated as possible. I see, how this class may be improved in terms of both performance and semantics, I may do this in a different PR if there's interest.

mosesn:
Thanks for the PR, this looks great!  Let's continue the discussion on the initial ticket, sorry I've been slow in responding, I've been caught up in other stuff.

isabelmartin:
Thanks @edio! This has been merged in [0c39217724c2c701b2b7f46932ee5869a997c061](https://github.com/twitter/finagle/commit/0c39217724c2c701b2b7f46932ee5869a997c061)

PR #694, opened 2018-05-16
Avoid NoClassDefFoundError if netty-transport-native-epoll is excluded

robertpanzer:
Problem

Currently Finagle fails at startup when netty-transport-native-epoll is not available on the classpath.
Even on OS X, or if epoll is explicitly disabled. 
(If I have seen correctly it is even disabled by default here: https://github.com/twitter/finagle/blob/835b6c3f155c066b679294083fba638111f52a39/finagle-netty4/src/main/resources/com/twitter/toggles/configs/com.twitter.finagle.netty4.json#L3-L7)

Solution

The solution is to move the creation of the different EventLoopGroups into own methods with an explicit type signature returning just EventLoopGroup.

Result

As a result it should be possible to exclude netty-transport-native-epoll from the classpath and therefore reduce the size of the bundle.

Side note: 
I see that I didn't add any tests. Unfortunately my sbt know how is not existing, so I don't know how to set up separate tests with different dependencies.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -24,6 +24,9 @@ Bug Fixes
     ServicePerEndpoint. It can be set via `c.t.f.thrift.RichClientParam` or a `with`-method
     as `Thrift{Mux}.client.withPerEndpointStats`. ``PHAB_ID=D169427``
 
+  * finagle-netty4: Avoid NoClassDefFoundError if netty-transport-native-epoll is not available
+    on the classpath.
+
 18.5.0
 -------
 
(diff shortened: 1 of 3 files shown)


robertpanzer:
Only guessing, but I guess the VM will automatically load all types that appear in a method, that's why it crashes even if the toggle is disabled.
By moving the creation of the EpollEventLoopGroup into an own method, that also has only the base type EventLoopGroup in the signature, we can defer loading it until the new method is really invoked.

isabelmartin:
Thanks Robert! This has been merged in [940809af1a5f1874af0513c9cdca3b1c3b2ce03c](https://github.com/twitter/finagle/commit/940809af1a5f1874af0513c9cdca3b1c3b2ce03c#diff-fcef520180ae4933a3171883ef45d69a)

PR #703, opened 2018-07-04
Add time unit to connection_duration documentation

Axxiss:
**Problem**

Time unit is not present in the metric name neither is mentioned in the documentation.

**Solution**

Add time unit to the documentation.

**Result**

Dev won't have to check finagle's source code to know which is the unit of `connection_duration`

### DIFF
--- a/doc/src/sphinx/metrics/Transport.rst
+++ b/doc/src/sphinx/metrics/Transport.rst
@@ -44,7 +44,7 @@ ChannelStatsHandler
   histograms.
 
 **connection_duration** `verbosity:debug`
-  A histogram of the duration of the lifetime of a connection.
+  A histogram of the duration of the lifetime of a connection, in milliseconds.
 
 **connection_received_bytes** `verbosity:debug`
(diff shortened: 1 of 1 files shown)


smlance:
Thanks @Axxiss. I think I agree that it is in milliseconds. I'll have this reviewed and merged internally.

Axxiss:
@smlance For the record, here is where the duration is tracked:

https://github.com/twitter/finagle/blob/911a01cee08852f47385ac7b9d709cf3fc5422a9/finagle-netty4/src/main/scala/com/twitter/finagle/netty4/channel/ChannelStatsHandler.scala#L33-L34
https://github.com/twitter/finagle/blob/911a01cee08852f47385ac7b9d709cf3fc5422a9/finagle-netty4/src/main/scala/com/twitter/finagle/netty4/channel/ChannelStatsHandler.scala#L133

smlance:
Thanks again, @Axxiss. This has been merged in https://github.com/twitter/finagle/commit/942fb33a5454a28702c52bae46321712fea41c13.

PR #706, opened 2018-07-10
add Whisk to adopters

viktortnk:
Whisk is actively using finagle for it's APIs (Finatra) and inter-service communication (finagle-thrift)


bryce-anderson:
Merged as 60187e275dfe75dbb81ae1a8e6e4e2f86430196d. Thanks!

PR #714, opened 2018-08-16
finagle-netty4: allow sockets to be configured with SO_REUSEPORT

zackangelo:
Problem

Would like to be able to build servers that open sockets with the `SO_REUSEPORT` option on Linux. `SO_REUSEPORT` allows multiple processes to bind and accept connections from the same address and port. 

Our particular use-case involves being able to support near zero-downtime restarts with finagle servers. We'd like to be able to start a new, simultaneously-running finagle process, listen on the same interface and port and then gracefully shutdown the old PID when the new process is healthy. 

Solution

Support for `SO_REUSEPORT` is available when using the native-epoll transport in netty4 (https://github.com/netty/netty/commit/d1d8a6b6cdfbf3946fbce0a7a0bf7657cbfade03).  Allow this option to be configured using `Transport.Options`.

### DIFF
--- a/finagle-core/src/main/scala/com/twitter/finagle/transport/Transport.scala
+++ b/finagle-core/src/main/scala/com/twitter/finagle/transport/Transport.scala
@@ -217,14 +217,24 @@ object Transport {
    *
    * @param reuseAddr enables or disables `SO_REUSEADDR` option on a
    *                  transport socket. Default is `true`.
+   *
+   * @param reusePort enables or disables `SO_REUSEPORT` option on a
+   *                  transport socket (Linux 3.9+ only). This option is only
+   *                  available when using finagle-netty4 and native epoll support
+   *                  is enabled. Default is `false`.
    */
-  case class Options(noDelay: Boolean, reuseAddr: Boolean) {
+  case class Options(noDelay: Boolean, reuseAddr: Boolean, reusePort: Boolean) {
+    def this(noDelay: Boolean, reuseAddr: Boolean) = this(noDelay, reuseAddr, reusePort = false)
+
     def mk(): (Options, Stack.Param[Options]) = (this, Options.param)
   }
 
   object Options {
     implicit val param: Stack.Param[Options] =
-      Stack.Param(Options(noDelay = true, reuseAddr = true))
+      Stack.Param(Options(noDelay = true, reuseAddr = true, reusePort = false))
+
+    def apply(noDelay: Boolean, reuseAddr: Boolean): Options =
+      this.apply(noDelay = noDelay, reuseAddr = reuseAddr, reusePort = false)
   }
 
   /**
--- a/finagle-netty4/src/main/scala/com/twitter/finagle/netty4/ConnectionBuilder.scala
+++ b/finagle-netty4/src/main/scala/com/twitter/finagle/netty4/ConnectionBuilder.scala
@@ -27,6 +27,7 @@ import io.netty.channel.socket.nio.NioSocketChannel
 import java.lang.{Boolean => JBool, Integer => JInt}
 import java.net.SocketAddress
 import java.nio.channels.UnresolvedAddressException
+
 import scala.util.control.NonFatal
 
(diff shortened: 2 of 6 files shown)


finagle-core/src/main/scala/com/twitter/finagle/transport/Transport.scala
```diff
@@ -217,14 +217,17 @@ object Transport {
    *
    * @param reuseAddr enables or disables `SO_REUSEADDR` option on a
    *                  transport socket. Default is `true`.
+   *
+   * @param reusePort enables or disables `SO_REUSEPORT` option on a
+    *                 transport socket (Linux 3.9+ only). Default is `false`.
```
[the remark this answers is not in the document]
zackangelo: okay will do!

zackangelo:
@dschobel ♻️ added some more detail in the comment

kevinoliver:
@zackangelo thanks. can you add a note about this being a breaking change to `Transport.Option` in CHANGES?

finagle-netty4/src/main/scala/com/twitter/finagle/netty4/ConnectionBuilder.scala
```diff
@@ -80,6 +82,7 @@ private final class ConnectionBuilder(
         .channel(channelClass)
         .option(ChannelOption.ALLOCATOR, allocator)
         .option[JBool](ChannelOption.TCP_NODELAY, noDelay)
+        .option[JBool](UnixChannelOption.SO_REUSEPORT, reusePort)
```
dschobel: does SO_REUSEPORT make sense on the client side?

the ci logs (https://travis-ci.org/twitter/finagle/jobs/416903070) show a lot of instances of

`WARNING: Unknown channel option 'io.netty.channel.unix.UnixChannelOption#SO_REUSEPORT' for channel '[id: 0xae222ddf]'
` and I'm wondering whether it's because we're using it on the client side or whether we need to guard the setting so that we don't spam people's logs

zackangelo: agree, it should be removed

zackangelo:
@kevinoliver does this introduce a compile-time breaking change? the new argument added to `Transport.Options` has a default value, so I don't think it'll break any existing code.

finagle-netty4/src/main/scala/com/twitter/finagle/netty4/ConnectionBuilder.scala
zackangelo: @dschobel do you want me to just ignore the presence of this setting in `ConnectionBuilder`?

zackangelo: Actually, along the same lines, would it be preferable if I only add the option if `nativeEpoll.enabled` is true?

dschobel: both sound like the right call to me.

kevinoliver:
> does this introduce a compile-time breaking change? the new argument added to Transport.Options has a default value, so I don't think it'll break any existing code.

unfortunately for Java users it is a break. this is why we add "telescoping" methods for this sort of API evolution.

zackangelo:
@kevinoliver would it be preferable for me to add another `apply` method on `Transport.Options` vs documenting a breaking change?

kevinoliver:
i think being backwards compatible is about as easy as adding a changelog note. so, lets add the extra method.

zackangelo:
@kevinoliver @dschobel ♻️ added alternate constructors/apply methods for `Transport.Options` and also only added the socket option if native epoll is being used

kevinoliver:
merged this in https://github.com/twitter/finagle/commit/0316ac8940f056b86e3d5463ba63b6033a539886 

thanks again @zackangelo

PR #718, opened 2018-08-27
Fixed Spelling.

jimmycasey:
Problem

Solution

Result

the solution.


mosesn:
looks good to me, thanks!

PR #723, opened 2018-09-01
Use current dtabs in NoBrokersAvailableException

### LINKED ISSUE #716: Missing dtab information in NoBrokersAvailableException
The message of the `NoBrokersAvailableException` always shows an empty `Dtab.base` and `Dtab.local` even if they are set.

### Expected behavior

The message of the `NoBrokersAvailableException` should display the actual `Dtab.base` and `Dtab.local` being used.

### Actual behavior

Exception message always shows both the Dtab.base and Dtab.local as empty. This causes confusion specifically in cases where the Dtab is set in other parts of the code. Then it's unclear whether the dtab was actually picked up.

Run the example code below:
```scala
object FinagleTest extends App {
  Dtab.base = Dtab.read("/s => /$/inet/localhost/8000")
  Dtab.local = Dtab.read("/s => /$/inet/non-existing-host/8000")

  val service = client.newService("/s/aweseomservice")

  val response = Await.result(service.apply(http.Request(Method.Get, "/hello")))
  println(response.getContentString())
}
```

It results in this exception: `Exception in thread "main" com.twitter.finagle.NoBrokersAvailableException: No hosts are available for /s/aweseomservice, Dtab.base=[], Dtab.local=[]. Remote Info: Not Available`

While I was expecting the Dtab.base and Dtab.local parts to be filled with the actual dtab information.

coduinix:
Problem
In some cases the Dtab information shows as empty in the NoBrokersAvailableException, even if there was a Dtab (base and/or local) set. This might be confusing when debugging resolver issues.

Solution

If no Dtab was given to the NoBrokersAvailableException, look up the Dtab when constructing the exceptionMessage.

Result

Actual effective Dtab is shown in NoBrokersAvaialbleException, fixes #716

### DIFF
--- a/finagle-core/src/main/scala/com/twitter/finagle/Exceptions.scala
+++ b/finagle-core/src/main/scala/com/twitter/finagle/Exceptions.scala
@@ -134,10 +134,18 @@ class GlobalRequestTimeoutException(timeout: Duration)
  *
  * [1] https://twitter.github.io/finagle/guide/Names.html
  */
-class NoBrokersAvailableException(val name: String, val baseDtab: Dtab, val localDtab: Dtab)
+class NoBrokersAvailableException(val name: String, val baseDtabFn: () => Dtab, val localDtabFn: () => Dtab)
     extends RequestException
     with SourcedException {
-  def this(name: String = "unknown") = this(name, Dtab.empty, Dtab.empty)
+
+  // backwards compatibility constructor
+  def this(name: String, baseDtab: Dtab, localDtab: Dtab) =
+    this(name, () => baseDtab, () => localDtab)
+
+  def this(name: String = "unknown") = this(name, () => Dtab.base, () => Dtab.local)
+
+  def baseDtab: Dtab = baseDtabFn()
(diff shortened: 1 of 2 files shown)


finagle-core/src/main/scala/com/twitter/finagle/Exceptions.scala
```diff
@@ -134,10 +134,18 @@ class GlobalRequestTimeoutException(timeout: Duration)
  *
  * [1] https://twitter.github.io/finagle/guide/Names.html
  */
-class NoBrokersAvailableException(val name: String, val baseDtab: Dtab, val localDtab: Dtab)
+class NoBrokersAvailableException(val name: String, val baseDtabFn: () => Dtab, val localDtabFn: () => Dtab)
```
mosesn: It's usually pretty unusual and baseDtab changes this late in the game.  how about we keep `baseDtab` a field until someone needs it?

nepthar:
Merged in 1d42c39. Thanks!

PR #724, opened 2018-09-05
finagle-serversets: initialize zkHealth to Unknown

sboobna:
Problem

On start up, the zkHealth reports "Healthy" when the client is not able to connect to zookeeper. On debugging, I noticed that the ClientHealth was being initialized to Healthy.

Solution

The ClientHealth should be initialized to Unknown, and it will become Healthy only when it connects to zookeeper (which internally means the session changing to SyncConnected)

Result

If the client isn't able to connect to zookeeper on startup, it will report its health as Unknown.


sboobna:
@kevinoliver

sboobna:
@kevinoliver the failures look unrelated to me

sboobna:
@kevinoliver thanks for the approval. What else do I need to do to get this merged in, since there are some unrelated test failures?

kevinoliver:
merged in 8e3b1cb4893cce389fd61243215ae382b8a14e05

thank you @sboobna

PR #737, opened 2018-10-17
change ResponseConformanceFilter log level from info to debug

dadjeibaah:
Problem

The ResponseConformanceFilter is responsible for ensuring that requests and responses messages that flow through a Finagle service are HTTP compliant. There are cases where the filter enforces certain request/response rules and every time that happens the Filter logs information at the info level. There may be certain situations where a server sends HTTP non-compliant responses by design i.e. sending 204 with bodies. A server that receives requests at a high QPS that result in these kinds of responses end up logging a lot of messages like `"Response with a status code of 204 must not have a body-message but it has a 10-byte payload, thus the content has been removed."` at an `INFO` level.

Solution

Modify all logging statements to `logger.debug` so that these messages are only visible at the `DEBUG` level since it seems more appropriate for these messages to be set at that level.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,6 +6,8 @@ Note that ``PHAB_ID=#`` and ``RB_ID=#`` correspond to associated messages in com
 
 Unreleased
 ----------
+* finagle-base-http: logging levels in `c.t.f.h.c.ResponseConformanceFilter` have
+  been change from `info` to `debug`
 
 Deprecations
 ~~~~~~~~~~~~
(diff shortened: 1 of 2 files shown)


bryce-anderson:
I'm not sure if I agree with this change in log level since the library is modifying the users input on the their behalf. If you're explicitly performing non-compliant behavior maybe the right solution is to change the log level of that logger alone in your application?

I'd also be interested in more opinions.

bryce-anderson:
@dadjeibaah, I had speculated they were related. 😄 We'd love a PR for 739. Thanks for your help!

PR #742, opened 2018-10-30
Handle empty address sets in TrafficDistributor

dadjeibaah:
Signed-off-by: Dennis Adjei-Baah <dennis@buoyant.io>

Problem

Finagle's `TrafficDistributor` is used to build endpoints with their associated weight classes. This endpoint-to-weight class grouping is used to create load balancers used to balance requests for a particular endpoint stack. A `TrafficDistributor`'s load balancer are tracked in the global `BalancerRegistry` every time there is an address set change for a particular service. The issue is that `TrafficDistributor` has two places in its code that creates a load balancer and one of these balancers never gets destroyed when a service is closed. This causes a leak in `BalancerRegistry` This scenario is evident when a service's address set alternates periodically between an empty address set to a valid populated address set.

Solution

This PR removes code that extraneously creates an empty load balancer and adds it to the `partition` method in `TrafficDistributor`. This causes `TrafficDistributor` to manage address state in one area and avoids the object leak.

### DIFF
--- a/finagle-core/src/main/scala/com/twitter/finagle/loadbalancer/TrafficDistributor.scala
+++ b/finagle-core/src/main/scala/com/twitter/finagle/loadbalancer/TrafficDistributor.scala
@@ -283,9 +283,6 @@ private class TrafficDistributor[Req, Rep](
   // ServiceFactories that can service requests.
   private[this] val underlying: Event[ServiceFactory[Req, Rep]] =
     weightClasses.foldLeft(init) {
-      case (_, Activity.Ok(List(WeightClass(lb, _, _,sz)))) if sz == 0 =>
-        pending.updateIfEmpty(Return(lb))
-        lb
       case (_, Activity.Ok(wcs)) =>
         val dist = new Distributor(wcs, rng)
         updateGauges(wcs)
--- a/finagle-core/src/main/scala/com/twitter/finagle/loadbalancer/TrafficDistributor.scala
+++ b/finagle-core/src/main/scala/com/twitter/finagle/loadbalancer/TrafficDistributor.scala
@@ -213,22 +213,22 @@ private class TrafficDistributor[Req, Rep](
           val lb = newBalancer(Activity(emptyEndpoints))
           balancers + (0.0 -> CachedBalancer(lb, emptyEndpoints, 0))
         } else {
-        weightedGroups.foldLeft(balancers) {
-          case (cache, (weight, factories)) =>
-            val unweighted = factories.map { case WeightedFactory(f, _) => f }
-            val newCacheEntry = if (cache.contains(weight)) {
-              // an update that contains an existing weight class updates
-              // the balancers backing collection.
-              val cached = cache(weight)
-              cached.endpoints.update(Activity.Ok(unweighted))
-              cached.copy(size = unweighted.size)
-            } else {
-              val endpoints: BalancerEndpoints[Req, Rep] = Var(Activity.Ok(unweighted))
-              val lb = newBalancer(Activity(endpoints))
-              CachedBalancer(lb, endpoints, unweighted.size)
-            }
-            cache + (weight -> newCacheEntry)
-        }
+          weightedGroups.foldLeft(balancers) {
+            case (cache, (weight, factories)) =>
+              val unweighted = factories.map { case WeightedFactory(f, _) => f }
+              val newCacheEntry = if (cache.contains(weight)) {
+                // an update that contains an existing weight class updates
+                // the balancers backing collection.
+                val cached = cache(weight)
+                cached.endpoints.update(Activity.Ok(unweighted))
+                cached.copy(size = unweighted.size)
+              } else {
+                val endpoints: BalancerEndpoints[Req, Rep] = Var(Activity.Ok(unweighted))
+                val lb = newBalancer(Activity(endpoints))
+                CachedBalancer(lb, endpoints, unweighted.size)
+              }
+              cache + (weight -> newCacheEntry)
+          }
         }
 
         // weight classes that no longer exist in the update are removed from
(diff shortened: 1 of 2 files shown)


finagle-core/src/test/scala/com/twitter/finagle/loadbalancer/TrafficDistributorTest.scala
```diff
@@ -435,6 +435,7 @@ class TrafficDistributorTest extends FunSuite {
       assert(sr.counters(Seq("test", "loadbalancer", "removes")) == i - 1)
       assert(sr.gauges(Seq("test", "loadbalancer", "meanweight"))() == i)
       assert(sr.numGauges(Seq("test", "loadbalancer", "meanweight")) == 1)
+      println(sr.counters(Seq("test", "closes")))
```
bryce-anderson: Stray?

finagle-core/src/main/scala/com/twitter/finagle/loadbalancer/TrafficDistributor.scala
```diff
@@ -272,12 +277,9 @@ private class TrafficDistributor[Req, Rep](
   // ServiceFactories that can service requests.
   private[this] val underlying: Event[ServiceFactory[Req, Rep]] =
     weightClasses.foldLeft(init) {
-      case (_, Activity.Ok(wcs)) if wcs.isEmpty =>
-        // Defer the handling of an empty destination set to `newBalancer`
-        val emptyBal = newBalancer(Activity(Var(Activity.Ok(Set.empty[EndpointFactory[Req, Rep]]))))
-        updateGauges(wcs)
-        pending.updateIfEmpty(Return(emptyBal))
-        emptyBal
+      case (_, Activity.Ok(List(WeightClass(lb, _, _,sz)))) if sz == 0 =>
```
roanta: do we still need this case? I think the `Distributor` should just do the right thing.

finagle-core/src/test/scala/com/twitter/finagle/loadbalancer/TrafficDistributorTest.scala
```diff
@@ -470,4 +471,20 @@ class TrafficDistributorTest extends FunSuite {
...
+    val dest = Var(init)
+    val dist = newDist(dest)
+
+    dest() = Activity.Ok(Set((1, 2.0),(2, 1.0),(3, 2.0)).map(x => WeightedAddress(Address(x._1), x._2)))
+    val bal0 = Await.result(dist())
+    assert(balancers.size == 2)
+
+    dest() = Activity.Ok(Set.empty[Address])
+    val bal1 = intercept[Exception] {Await.result(dist())}
+    assert(balancers.size == 3)
```
roanta: Are we sure this is getting rid of (or calling `close` on) the empty balancer? Can we validate this by adding a hook to the `Balancer` close method in `Ctx`?

finagle-core/src/test/scala/com/twitter/finagle/loadbalancer/TrafficDistributorTest.scala
dadjeibaah: thanks for catching that!

finagle-core/src/test/scala/com/twitter/finagle/loadbalancer/TrafficDistributorTest.scala
dadjeibaah: My manual tests of this branch, or at least looking at the `Balancers` in `loadbalancers.json`, indicated that the `Balancers` were being closed. but I can add your suggestion to verify.

finagle-core/src/main/scala/com/twitter/finagle/loadbalancer/TrafficDistributor.scala
```diff
@@ -208,8 +208,12 @@ private class TrafficDistributor[Req, Rep](
...
         val weightedGroups: Map[Double, Set[WeightedFactory[Req, Rep]]] =
           activeSet.groupBy(_.weight)
-
-        val merged = weightedGroups.foldLeft(balancers) {
+        val merged = if (weightedGroups.isEmpty) {
+          val emptyEndpoints: BalancerEndpoints[Req, Rep] = Var(Activity.Ok(Set.empty[EndpointFactory[Req, Rep]]))
+          val lb = newBalancer(Activity(emptyEndpoints))
+          balancers + (0.0 -> CachedBalancer(lb, emptyEndpoints, 0))
+        } else {
+        weightedGroups.foldLeft(balancers) {
```
kevinoliver: nitpick: the indentation on the else block looks like it needs +2

finagle-core/src/test/scala/com/twitter/finagle/loadbalancer/TrafficDistributorTest.scala
```diff
@@ -470,4 +476,41 @@ class TrafficDistributorTest extends FunSuite {
...
+    val init: Activity.State[Set[Address]] = Activity.Pending
+    val dest = Var(init)
+    val dist = newDist(dest)
+
+    dest() = Activity.Ok(Set((1, 2.0),(2, 1.0),(3, 2.0)).map(x => WeightedAddress(Address(x._1), x._2)))
+    val bal0 = Await.result(dist())
+    assert(balancers.size == 2)
+
+    dest() = Activity.Ok(Set.empty[Address])
+    val ex = intercept[Exception] {Await.result(dist())}
```
kevinoliver: is this a no brokers available exception? can we test for that instead of any Exception? (same in the test below)

dadjeibaah: Yes it is, thanks for pointing that out. I will make the suggested change.

vkostyukov:
Merged in 21664df3700257de3165511c2625d232c399322b.

PR #749, opened 2019-01-02
B3 header support

crispywalrus:
Problem

OpenZipkin has created a new set of header to carry trace information. These headers are not compatible with the old ones and are needed to interoperate with services that use the new version of the header.

Solution

If a "b3" header exists in the request this extracts all the new information and populates finagles tracing header with the existing values.

### DIFF
--- a/finagle-base-http/src/main/scala/com/twitter/finagle/http/TraceInfo.scala
+++ b/finagle-base-http/src/main/scala/com/twitter/finagle/http/TraceInfo.scala
@@ -12,6 +12,47 @@ private object TraceInfo {
   }
 
   def letTraceIdFromRequestHeaders[R](request: Request)(f: => R): R = {
+    // rather than rewrite all this to handle more I'm just shimming
+    // in a check for the new trace b3 header and if it exists I'm
+    // extracting out its components to be handled by the existing
+    // code.
+    if (request.headerMap.contains(Header.TraceContext)) {
+      def handleOne(headers: HeaderMap,value: String): Unit =
+        value match {
+          case "d" => headers.set(Header.Flags,"1")
+          case "1" => headers.set(Header.Sampled,"1")
+          case _ => ()
+        }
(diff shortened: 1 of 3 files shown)


crispywalrus:
@adriancole @mosesn this is the result of the last changes to https://github.com/crispywalrus/finagle/pull/1 to add OpenZipkin support to finagle.

finagle-base-http/src/main/scala/com/twitter/finagle/http/TraceInfo.scala
```diff
@@ -11,7 +11,53 @@ private object TraceInfo {
     while (iter.hasNext) headers -= iter.next()
   }
 
+  def convertB3Trace(request: Request) =
+    if (request.headerMap.contains(Header.TraceContext)) {
+      def handleSampled(headers: HeaderMap,value: String): Unit =
+        value match {
+          case "0" => headers.set(Header.Flags,"0")
+          case "d" => headers.set(Header.Flags,"1")
```
adriancole: don't recall if in scala case falls through, but debug implies sampled. so you could also set the sampled header here

kevinoliver:
This approach seems reasonable. I'd like to see some unit tests before going forward.

luciferous:
Hi @crispywalrus, how is testing going? Is there anything we can do to assist?

crispywalrus:
@kevinoliver this adds some basic test that show a b3 header ends up being parsed into finagles X-B3 headers. I've chosen to remove the `b3` header as internal code produces `X-B3` headers and it cuase problems to have both versions in my test rig. Sorry this took so long to finish off.

adriancole:
hey folks. can you help get this merged? @crispywalrus has responded and we are nearing a year on the original feature request.

crispywalrus:
@enbnt why did this get closed?

enbnt:
@crispywalrus because it has landed in 8721837 😄

crispywalrus:
Great!

luciferous:
Thanks for the patience, everyone!

PR #761, opened 2019-03-28
finagle-mysql: Make DateValue.unapply charset agnostic

CrimsonVoid:
Problem

Some MySQL-speaking databases send DateValue's with UTF-8 charsets, which results in failing pattern matches. It doesn't seem like there needs to be a restriction on what charset `DateValue.unapply` accepts when `Value.isBinary` is unset.

Solution

Use v.charset for String creation


vkostyukov:
Thanks @CrimsonVoid! I will pull this internally.

vkostyukov:
Merged in d3d2f8cbb50785bf940427d539a30befc011f0d5. Thanks again @CrimsonVoid!

PR #766, opened 2019-04-30
finagle-core: Fix rate limiter if negative rate or after window

glegoux:
Problem

When rate is negative then request could be executed, and
after one elapsed time windows rate is not well modified.

Solution

Refuse request when rate is negative and reset correctly limit
after one elapsed time window.

### DIFF
--- a/finagle-core/src/main/scala/com/twitter/finagle/service/RateLimitingFilter.scala
+++ b/finagle-core/src/main/scala/com/twitter/finagle/service/RateLimitingFilter.scala
@@ -11,6 +11,8 @@ import com.twitter.finagle.{RefusedByRateLimiter, Service, SimpleFilter}
 class LocalRateLimitingStrategy[Req](categorizer: Req => String, windowSize: Duration, rate: Int)
     extends (Req => Future[Boolean]) {
 
+  require(rate > 0, s"rate $rate must be strictly positive.")
+
   private[this] val rates = mutable.HashMap.empty[String, (Int, Time)]
 
   def apply(req: Req): Future[Boolean] = synchronized {
@@ -18,9 +20,7 @@ class LocalRateLimitingStrategy[Req](categorizer: Req => String, windowSize: Dur
     val id = categorizer(req)
     val (remainingRequests, timestamp) = rates.getOrElse(id, (rate, now))
 
-    val accept = if (rate <= 0) {
-      false
-    } else if (timestamp.until(now) > windowSize) {
+    val accept = if (timestamp.until(now) > windowSize) {
       rates(id) = (rate - 1, now)
       true
     } else {
--- a/finagle-core/src/main/scala/com/twitter/finagle/service/RateLimitingFilter.scala
+++ b/finagle-core/src/main/scala/com/twitter/finagle/service/RateLimitingFilter.scala
@@ -21,7 +21,7 @@ class LocalRateLimitingStrategy[Req](categorizer: Req => String, windowSize: Dur
     val (remainingRequests, timestamp) = rates.getOrElse(id, (rate, now))
 
     val accept = if (timestamp.until(now) > windowSize) {
-      rates(id) = (rate - 1, now)
+      rates(id) = (rate, now)
       true
     } else {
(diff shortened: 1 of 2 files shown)


finagle-core/src/main/scala/com/twitter/finagle/service/RateLimitingFilter.scala
```diff
@@ -13,20 +13,23 @@ class LocalRateLimitingStrategy[Req](categorizer: Req => String, windowSize: Dur
...
 
-  def apply(req: Req) = synchronized {
+  def apply(req: Req): Future[Boolean] = synchronized {
     val now = Time.now
     val id = categorizer(req)
     val (remainingRequests, timestamp) = rates.getOrElse(id, (rate, now))
 
-    val accept = if (timestamp.until(now) > windowSize) {
-      rates(id) = (rate, now)
+    val accept = if (rate <= 0) {
```
vkostyukov: Negative (or zero) `rate` is a misconfiguration in my books. Maybe we can just fail fast early, at the `LocalRateLimitingStrategy ` construction time?

glegoux: Ok, I added patch so that `LocalRateLimitingStrategy` fails at construction time.

finagle-core/src/main/scala/com/twitter/finagle/service/RateLimitingFilter.scala
```diff
@@ -11,22 +11,25 @@ import com.twitter.finagle.{RefusedByRateLimiter, Service, SimpleFilter}
...
 
-  def apply(req: Req) = synchronized {
+  def apply(req: Req): Future[Boolean] = synchronized {
     val now = Time.now
     val id = categorizer(req)
     val (remainingRequests, timestamp) = rates.getOrElse(id, (rate, now))
 
     val accept = if (timestamp.until(now) > windowSize) {
-      rates(id) = (rate, now)
+      rates(id) = (rate - 1, now)
```
vkostyukov: Why did this change?

finagle-core/src/main/scala/com/twitter/finagle/service/RateLimitingFilter.scala
vkostyukov: 👍

finagle-core/src/main/scala/com/twitter/finagle/service/RateLimitingFilter.scala
glegoux: I added a test for that, else counter is not well reset after one elapsed window time.

glegoux: I removed:
```
if (rate <= 0) {
      false
```
Because now rate is necessarily strictly positive at construction time (with require condition rate <= 0 is impossible). 
`remainingRequests` resolved case of no more remaining request.

glegoux: I added a test for that: 
```
 rates(id) = (rate - 1, now)
```
Else counter is not well reset after a elapsed time window

glegoux: Do you agree with these changes ?

vkostyukov: Sorry missed notifications about this.

No, I do not agree. This is clearly a behavior change and I'm not sure why to have to do that. Is in optimization? Or code clean up?

glegoux: It was about debut your rate limiter. As you want, I removed that and test in relation to this change. I let only test if rate limiter is negative.

vkostyukov:
I'm pulling this internally.

vkostyukov:
Merged in 03d1ad8bc6de13676d3164484c4448b8936007b5.

PR #763, opened 2019-04-08
finagle-redis: Add RPOPLPush Command

martinffx:
# Problem

In order to implement a reliable queueing mechanism using redis
you need this command.

# Solution

Added command to `ListCommands`

PR #767, opened 2019-05-18
Add rpoplpush command redis

martinffx:
### Problem

In order to implement a reliable queueing mechanism using redis
you need this command.

### Solution

Added command to ListCommands

Replaces: https://github.com/twitter/finagle/pull/763


martinffx:
@yufangong @kevinoliver I took a while to get back to this but here is an updated PR, where I've manually tested and made some changes.

finagle-redis/src/main/scala/com/twitter/finagle/redis/ListCommands.scala
```diff
@@ -148,4 +148,17 @@ private[redis] trait ListCommands { self: BaseClient =>
...
+   * Atomically returns and removes the last element (tail) of the list stored at source,
+   * and pushes the element at the first element (head) of the list stored at destination
+   *
+   * @return `Some` of the value of the popped element, or `None` if the list is
+   *         empty.
+   */
+  def rPopLPush(source: Buf, dest: Buf): Future[Option[Buf]] =
+    doRequest(RPopLPush(source, dest)) {
+      case BulkReply(message) => Future.value(Some(message))
+      case EmptyBulkReply => Future.value(None)
```
luciferous: Let's use `Future.None`

luciferous:
Merged in b78da58ecbb2e9d749de513bf5fa38073d3b49bc.

PR #773, opened 2019-06-14
Allow to use SOCKS proxy for localhost

gpoulin:
Problem

SOCKS proxy is bypassed when connecting to localhost and there's no way to override this behaviour. Using a SOCKS proxy when connecting to localhost can be useful to easily capture communication between local services for debugging.

Solution

Add a flags to enable SOCKS proxy for localhost.

Result

You can now force the use of SOCKS proxy for localhost with `-com.twitter.finagle.socks.socksProxyForLocalhost`. Without this flag, the behaviour of bypassing the SOCKS proxy is kept.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -55,6 +55,9 @@ New Features
   Use `withExecutionOffloaded` configuration method (on a client or a server) to access
   new functionality. ``PHAB_ID=D325049``
 
+* finagle-core: Allow to not bypass SOCKS proxy for localhost by usingg the GlobalFlag
+  `-com.twitter.finagle.socks.socksProxyForLocalhost`
+
 * finagle-http: Added counters for request/response stream as: `stream/request/closed`,
   `stream/request/failures`, `stream/request/failures/<exception_name>`, `stream/request/opened`,
   `stream/request/pending` and `stream/response/closed`, `stream/response/failures`,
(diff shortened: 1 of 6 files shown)


CHANGELOG.rst
```diff
@@ -32,6 +32,9 @@ New Features
   adding and removing cookies in bulk, without triggering a header rewrite on each item.
   ``PHAB_ID=D318013``
 
+* finagle-core: Allow to not bypass SOCKS proxy for localhost by usingg the GlobalFlag
+  `-com.twitter.finagle.sockssocksProxyForLocalhost`
```
kevinoliver: typo here with an extra socks. (though in real life, you do want pairs of socks). same typo in the commit message.

also, mind sorting this entry up with the other finagle-core changes (line 22ish)

finagle-core/src/main/scala/com/twitter/finagle/socks/flags.scala
```diff
@@ -8,6 +8,9 @@ private[finagle] object socksProxyHost extends GlobalFlag("", "SOCKS proxy host"
 private[finagle] object socksProxyPort extends GlobalFlag(0, "SOCKS proxy port") {
   override val name = "socksProxyPort"
 }
+private[finagle] object socksProxyForLocalhost extends GlobalFlag(false, "If true SOCKS proxy is also use for localhost") {
```
kevinoliver: its nice when we state the default. how about something like,
"Bypass the SOCKS proxy for localhost connections, defaults to true."?

kevinoliver:
thanks for the PR. looks really close to me.

mosesn:
@gpoulin have you had a chance to address Kevin's comments yet?

gpoulin:
@mosesn Sorry about that, I forgot it. I should have address the comments.

mosesn:
thanks @gpoulin!  I'll take a stab at merging this in, sorry for the delay!

mosesn:
@gpoulin this was successfully merged in here: https://github.com/twitter/finagle/commit/5521bc830f792862bb2aa06847f3f8493c673412

thanks for the contribution!

PR #774, opened 2019-06-15
Update PeakEwma.scala

crispywalrus:
update link to Eckner paper.

Problem

Solution

Result

the solution.


crispywalrus:
addressed in #774

yufangong:
Thanks @crispywalrus, this is merged at https://github.com/twitter/finagle/commit/96cb072f1678b45811e3e5d3800b44cc3ea4e441.

PR #775, opened 2019-06-17
Change Sigma to Merit in ADOPTERS.md

omerzach:
Problem

Sigma changed its name to "Merit" (and still uses Finagle)
https://www.merits.com/blog/sigma-rebrands-as-merit

Solution

Replace ADOPTERS.md entry for Sigma with entry for Merit


yufangong:
Thanks for the update @omerzach, this is merged at https://github.com/twitter/finagle/commit/9dcfa22f5a3fdea2956740b9c360c00aee37df42.

PR #779, opened 2019-07-15
[Testing] add jdk11 to travis config

dotordogh:
Problem

We should catch compatibility issues with JDK11 sooner rather than later.

Solution

Turn on testing with JDK11 in Travis CI.

Result

Builds will likely fail, but it is helpful to know what problems we might
encounter migrating to JDK11.


dotordogh:
@ryanoneill you bet! Doing that now!

PR #781, opened 2019-07-25
Remove PayloadSizeHandler from Streaming.Enabled case

### LINKED ISSUE #780: MaxRequestSize/MaxResponseSize and fixedLengthStreamedAfter unexpected behavior
`http.param.MaxRequestSize` and `http.param.MaxResponseSize` have ambiguous semantics wheh combined with `fixedLengthStreamedAfter`

### Expected behavior
I would expect that if `fixedLengthStreamedAfter` is set say 5 kilobytes and `MaxRequestSize` is set at 4 kilobytes, a message with `Content-Length` set to 6 kilobytes is processed as a streaming message without a problem

### Actual behavior
What happens instead is that the message is dropped as it exceeds `MaxRequestSize`. To me (and I might be totally off here), this does not sounds correct. What I expect to happen is for fixed size message to always be streamed after a certain amound of length, despite the fact that there is limit on the `MaxRequestSize`. How fo these two settings relate in more detail? Isn't the purpose of `fixedLengthStreamedAfter` to enable the processing of arbitrary long messages without imposing a hard limit. Moreover, it seems at the moment `http.param.MaxRequestSize` and `http.param.MaxResponseSize` are capped at 2GB. This effectivelly means that there is no way to process a message that has `Content-Length`  set and is larger than 2GB. Is it possible that the cap from these settings is removed. And if so, is that a good idea at all?  Thanks for the clarification in advance.

Also, another question I have is how do the `MaxHeaderSize` and `MaxInitialLineSize` params relate to `fixedLengthStreamedAfter`? Are these necessary when the message is streamed? And if so, do they become reduntant when the message is larger than `fixedLengthStreamedAfter`?

zaharidichev:
Problem

At the moment messages that are larger than 2GB cannot be processed due to the fact that the MaxRequest/Response size is capped at 2GB. This renders to be a problem especially in the situation when we have `fixedLengthStreamedAfter`. This is because even though a request for example shall be streamed if its above a certain size, if it is larger than `MaxRequestSize` it will be rejected. 

Solution
Remove the PayloadSizeHandler when Streaming is Enabled in order to allow for arbitrary long  messages to be processed in a streaming fashion. 

Fixes #780


zaharidichev:
I see, was not aware of this particular restriction. Alright then we should come up with a better solution. So if `fixedLengthStreamedAfter` is set to `x` we know for a fact that any message that is larger than `x` will be streamed, right ? Lets then do this. Put a hard cap of 2GB on `fixedLengthStreamedAfter` and if a message is being processed that WILL be streamed (falls above the `fixedLengthStreamedAfter`), we do not limit it by `MaxRequest/RespSize`. Does that sounds better?

bryce-anderson:
@zaharidichev, I think we're on the same page. The problem seems to be here:
https://github.com/twitter/finagle/blob/develop/finagle-netty4-http/src/main/scala/com/twitter/finagle/netty4/http/package.scala#L173-L176
We always add the payload size handler when I think what we really want is to only add that in the non-streaming case, and we don't because in that case it's handled by the message aggregator. I'm not sure why we have it there but I'm sure there is something in the git history.

luciferous:
@bryce-anderson It was added in 485608165d58592b5dc9993dc9764d17c4489fc7. Reading through the RB and the linked ticked, I think it was to address the previous behavior where even if the server was streaming, a fixed-length HTTP message would not get forced into a stream. However, this was changed with the addition of `fixedLengthStreamedAfter`... I think.

@zaharidichev Would you mind taking on the task of removing the PayloadSizeHandler from the Streaming.Enabled case? Also, @bryce-anderson do you know of a good way to test the case that 485608165d58592b5dc9993dc9764d17c4489fc7 was protecting against?

bryce-anderson:
Ugg... From the patch you linked, it looks like behavior was very explicit
in that it wanted to reject messages that were too large, even if
streaming. While we can change the contract I'd be quite hesitant to do so.
I think we're grasping at straws, but maybe we can limit the max [req|resp]
size to Int.MaxValue in the non-streaming case? The question is how to fail
if it's misconfigured: early or just log a warning and take the smaller
number?

luciferous:
@bryce-anderson I believe "streaming" back then didn't "dechunk" fixed length messages, which is why it needed to have a max size. Today, we turn the fixed length into chunks, as if it came in with `Transport-Encoding: chunked`.

bryce-anderson:
I don't know how to test for the 'to large but we wanted streaming' case in existing servers but I agree that the intention was probably not to limit streamed request sizes in the server, just as we don't limit streaming response sizes in the client.

luciferous:
@bryce-anderson Cool.
@zaharidichev Let me know if you want advice removing the PayloadSizeHandler from the Streaming.Enabled case.

zaharidichev:
@luciferous, will take a look at it shortly. Thanks a lot!

zaharidichev:
Sorry for taking a bit longer on this one. had a pretty eventful week at home. Will get to look at it shortly.

luciferous:
@zaharidichev Thanks for the update, and for picking this up!

zaharidichev:
@luciferous Would that last change suffice?

yufangong:
Hi @zaharidichev, Can you update the description of this change?

zaharidichev:
@yufangong  will update it shortly to reflect the essence of things

luciferous:
Closed by d314bd6293fa7b42a8a64107eda8cfb10d37b9cf.

PR #782, opened 2019-07-29
finagle: allow testing against Netty SNAPSHOT in build.sbt

enbnt:
Problem

We have a manual process of validating Netty versions post-release. This
results in discovering breakages late in the cycle and can delay moving
to the next release.

Solution

Modify the build.sbt file to allow for providing properties to change
the Netty build version. If the build is a SNAPSHOT build, the snapshot
repo will be added to the resolvers.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,8 @@ git:
   quiet: true
 
 env:
+  - USE_NETTY_SNAPSHOT=false
+  - USE_NETTY_SNAPSHOT=true
   global:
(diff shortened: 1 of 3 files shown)


enbnt:
Hey, this is just me testing out some changes to allow for Travis CI to test against the Netty nightly SNAPSHOT builds, along with the pinned release version we are on. I'm not having much luck verifying the Travis CI changes locally, so am hoping that a build will trigger on this pull request. Consider this a work in progress.

enbnt:
This change has been merged

PR #788, opened 2019-08-14
WIP: Finagle Core -> 2.13

martijnhoekstra:
Problem

There is no 2.13 release for finagle core

Solution

Make a 2.13 build

Result

A 2.13 build


finagle-core/src/main/scala/com/twitter/finagle/Dtab.scala
```diff
@@ -112,9 +103,9 @@ case class Dtab(dentries0: IndexedSeq[Dentry]) extends IndexedSeq[Dentry] {
    */
   def simplified: Dtab =
     Dtab({
-      val simple = this map {
+      val simple = self.map { entry => entry match {
```
martijnhoekstra: Todo: figure out why this is needed

finagle-core/src/main/scala/com/twitter/finagle/context/MarshalledContext.scala
```diff
@@ -130,9 +130,9 @@ final class MarshalledContext private[context] extends Context {
 
   // Exposed for testing
   private[context] def marshal(env: Map[Buf, Cell]): Iterable[(Buf, Buf)] = {
-    env.mapValues {
-      case Real(k, v) => k.marshal(v.x)
-      case Translucent(_, vBuf) => vBuf
+    env.transform {
```
martijnhoekstra: todo: What?

finagle-core/src/test/scala/com/twitter/finagle/loadbalancer/TrafficDistributorTest.scala
```diff
@@ -227,7 +228,7 @@ class TrafficDistributorTest extends FunSuite {
       }.head
 
       result.foreach {
-        case ((w, _, l)) => assert(math.abs(l / w - baseline) <= baseline * ε)
+        case ((w, _, l)) => assert(math.abs(l / w - baseline) <= baseline * ε, "second one")
```
martijnhoekstra: This test is failing, todo: figure out why (and fix it)

martijnhoekstra:
blocked by https://github.com/twitter/scrooge/pull/314 and (directly and transitively) https://github.com/twitter/util/issues/253

martijnhoekstra:
Todo: rebase (redo?) on top of https://github.com/twitter/finagle/pull/789

finagle-core/src/test/scala/com/twitter/finagle/loadbalancer/TrafficDistributorTest.scala
martijnhoekstra: Increasing the size for R makes the test pass. I'm not sure if there is an underlying problem here, or that it was just unlucky RNG from the given seed.

finagle-base-http/src/main/scala/com/twitter/finagle/http/ProxyCredentials.scala
```diff
@@ -2,13 +2,13 @@ package com.twitter.finagle.http
...
-import scala.collection.JavaConversions._
+import scala.collection.JavaConverters._
 
 object ProxyCredentials {
   def apply(credentials: java.util.Map[String, String]): Option[ProxyCredentials] =
-    apply(credentials.toMap)
+    apply(credentials.asScala)
 
-  def apply(credentials: Map[String, String]): Option[ProxyCredentials] = {
+  def apply(credentials: scala.collection.Map[String, String]): Option[ProxyCredentials] = {
```
mosesn: as I mentioned on the other review, let's keep this one immutable.Map

finagle-core/src/main/scala-2.13+/com/twitter/finagle/Dtab.scala
```diff
@@ -0,0 +1,33 @@
...
+ * @see The [[https://twitter.github.io/finagle/guide/Names.html#interpreting-paths-with-delegation-tables user guide]]
+ *      for further details.
+ */
+case class Dtab(dentries0: IndexedSeq[Dentry]) extends DtabBase
+
+object Dtab extends DtabCompanionBase {
+  
+}
+
+final class DtabBuilder extends ReusableBuilder[Dentry, Dtab] {
```
mosesn: can we document the distinction between the 2.12 stuff vs the 2.13 stuff?  would we be able to just use this if we didn't need to build for 2.11, for example?

finagle-core/src/main/scala/com/twitter/finagle/Dtabs.scala
```diff
@@ -0,0 +1,13 @@
+package com.twitter.finagle
+
+/**
+ * Static methods for Java callers
```
mosesn: I love that you're thinking of java users here!  But shouldn't java-users be able to use static forwarders on Dtab?  If they can't, why can't they?

finagle-core/src/main/scala/com/twitter/finagle/Dtabs.scala
```diff
@@ -0,0 +1,13 @@
+package com.twitter.finagle
+
+/**
+ * Static methods for Java callers
+ */
+object Dtabs {
+  def empty() = Dtab(Vector.empty[Dentry])
```
mosesn: type annotations please!

finagle-core/src/main/scala/com/twitter/finagle/context/MarshalledContext.scala
mosesn: this is a bummer.  should we keep it mapValues and suppress deprecations?  or add some kind of TODO to change it back after the new mapValues is rolled out?

finagle-core/src/test/scala/com/twitter/finagle/util/CachedHashCodeTest.scala
```diff
@@ -48,8 +48,12 @@ class CachedHashCodeTest extends FunSuite {
   }
 
   test("same hashCode mixed in or not") {
-    case class NotMixedIn(s: String)
-    case class MixedIn(s: String) extends CachedHashCode.ForCaseClass
+    case class NotMixedIn(s: String) {
+      override def productPrefix = "prefix"
```
mosesn: why do we have to do this?  can you add a comment?

finagle-core/src/main/scala-2.13+/com/twitter/finagle/Dtab.scala
martijnhoekstra: The 2.12 version has `CanBuildFrom`, and 2.13 doesn't. Without it you will never be able to map a `Dtab` with a `Dentry => Int` and get back a `BitSet`.

It's a bit of a niche concern I suppose. You could consider breaking downstream and just have a Dentry => Dentry overload of map, and if 2.13 was all we ever knew and this was greenfields, it would probably be a good idea.  Whether it's enough of a justification to break downstream if they are doing something I consider moderately silly is really not up to me.

finagle-core/src/main/scala/com/twitter/finagle/context/MarshalledContext.scala
martijnhoekstra: `transform` is AFAIK the fastest of the current options. I have no strong opinions on the matter, other than that unrelatedly the first case should maybe have been `case (_, Real(k, Some(v)) => k.marshal(v)`

finagle-base-http/src/main/scala/com/twitter/finagle/http/ProxyCredentials.scala
martijnhoekstra: This commit is also in the other one -- I'll rebase again once it's in, and the change should come along with that.

finagle-core/src/main/scala/com/twitter/finagle/Dtabs.scala
martijnhoekstra: The java compilation test started failing, so I just did what I had to do. I assumed that the underlying cause there are no static forwarders anymore is because we go through a common trait now, based on the flimsiest of intuition. I'll do some experiments to find the cause, and see what bytecode is generated under what conditions.

finagle-core/src/test/scala/com/twitter/finagle/util/CachedHashCodeTest.scala
martijnhoekstra: Added comment above:

    //case class productPrefix (class name) gets mixed in
    //into hash in 2.13+, so we have to override
    //it if we want to check whether to separate
    //case classes would have hashed the same if
    //the only difference had been the mixin.

finagle-core/src/main/scala/com/twitter/finagle/context/MarshalledContext.scala
mosesn: yeah, that makes sense–but presumably mapValues will be cheaper than transform when it's reimplemented, since it won't be wrapping the kv pairs in tuples, so I'd like a reminder to switch back to mapValues when it exists (hence the TODO).

I'd be down to switch it, except that I don't know if the Some unapply gets desugared to allocate an Option?  my hunch is that case class unapply is special and avoids it, but idk.  as an aside, I was looking at the API, and it looks like we can use .value instead of .get if we know it's a `Some`, which `Real` already guarantees for us.  should we use .value?

finagle-core/src/main/scala/com/twitter/finagle/Dtabs.scala
mosesn: I'd rather keep those static forwarders if we can.  might be worth duplicating more code, or else making DtabBase an abstract class.

finagle-core/src/main/scala/com/twitter/finagle/context/MarshalledContext.scala
martijnhoekstra: case class unapply is indeed special and doesn't have to allocate anything. `value` was introduced in `2.12`, `get` is needed for 2.11 compat. `get` probably gets jitted to direct field access pretty quickly, so from a performance perspective it doesn't matter much. But if you're still pining for a reason to get rid of 2.11, this may be your ticket -- but again, I'm fine with whatever.

mosesn: if case class unapply is special, let's use it.  :shipit:

finagle-core/src/main/scala/com/twitter/finagle/Dtabs.scala
martijnhoekstra: I'll toy with it a bit. I'll get them back one way or another.

martijnhoekstra: I thought I broke my local cache with that last one. I either broke scalac or sbt. Oh well.

mosesn: thanks!

mosesn:
Now that your patch for fixing up the deprecations is merged in, can you rebase against develop when you get a chance?

martijnhoekstra:
Yes, as soon as I find the time

martijnhoekstra:
CI depends on finagle-memcached, which depends on bijection-core, which is on its way at https://github.com/twitter/bijection/pull/289

Not sure how easy memcached will be once the dependencies are met, but it seems out of scope for this PR

mosesn:
Agreed.  Let's not add 2.13.x to travisci in this review, imo.

finagle-core/src/main/scala-2.13+/com/twitter/finagle/Dtab.scala
mosesn: right.  I'm wondering if we can add a comment to this file that explains why we have to make this change?  it sounds like behavior is changing incompatibly when we switch to 2.13.x anyway, so maybe the TODO is to make the breaking change in <= 2.12.x to unify the builds and prepare users for 2.13.x?  we could add that as a comment too.

finagle-core/src/test/scala/com/twitter/finagle/loadbalancer/TrafficDistributorTest.scala
```diff
@@ -68,8 +68,9 @@ private object TrafficDistributorTest {
     override def toString: String = s"Balancer($endpoints)"
   }
 
-  // Return the distribution for the the given `balancer` as a tuple
-  // of (weight, size, offer load).
+  /** Return the distribution for the the given `balancer` as a tuple
```
mosesn: can we start the sentence that begins with "Return" on the next line?  I think that's the scaladoc style we normally use.

finagle-core/src/test/scala/com/twitter/finagle/loadbalancer/TrafficDistributorTest.scala
```diff
@@ -204,7 +205,7 @@ class TrafficDistributorTest extends FunSuite {
     }
 
     locally {
-      val R = 10 * 1000
+      val R = 100 * 1000
```
mosesn: it's really bizarre that you found it didn't work.  can we try doing it with 10K again?  I'd rather not change it if we don't have to.

finagle-toggle/src/main/scala/com/twitter/finagle/toggle/JsonToggleMap.scala
```diff
@@ -126,13 +126,13 @@ object JsonToggleMap {
       }
 
       val metadata: immutable.Seq[Toggle.Metadata] =
-        toggles.map { jsonToggle =>
+        toggles.map(jsonToggle => {
```
mosesn: why did we have to switch this to use curlies inside instead of outside?

finagle-toggle/src/main/scala/com/twitter/finagle/toggle/ToggleMap.scala
```diff
@@ -351,9 +344,9 @@ object ToggleMap {
   class Immutable(metadata: immutable.Seq[Toggle.Metadata]) extends ToggleMap {
 
     private[this] val toggles: immutable.Map[String, Toggle[Int]] =
-      metadata.map { md =>
+      metadata.map(md =>
```
mosesn: ditto, why can't this use curlies anymore?

finagle-core/src/test/scala/com/twitter/finagle/loadbalancer/TrafficDistributorTest.scala
martijnhoekstra: I'll see if changing the seed does anything -- it's taken from a hash from a string, and there are some hash changes.

finagle-core/src/test/scala/com/twitter/finagle/loadbalancer/TrafficDistributorTest.scala
martijnhoekstra: Will do

finagle-toggle/src/main/scala/com/twitter/finagle/toggle/JsonToggleMap.scala
martijnhoekstra: I could have sworn I had some deprecation warnings for this,  but I can't reproduce that anymore, so it can be changed back.

finagle-core/src/main/scala-2.13+/com/twitter/finagle/Dtab.scala
martijnhoekstra: I'll try to make this a bit neater, and make some compilation and run tests to demonstrate what behaviour is unified (and document what is and isn't). For an integrated collection in 2.13 the companion should extend `Factory` (among other reasons?) so that `Iterable#to(Dtab)` works, and that should also work on 2.12 with the scala.collection.compat library.

mosesn:
Sounds like a plan, thank you!

martijnhoekstra:
I am a bit bogged down, I'll have some more time tomorrow, but don't hold off for that if you want to push it through.

mosesn:
Weird, I ran into this error:

```
[error] /Users/mnakamura/workspace/sources/0/finagle/finagle-core/src/test/java/com/twitter/finagle/DtabCompilationTest.java:11:1: non-static method empty() cannot be referenced from a static context
[error]     Dtab d = Dtab.empty();
[error] /Users/mnakamura/workspace/sources/0/finagle/finagle-core/src/test/java/com/twitter/finagle/DtabCompilationTest.java:11:1: incompatible types: scala.collection.IterableOps cannot be converted to com.twitter.finagle.Dtab
[error]     Dtab d = Dtab.empty();
[error] /Users/mnakamura/workspace/sources/0/finagle/finagle-core/src/test/java/com/twitter/finagle/DtabCompilationTest.java:16:1: non-static method empty() cannot be referenced from a static context
[error]     Dtab.setBase(Dtab.empty());
[error] /Users/mnakamura/workspace/sources/0/finagle/finagle-core/src/test/java/com/twitter/finagle/DtabCompilationTest.java:16:1: incompatible types: scala.collection.IterableOps cannot be converted to com.twitter.finagle.Dtab
[error]     Dtab.setBase(Dtab.empty());
[error] /Users/mnakamura/workspace/sources/0/finagle/finagle-core/src/test/java/com/twitter/finagle/StackParamCompilationTest.java:80:1: non-static method empty() cannot be referenced from a static context
[error]               public Dtab apply() { return Dtab.empty(); }
[error] /Users/mnakamura/workspace/sources/0/finagle/finagle-core/src/test/java/com/twitter/finagle/StackParamCompilationTest.java:80:1: incompatible types: scala.collection.IterableOps cannot be converted to com.twitter.finagle.Dtab
[error]               public Dtab apply() { return Dtab.empty(); }
[info] Some messages have been simplified; recompile with -Xdiags:verbose to get full output
[error] (Test / compileIncremental) javac returned non-zero exit code
[error] Total time: 23 s, completed Aug 29, 2019 3:14:02 PM
```

martijnhoekstra:
That's the same problem as the one I initially fixed by making the DTabs
object, and thought was also fixed with the abstract base classes instead
of the base traits.

mosesn:
> That's the same problem as the one I initially fixed by making the DTabs object, and thought was also fixed with the abstract base classes instead of the base traits.
Hmm, doesn't seem to be fixed on my machine.  I think this might be a bug in scalac–when I javap, I see two methods named empty, and neither of them is static.

```
  public scala.collection.IterableOps empty();
  public java.lang.Object empty();
```

mosesn:
ahhhhhh, it's because DtabBase extends IndexedSeq[Dentry] which has an `empty` method.  the empty methods collide.

mosesn:
https://github.com/scala/scala/commit/a864ae04c48b9be94b00345668e8ecaf85fc24da
😠  ok, looks like we actually need to come up with a solution for this.

martijnhoekstra:
Yikes. 

What are our options here, other than lament the state of affairs we're in?

* Do a static forwarder object: that's a breaking change
* Find a different name for `empty`: that's also a breaking change, though it breaks less. It's also inelegant. 
* Find out if we can convince scalac to emit a static override. Is that a thing in the JVM? It's not a thing in Java, probbaly for good reason. This seems like a *really* bad idea
* Other?

mosesn:
I think I'm going to add a new name for java compat.  How does Dtab.emptyDtab sound to you?

martijnhoekstra:
A bit verbose, but I don't think there are much better options. Another option might be `blank` -- a blank table is empty, semantically it fits nicely, but that is probably a lot less discoverable, particularly through auto-completion tools.

mosesn:
the R = 100 * 1000 thing is bewildering. I think we're going to increase it for now, and reduce it again when we get a handle on what's going on.

martijnhoekstra:
The R = n thing seems to be fully deterministic btw.

What I noticed before was that `val classes = weightClasses.flatMap(weightClass.tupled).toSet` iterates over its values in a different order in 2.12 and 2.13, so that something may have randomly changed is plausible. That it's a red herring maybe more plausible.

changing the seed to `"seed2"` in `newDist` fixes the problem for me.

mosesn:
???? it sounds like this is just a bad test.  I tried changing the seed to "deadbeef" and that didn't work.

martijnhoekstra:
I don't know what I'm seeing and not seeing anymore.

mosesn:
I can reproduce that "seed2" fixes it.  This is wild.  We need to come back to this later and understand what's wrong with this test.

martijnhoekstra:
"seed2" break the test on 2.12. The test is just too sensitive to "small" diviations I suppose.

mosesn:
ugh, you're right.  I'll leave it as 100 * 1000 with "seed" for now.

mosesn:
@martijnhoekstra we had some questions about why we're switching to VectorBuilder instead of ArrayBuffer or ListBuffer (since we've found in the past that Vectors can be pretty heavy in comparison).  What's your thinking there?

martijnhoekstra:
As discussed on gitter -- the builders can build in O(1), but Vectors may not be suitable for anything. Getting a Builder for some collection is best done through `CC.newBuilder`, so a ListBuilder through List.newBuilder and an array backed seq through `ArraySeq.newBuilder`. This requires scala.collection.compat to make cross-compile.

mosesn:
this ended up being way more trouble than I thought it would be . . . but it's in!!! merged in here: https://github.com/twitter/finagle/commit/f21a54e0d759e005e089e9996113548078044498

thanks so much @martijnhoekstra!

martijnhoekstra:
Great! This should unblock all of finagle, and lots of other stuff as well.

PR #789, opened 2019-08-21
Update scalatest/scalacheck and address deprecations

martijnhoekstra:
Problem

Many deprecated constructs make upgrading difficult

Solution

Update deprecated terms and language constructs

Result

Far fewer deprecation warnings

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -121,8 +121,8 @@ val sharedSettings = Seq(
   crossScalaVersions := Seq("2.11.12", "2.12.8"),
   fork in Test := true, // We have to fork to get the JavaOptions
   libraryDependencies ++= Seq(
-    "org.scalacheck" %% "scalacheck" % "1.13.4" % "test",
-    "org.scalatest" %% "scalatest" % "3.0.0" % "test",
+    "org.scalacheck" %% "scalacheck" % "1.14.0" % "test",
+    "org.scalatest" %% "scalatest" % "3.0.8" % "test",
     // See https://www.scala-sbt.org/0.13/docs/Testing.html#JUnit
     "com.novocode" % "junit-interface" % "0.11" % "test",
     "org.mockito" % "mockito-all" % "1.9.5" % "test"
--- a/finagle-base-http/src/main/scala/com/twitter/finagle/http/ProxyCredentials.scala
+++ b/finagle-base-http/src/main/scala/com/twitter/finagle/http/ProxyCredentials.scala
@@ -2,13 +2,13 @@ package com.twitter.finagle.http
 
 import com.twitter.util.Base64StringEncoder
 import java.nio.charset.StandardCharsets
-import scala.collection.JavaConversions._
+import scala.collection.JavaConverters._
 
 object ProxyCredentials {
   def apply(credentials: java.util.Map[String, String]): Option[ProxyCredentials] =
-    apply(credentials.toMap)
+    apply(credentials.asScala)
 
-  def apply(credentials: Map[String, String]): Option[ProxyCredentials] = {
+  def apply(credentials: scala.collection.Map[String, String]): Option[ProxyCredentials] = {
     for {
       user <- credentials.get("http_proxy_user")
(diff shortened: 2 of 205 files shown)


martijnhoekstra:
CI failures seem spurious.

build.sbt
```diff
@@ -121,8 +121,8 @@ val sharedSettings = Seq(
   crossScalaVersions := Seq("2.11.12", "2.12.8"),
   fork in Test := true, // We have to fork to get the JavaOptions
   libraryDependencies ++= Seq(
-    "org.scalacheck" %% "scalacheck" % "1.13.4" % "test",
-    "org.scalatest" %% "scalatest" % "3.0.0" % "test",
+    "org.scalacheck" %% "scalacheck" % "1.14.0" % "test",
```
mosesn: we need to update scalacheck and scalatest atomically because we have a monorepo, so can you revert this for now?  if there's some reason why we need to update scalacheck and scalatest, we can work on making it happen internally.

finagle-base-http/src/main/scala/com/twitter/finagle/http/ProxyCredentials.scala
```diff
@@ -2,13 +2,13 @@ package com.twitter.finagle.http
...
-import scala.collection.JavaConversions._
+import scala.collection.JavaConverters._
 
 object ProxyCredentials {
   def apply(credentials: java.util.Map[String, String]): Option[ProxyCredentials] =
-    apply(credentials.toMap)
+    apply(credentials.asScala)
 
-  def apply(credentials: Map[String, String]): Option[ProxyCredentials] = {
+  def apply(credentials: scala.collection.Map[String, String]): Option[ProxyCredentials] = {
```
mosesn: this is changing the type from immutable to scala.collection.Map.  let's keep it immutable, imo.

finagle-http2/src/main/scala/com/twitter/finagle/http2/transport/AdapterProxyChannelHandler.scala
```diff
@@ -10,6 +10,8 @@ import io.netty.handler.codec.http.HttpClientUpgradeHandler.UpgradeEvent
 import io.netty.handler.codec.http.{HttpObject, LastHttpContent}
 import io.netty.util.ReferenceCountUtil
 import io.netty.util.concurrent.PromiseCombiner
+import io.netty.util.concurrent.ImmediateEventExecutor
+import io.netty.util.concurrent.{Future => NFuture}
```
mosesn: can we call this a NettyFuture instead?  I think that's what we typically do in finagle.

finagle-http2/src/main/scala/com/twitter/finagle/http2/transport/RichHttpToHttp2ConnectionHandler.scala
```diff
@@ -8,6 +8,8 @@ import io.netty.handler.codec.http._
 import io.netty.handler.codec.http2.Http2Exception.HeaderListSizeException
 import io.netty.handler.codec.http2._
 import io.netty.util.ReferenceCountUtil
+import io.netty.util.concurrent.{Future => NFuture}
```
mosesn: ditto

build.sbt
martijnhoekstra: These are the only versions published for 2.13, unfortunately.

mosesn: oh, we actually already updated to those versions internally, for exactly that reason.  my bad!!  thanks =)

mosesn:
Thanks!  I'll see if we can get this merged in soon.

martijnhoekstra:
Great, that will make the diff of the other pr drastically smaller.

yufangong:
Hey @martijnhoekstra, this is merged https://github.com/twitter/finagle/commit/b2ce58c8ef9c48de488b91c3d5348a66d2a803ac.
Thanks a lot!

PR #792, opened 2019-08-29
finagle: add jdk11 support to travis-ci

enbnt:
Problem

We want to test compile and runtime of JDK11.

Solution

Let's add it to the Travis CI config, but allow those tests
to fail while we work out the details.


enbnt:
This has been merged since October 24, 2019 (D365075).

PR #794, opened 2019-09-13
Thriftmux -> 213

martijnhoekstra:
Problem

No 2.13 cross-building for finagle-thrift, finagle-mux and finagle-thrift-mux

Solution

Cross-build.


finagle-thrift/src/main/scala/com/twitter/finagle/Thrift.scala
```diff
@@ -13,6 +13,7 @@ import com.twitter.finagle.param.{
 import com.twitter.finagle.server.{Listener, StackServer, StdStackServer}
 import com.twitter.finagle.service.{ResponseClassifier, RetryBudget}
 import com.twitter.finagle.stats.{ExceptionStatsHandler, StatsReceiver}
+import com.twitter.finagle
 import com.twitter.finagle.thrift.{ClientId => _, _}
```
mosesn: can we instead provide an alias for ClientId that's FinagleClientId?

mosesn:
thanks @martijnhoekstra this got merged here: https://github.com/twitter/finagle/commit/47ee31fd5bf93b49aa304dfef6aae39abdb9be17

PR #796, opened 2019-09-13
finagle-tunable: -> 2.13

martijnhoekstra:
Problem

no 2.13 build for finagle-tunable

Solution

trivial change: build update


mosesn:
We merged this in last week, but it looks like we lost appropriate attribution, I'm sorry about that.  https://github.com/twitter/finagle/commit/69ded534047f1115ad46de76bf293ab747d67ea5

PR #797, opened 2019-09-24
base-http: -> 2.13

martijnhoekstra:
Problem

no 2.13 cross build

Solution

~* HeaderMap: The big one. Rework into a trie as not to depend on HashTable's implementation.~
* Other maps: split into 2.12- and 2.13+ versions
* build: enable 2.13

Result

~Performance changes in 2.12 vs baseline are a mixed bag. ~

~* Create: 2.5 times as fast~
~* Create and add a single element: 1.3 times as fast~
~* Create and add 24 life-like headers: 34% slower~
~* Get one out of 20 random headers: 30% slower~
~* Get one out of 24 life-like headers: 2.98 times as slow (!!)~
~* Foreach: 2.5 times as fast~
~* Iterating keys and values separtely is slow. There is likely still a lot to get there.~


finagle-base-http/src/test/scala/com/twitter/finagle/http/TraceInfoTest.scala
```diff
@@ -62,7 +63,7 @@ class TraceInfoTest extends FunSuite {
     // clears and starts an unsampled trace
     TraceInfo.setClientRequestHeaders(req)
 
-    assert(req.headerMap.keys == Set("X-B3-TraceId", "X-B3-SpanId"))
+    assert(req.headerMap.keys.toSet == Set("X-B3-TraceId", "X-B3-SpanId"))
```
martijnhoekstra: todo: Why is this needed?

mosesn:
Hmm, how would you feel about breaking this up into two commits, one of which does the 2.13 work and one of which depends on the 2.13 one and has the Map changes?  In the mean time we can just stub out the Map stuff with trivial implementations (like the Seq one we discussed).

Also, @yufangong tested out removing the `extends Map` from HeaderMap and it seems that everything still compiles after we just add back in a couple of methods, so we might be able to simplify things by breaking the API a tiny bit.

martijnhoekstra:
The simplest thing that could possibly work for a backing store for Headers would be a plain mutable map over a string wrapper that compares case-insensitively -- easier and probably faster than the Seq one. It'll cost you an allocation per header. I can do that pretty quickly.

Re-evaluating whether something more sophisticated is a net positive can come at a later time.

martijnhoekstra:
@mosesn this version includes both a trie-backed version and (by default) a hashmap backed version. I don't have the resources to benchmark everything.

If you want to split out the trie implementation, you can just leave out **/TrieHeaderMap.scala and **/TrieHeaderMapBase.scala

finagle-base-http/src/main/scala/com/twitter/finagle/http/HeaderMap.scala
```diff
@@ -112,11 +103,124 @@ object HeaderMap {
...
+        }
+        builder.result()
+      }
+      
+    }
+
+    def iterated: Iterator[HeaderMap.NameValue] = {
+      var cur = this
+      new AbstractIterator[HeaderMap.NameValue](){
+        def hasNext = cur != null
```
mosesn: type annos please

finagle-base-http/src/main/scala/com/twitter/finagle/http/HeaderMap.scala
```diff
@@ -112,11 +103,124 @@ object HeaderMap {
...
+        def next() = {
```
mosesn: type annos please

finagle-base-http/src/main/scala/com/twitter/finagle/http/MapBackedHeaderMapBase.scala
```diff
@@ -0,0 +1,102 @@
...
+
+  // Does not validate key and value.
+  final def setUnsafe(key: String, value: String): HeaderMap = underlying.synchronized {
+    underlying.+=((key.toLowerCase, new Header(key, value)))
+    this
+  }
+
+  // ---- Map/MapLike -----
+
+  /*final*/ def get(key: String): Option[String] = underlying.synchronized {
```
mosesn: why is this commented out?

finagle-base-http/src/test/scala/com/twitter/finagle/http/DefaultHeaderMapTest.scala
```diff
@@ -16,12 +16,12 @@ class DefaultHeaderMapTest extends AbstractHeaderMapTest with ScalaCheckDrivenPr
   final def newHeaderMap(headers: (String, String)*): HeaderMap = DefaultHeaderMap(headers: _*)
 
   test("apply()") {
```
mosesn: can we push all of these tests down into AbstractHeaderMapTest?

finagle-base-http/src/test/scala/com/twitter/finagle/http/DefaultHeaderMapTest.scala
```diff
@@ -3,7 +3,7 @@ package com.twitter.finagle.http
 import org.scalacheck.Gen
 import org.scalatestplus.scalacheck.ScalaCheckDrivenPropertyChecks
 
-class DefaultHeaderMapTest extends AbstractHeaderMapTest with ScalaCheckDrivenPropertyChecks {
+class TrieHeaderMapTest extends AbstractHeaderMapTest with ScalaCheckDrivenPropertyChecks {
```
mosesn: let's make sure we're testing the Map-backed HeaderMap too

finagle-base-http/src/main/scala/com/twitter/finagle/http/HeaderMap.scala
```diff
@@ -112,11 +103,124 @@ object HeaderMap {
...
+          cur = cur.next
+          result
+        }
+      }
+    }
+
+    def values: Seq[String] =
+      if (next == null) value :: Nil
+      else {
+        val result = new mutable.ListBuffer[String] += value
```
mosesn: should we use List.newBuilder here too?

finagle-base-http/src/main/scala/com/twitter/finagle/http/HeaderMap.scala
```diff
@@ -112,11 +103,124 @@ object HeaderMap {
...
+          i = i.next
+        } while (i != null)
+
+        result.toList
+      }
+
+    def names: List[String] =
+      if (next == null) name :: Nil
+      else {
+        val result = new mutable.ListBuffer[String] += name
```
mosesn: should we use List.newBuilder here too?

finagle-base-http/src/main/scala/com/twitter/finagle/http/MapBackedHeaderMapBase.scala
martijnhoekstra: There was a comment about that. It got lost I think, I'll bring it back. https://github.com/scala/bug/issues/11749

martijnhoekstra:
I agree that extending Map doesn't bring much in principle. The problem with no longer extending Map is that you may break downstream: Map is part of the pubic signature and is exposed through request/response as Headers.

Is not extending map and possibly breaking downstream worth the benefit of not having to do that weird thing with version-dependent source directories? In the sbt world, I'd cautiously say no: there is very little version-specific code, and it's fairly discoverable. Not sure if the same goes for the pants/monorepo world. It probably also depends on whether any of your internal users use it as a Map.

finagle-base-http/src/test/scala/com/twitter/finagle/http/DefaultHeaderMapTest.scala
martijnhoekstra: Yes, that sounds good. Is there a reason why these tests aren't on `AbstractHeaderMapTest`? Should I duplicate them or move them there?

finagle-base-http/src/test/scala/com/twitter/finagle/http/DefaultHeaderMapTest.scala
martijnhoekstra: I really shouldn't read review comments bottom to top

mosesn:
We typically use our internal repository as a model for how we expect that people are using our code in the wild, since it gives us a bunch of examples.  @yufangong found that there are a pretty limited number of ways that people use Map-ish methods internally, so I think it should be safe to break.

@yufangong can you elaborate on what you found in your investigation?

luciferous:
Just starting to take a look, seems like a lot of work/thought has been put into this. Thanks for digging in, @martijnhoekstra.

I think I'm agreeing with @mosesn about breaking up the commit. Is there a smaller and simpler (even if not efficient) replacement implementation for HeaderMap that will allow us to break with HashTable? It looks to me that we'd need to spend a lot more time studying this trie-based approach, and maybe we can do that separately.

martijnhoekstra:
I can think of a couple of things to make the diff here smaller and more manageable. I'm going to propose a few preliminary PR's, and we can see for individual changes whether they work out, or want to go a different direction. That's for all or your time and effort y'all!

martijnhoekstra:
Diff-shrinker 1: https://github.com/twitter/finagle/pull/798

martijnhoekstra:
Diff-shrinker 2: https://github.com/twitter/finagle/pull/799

martijnhoekstra:
Diff-shrinker 3 will then introduce MapBackedHeaderMap on top of diff-shrinker 1 and 2, and then this PR can be rebased on top of that, do the 212/213 compat, and set MapBackedHeaderMap as the default for 2.13 (or for everything).

yufangong:
> We typically use our internal repository as a model for how we expect that people are using our code in the wild, since it gives us a bunch of examples. @yufangong found that there are a pretty limited number of ways that people use Map-ish methods internally, so I think it should be safe to break.
> 
> @yufangong can you elaborate on what you found in your investigation?
> 
> I think my bigger concern with the current approach is the sheer number of changes you had to make to get it to work. Right now this is pretty challenging to review–I can't really see what you copied over from before (or changed minimally) and what you changed substantially.

Sorry, I missed this. I first try to remove the HashMap Headers extended and leave something as ???, some internal things failed because of that (not compiling errors), so I assume not many models are relying on the HashMap impl. This PR seems now breaking `finagle-base-http`, so I cannot get an actual result.

martijnhoekstra:
This PR shouldn't be breaking. The failed CI is the `finagle-http` streaming test that's sometimes failing elsewhere too intermittently, but let's look at that again when the preliminary work is done, and it needs to be looked at all over again.

martijnhoekstra:
Diff-shrinker 3: https://github.com/twitter/finagle/pull/805

martijnhoekstra:
rebased on top of https://github.com/twitter/finagle/pull/805

When that gets in, rebase again on develop, which is equivalent to cherry-picking https://github.com/twitter/finagle/pull/797/commits/71dff8a9c6a31d5896ade880bbcca1c533b1b42d

I'll be on holiday for the next week, and won't have my laptop on me (and have different things on my mind other than rebasing pull requests).

@spockz could you maybe see if you can get this moving once #805 is merged?

martijnhoekstra:
rebased. Diff shrunk significantly. Is the builder approach of avoiding Vector and its builders still good?

finagle-base-http/src/test/scala/com/twitter/finagle/http/MapParamMapTest.scala
```diff
@@ -22,7 +22,7 @@ class MapParamMapTest extends FunSuite {
   }
 
   test("+") {
-    val paramMap = MapParamMap() + ("a" -> "1") + ("b" -> "2") + ("a" -> "3")
+    val paramMap = MapParamMap().setParam("a" -> "1").setParam("b" -> "2").setParam("a" -> "3")
```
martijnhoekstra: This is really unfortunate, but `+` is final and returns `Map`.

mosesn: HashMap gets around it somehow.  I think it's with StrictOptimizedMapOps, but I'm not sure.

martijnhoekstra: I got mixed up with `+=` on `mutable.Map` -- `+` is not final at all, specifically to allow overriding with a narrower return type.

mosesn: oh, phew!

martijnhoekstra: phew indeed

finagle-base-http/src/main/scala/com/twitter/finagle/http/CookieMap.scala
```diff
@@ -23,9 +23,9 @@ private[finagle] object CookieMap {
  * the ''message''
  */
 class CookieMap private[finagle] (message: Message, cookieCodec: CookieCodec)
-    extends mutable.Map[String, Cookie]
-    with mutable.MapLike[String, Cookie, CookieMap] {
-
+    extends CookieMapVersionSpecific(message, cookieCodec) with mutable.Map[String, Cookie]
+    {
+
```
h-vetinari: spurious spaces

finagle-base-http/src/main/scala/com/twitter/finagle/http/CookieMap.scala
```diff
@@ -113,28 +113,28 @@ class CookieMap private[finagle] (message: Message, cookieCodec: CookieCodec)
...
-  def +=(cookie: (String, Cookie)): this.type = {
+  def addCookie(cookie: (String, Cookie)): this.type = {
     val (n, c) = cookie
     setNoRewrite(n, c)
     rewriteCookieHeaders()
     this
   }
 
-  /**
+   /**
```
h-vetinari: spurious space

finagle-base-http/src/main/scala/com/twitter/finagle/http/headers/JTreeMapBackedHeaderMap.scala
```diff
@@ -102,10 +99,11 @@ final private class JTreeMapBackedHeaderMap extends HeaderMap {
     }
   }
 
-  def removed(key: String): this.type = underlying.synchronized {
+  def removeHeader(key: String): this.type = underlying.synchronized {
     underlying.remove(key)
     this
   }
+
```
h-vetinari: potentially spurious newline

martijnhoekstra:
Surely, you mean to say that those files don't have spurious final newlines

h-vetinari:
Git (and github) are quite clear that the _lack_ of final newlines is not desired (`git diff` will complain, and the github UI shows it as missing as well). ;-)

If it's an intentional choice then sure - but there is a final newline in `DefaultHeaderMap.scala` for example, so it's not even consistent...

martijnhoekstra:
Neither Git, nor Github complains to me. Maybe you have different settings, or different interpretations of the same interface.

I added a commit I removing the spurious newline at the end of DefaultHeaderMap for consistencies sake though.

h-vetinari:
> Neither Git, nor Github complains to me. Maybe you have different settings, or different interpretations of the same interface.

Running
`git diff upstream/master -- finagle-base-http/src/main/scala-2.12-/com/twitter/finagle/http/headers/D*` on this PR yields
```
diff --git a/finagle-base-http/src/main/scala-2.12-/com/twitter/finagle/http/headers/DefaultHeaderMap.scala b/finagle-base-http/src/main/scala-2.12-/com/twitter/finagle/http/headers/DefaultHeaderMap.scala
new file mode 100644
index 000000000..ad55b6933
--- /dev/null
+++ b/finagle-base-http/src/main/scala-2.12-/com/twitter/finagle/http/headers/DefaultHeaderMap.scala
@@ -0,0 +1,7 @@
+package com.twitter.finagle.http.headers
+
+import com.twitter.finagle.http.HeaderMap
+
+object DefaultHeaderMap {
+  def apply(headers: (String, String)*): HeaderMap = HashBackedHeaderMap(headers: _*)
+}
\ No newline at end of file
```
and similarly for all the other files missing a newline.  Note the last line - that's git complaining roughly because of [the posix standard](https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline). You'll also see it in your diff on github as a crossed red circle with a return arrow.

martijnhoekstra:
I see it mentioning the lack of a newline at the end of the file, but not complaining about it. The Scala spec doesn't refer to POSIX with regards to its lines either.

mosesn:
fwiw, whether files have newlines at the end of them or not has no bearing on whether they'll be merged in or not.

@martijnhoekstra is this ready to rock?  should we review it?

martijnhoekstra:
yeah, this is ready to rock @mosesn

mosesn:
awesome!  we're heads down on a tech debt sprint this week, but we'll review your patch next week.  thanks so much for going through all of this trouble, I'm pretty excited about where we're going to end up.

martijnhoekstra:
@mosesn great to hear! The rest of finagle-http is pretty review-light. I've tested the commits by @spockz at https://github.com/martijnhoekstra/finagle/pull/1/commits , and as far as I'm concerned they look good to me to be cherry-picked in as well. Can we make that happen within the scope of this PR as well? It would be really nice if we can get finagle-http published this year still.

finagle-base-http/src/main/scala/com/twitter/finagle/http/CookieMap.scala
```diff
@@ -128,13 +128,13 @@ class CookieMap private[finagle] (message: Message, cookieCodec: CookieCodec)
     this += ((cookie.name, cookie))
   }
 
-  override def ++=(xs: TraversableOnce[(String, Cookie)]): this.type = {
```
bryce-anderson: This looks like it's replaced with `addCookies`, and thus an API change.

finagle-base-http/src/main/scala-2.13+/com/twitter/finagle/http/CookieMapVersionSpecific.scala
```diff
@@ -0,0 +1,16 @@
...
+
+  def addCookie(cookie: (String, Cookie)): this.type
+  def addCookies(cookies: IterableOnce[(String, Cookie)]): this.type
+  def removeCookie(name: String): this.type
+  def removeCookies(names: IterableOnce[String]): this.type
+
+  def addOne(cookie: (String, Cookie)): this.type = addCookie(cookie)
+  def subtractOne(name: String): this.type = removeCookie(name)
+
+}
```
bryce-anderson: Can some of these be protected? And I'm not sure where the `addCookies` and `removeCookies` are used.

finagle-base-http/src/main/scala-2.12-/com/twitter/finagle/http/headers/HashBackedHeaderMap.scala
```diff
@@ -7,7 +7,7 @@ import com.twitter.finagle.http.HeaderMap
  * Mutable, thread-safe [[HeaderMap]] implementation backed by a
  * HeaderHash.
  */
-private final class HashBackedHeaderMap extends http.HeaderMap {
+private[http] final class HashBackedHeaderMap extends http.HeaderMap {
```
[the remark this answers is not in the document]
martijnhoekstra: The other PR was merged keeping the old as the default, so I figured it was a good idea to keep it here too. @mosesn what's your take on that?

finagle-base-http/src/main/scala-2.13+/com/twitter/finagle/http/CookieMapVersionSpecific.scala
martijnhoekstra: Yeah, those sound like good plans.

finagle-base-http/src/main/scala-2.12-/com/twitter/finagle/http/headers/HashBackedHeaderMap.scala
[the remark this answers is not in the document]
mosesn: I think Bryce's proposal makes sense, we're going to swap the default after we finish testing it, after all.

finagle-base-http/src/main/scala-2.13+/com/twitter/finagle/http/ParamMapVersionSpecific.scala
```diff
@@ -0,0 +1,12 @@
+
+package com.twitter.finagle.http
+
+import scala.collection.immutable
+protected abstract class ParamMapVersionSpecific extends immutable.Map[String, String] {
+  def setParam[B >: String](kv: (String, B)): ParamMap
+  def updated[V1 >: String](key: String, value: V1): ParamMap = setParam(key, value)
+
+  def clearParam(name: String): ParamMap
+  def removed(name: String): ParamMap = clearParam(name)
```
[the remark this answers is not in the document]
martijnhoekstra: I've made those protected

finagle-base-http/src/main/scala/com/twitter/finagle/http/CookieMap.scala
martijnhoekstra: in 2.12, ++= comes from the override in `CookieMapVersionSpecific`
in 2.13, ++= comes from `Map` where it is an alias for `addAll` which is overridden in `CookieMapVersionSpecific`

finagle-base-http/src/main/scala-2.12-/com/twitter/finagle/http/headers/HashBackedHeaderMap.scala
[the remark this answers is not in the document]
martijnhoekstra: Implemented in https://github.com/twitter/finagle/pull/797/commits/caf06dcf8b1e6d6d6ddd6f921a577f321b5924ff

yufangong:
Hey @martijnhoekstra, it looks like there are some conflicts between the branch and develope (DefaultHeaderMap), can you do a rebase? Thank you!

martijnhoekstra:
rebased

martijnhoekstra:
test failures look spurious

yufangong:
Sorry for the delay, just got back from holiday breaks.
One thing caught our eyes that `ArraySeq.newBuilder`, we have found out `ArraySeq.newBuilder` being more expensive than `new ArrayBuffer`, see https://github.com/twitter/finagle/commit/34a2c66d14a3ab42973ca4b201f3d1cc4c13ee6b.
Is there any reason we want to use ArraySeq here? Otherwise, I will just change this part to use ArrayBuffer on our end. Thanks.

martijnhoekstra:
Calling `toSeq` on a mutable seq in 2.12 is a noop, but is a conversion in 2.13, usually involving copying the entire structure and should be much slower than the builder/result construct.

The difficulty is balancing the requirements of staying on `collection.Seq` in both 2.12 and 2.13, not regressing 2.12 performance (too much), keeping the diff as small as possible, and not regressing 2.12 -> 2.13 too much. I thought I found a good balance there with the builder, but as the facts show, the regressions on 2.12 are still significant.

yufangong:
Oh so good to know, we neglected 2.13 performance before, this is a good sign for us to start building some perf tools running on 2.12 and 2.13. 

And yes, the regressions on 2.12 are showing pretty significant perf drops by using builder/result. so we will have to keep the `toSeq` around for the majority. Maybe worth to file an issue as well to start the discussion on 2.13?

martijnhoekstra:
@yufangong I created https://github.com/twitter/finagle/issues/818

yufangong:
Merged here: https://github.com/twitter/finagle/commit/3007e5db0279721194bfeef5f23c13f34aa03fa3.
Thank you!

PR #798, opened 2019-09-27
consolidate header map tests

martijnhoekstra:
tests that apply to any header map
are consolidated into AbstractHeaderMapTest

individual additional tests can go in extensions
but so far, that's not needed.

Problem

Tests for HeaderMap are split between AbstractHeaderMapTest and DefaultHeaderMapTest. When additional HeaderMap implementations were to be proposed other than the current DefaultHeaderMap (and they just might), tests will need to be copied for each implementation.

Solution

Move tests from DefaultHeaderMapTest to AbstractHeaderMapTests

Result

Any HeaderMap implementation can be tested fully by extending AbstractHeaderMapTests and setting the newHeaderMap constructor, making for a small diff and still have good test coverage if such a thing were to ever happen.


luciferous:
Looks good, also assuming tests were just moved, please call out if not.

yufangong:
landed https://github.com/twitter/finagle/commit/aec8f3befe848c121f16a036ff357af180a667a0
iirc this and https://github.com/twitter/finagle/pull/799 have some related changes, if thats right do you mind doing a rebase on that one? ty!

martijnhoekstra:
Thanks! I rebased #799 on top of this. Its failing test now pass, suggesting that the test is indeed just flaky.

PR #799, opened 2019-09-27
Split header files

martijnhoekstra:
Problem

DefaultHeaderMap.scala contains generic stuff that is usable for any HeaderMap, like the Header class, and header validation in the companion. 

Solution

Factor out individual units to their own files, rename the class DefaultHeaderMap to HashBackedHeaderMap.

Result

New HeaderMap implementations don't have to depend on DefaultHeaderMap. 

Discussion

This is primarily intended to decrease diff size for future HeaderMap implementations to be added. It leaves a lot of files in the http directory though. Some plausible solutions to that are not worrying and calling it good, including the backingstore implementation into the map implementation file, or moving all header-related files to their own directory and package. Maybe then give dispensation to HeaderMap and DefaultHeaderMap, in order to not break the public API, or include forwarders/aliases in the package object.

Also, I retained a DefaultHeaderMap constructor, but I don't like it much, and I'm actually thinking that should just be HeaderMap.apply


martijnhoekstra:
Note: functionally this is a noop. No implementation is changed, and no public API/name is changed.

martijnhoekstra:
Added a commit to package all things headers into their own package. Dispensation for the public entrypoint HeaderMap, and for the header validation module, which get forwarders in the http package object. I think it's a net positive, but sensible people can argue about that.

martijnhoekstra:
I can't explain how https://github.com/twitter/finagle/pull/799/commits/d4beb9fcd1f0138e1fad920278ababc974f8a04f causes the CI failure. I think it's spurious, but it's closeby enough that it's probably worth investigating.

martijnhoekstra:
`Http2StreamingTest` now passes (though it fails on current develop). It doesn't look like something that ought to be flakey or non-deterministic, but it seems it is. Probably a sperate issue out of the scope of this migration.

yufangong:
> `Http2StreamingTest` now passes (though it fails on current develop). It doesn't look like something that ought to be flakey or non-deterministic, but it seems it is. Probably a sperate issue out of the scope of this migration.

Thanks for reporting, we are going to investigate that test flaky as another task internally, in the meantime, let me pull this PR in!

yufangong:
Merged here https://github.com/twitter/finagle/commit/d5767bb0d26523b246a3c3ed19a4589a0bd746b9, in order to reduce breaking api changes, we moved HeaderMap back to the finagle.http package, while keeping other header related files in f.t.header. thanks!

PR #803, opened 2019-10-03
Enable 2.13 build for finagle-zipkin-{core,scribe}

DieBauer:
finagle-zipkin-{core,scribe}: Compile for scala 2.13

Problem

finagle-zipkin-{core,scribe} is not available for scala 2.13

Solution

Cross compile the modules for scala 2.13

Result
Have finagle-zipkin-{core,scribe} available for scala 2.13

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -465,7 +465,8 @@ lazy val finagleZipkinCore = Project(
   id = "finagle-zipkin-core",
   base = file("finagle-zipkin-core")
 ).settings(
-  sharedSettings
+  sharedSettings,
+  withTwoThirteen
 ).settings(
   name := "finagle-zipkin-core",
   libraryDependencies ++= Seq(
@@ -478,7 +479,8 @@ lazy val finagleZipkinScribe = Project(
   id = "finagle-zipkin-scribe",
   base = file("finagle-zipkin-scribe")
 ).settings(
-  sharedSettings
+  sharedSettings,
+  withTwoThirteen
 ).settings(
   name := "finagle-zipkin-scribe",
   libraryDependencies ++= scroogeLibs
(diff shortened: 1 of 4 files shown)


DieBauer:
Added zipkin-scribe as well. Tests pass with `+finagle-zipkin-scribe/test`

mosesn:
I think the default Seq is immutable now, not mutable.  Let's prefer the immutable Seq!

DieBauer:
Alright! The interface of logSpan used to accept both. So let me see if I can change that. I'll also see if is already some benchmarking present.

mosesn:
@DieBauer did you find any benchmarking stuff you wanted to poke around at?  and does the current version build on 2.13?

DieBauer:
I was looking into it, but as it turns out eventhough the build of `finagle-benchmark` depends on `finagle-zipkin-scribe` there is no actual source dependency. So benchmarking would mean from scratch. 

This would mean we can also drop the dependency on `scribe` in the benchmark. 

I'm not sure what other benchmark I can perform, any idea? 

What is now in this branch builds and tests on my machine for `+test`.

The build failing looks like a flaky Travis test: 

```
[info] Http2StreamingTest:

[info] - client: discard respond reader *** FAILED ***
[info]   Some(Return(Response("HTTP/1.1 Status(200)"))) was not empty (AbstractStreamingTest.scala:151)
```

DieBauer:
Yes I did :)

mosesn:
great, thanks!  we're cutting a release right now (I don't think this is going to make it into the October release, unforch) and we'll merge it in once that's wrapped up.

jyanJing:
Thank you for working on this @DieBauer ! The change is merged in [68e7a04](https://github.com/twitter/finagle/commit/68e7a04c06c10a126edb3b7766ef263ce2769088)

PR #805, opened 2019-10-04
finagle-base-http: add MapBackedHeaderMap

martijnhoekstra:
Problem

The default headermap implementation is not future-proof, and relies on HashMap internals that will change in scala 2.13 

Solution

Provide an alternative headermap implemtation that is backed by Map, an interface that will work across versions. 

Result

The option to use a 2.13-proof headermap.


martijnhoekstra:
Needs to be benchmarked. I'll cherry-pick the benchmark extensions from https://github.com/twitter/finagle/pull/797/files#diff-b61dcd2fdf0b7a577f018c14e2d83800 into this, and we can bench-off

martijnhoekstra:
flaky test error presumed spurious

martijnhoekstra:
https://docs.google.com/spreadsheets/d/1C402zjjqmEhC-Aw5LOf_XwzjYUFcPX1XT0kI-B_7lVQ/edit?usp=sharing has benchmarks (while laptop wasn't fully idle, so a bit dirty).

get takes four times as long in this impl vs baseline. How that reflects to real world performance is anybodies guess without macrobenchmarking/end-to-end tests

mosesn:
Thanks!  I really appreciate you breaking up the patch, it's much easier to read now.

mosesn:
How does it work if it's backed by a java hashmap?  I wonder if that will make it easier to avoid allocs etc

martijnhoekstra:
I can run another bench backed with a java hashmap, but if we want real insights in allocs, we should run under JFR, and analyse those results. I never did that before though. 

I'm also not too sure the performance bottleneck hinges on allocation. That's also something that JFR dumps or flamegraphs could help us out more with than looking at microbenchmarks and guessing.

This weekend my volunteer/life balance will favour the living, so if there will be anything from me, it'll be after the weekend.

mosesn:
sgtm, no rush!  jmh has an alloc mode also, I think it's `-prof gc`

martijnhoekstra:
Also related, the 2.13 Map implementation changed for performance reasons, and should be faster than the 2.12 Map. If performance of the 2.13 one is good (that's a big if, but still), you could just keep the current impl as the default for 2.12. It's just a matter of making what `HeaderMap.apply` returns version specific.

bryce-anderson:
I have an idea along the lines moses is suggesting that might be worth trying. We should try using a Java collection as the backing collection but use the `java.util.TreeMap` which allows you to provide a `Comparitor[String]` that is not case sensitive. For example this should work and be pretty snappy:
```scala
val SharedComparitor = new Comparator[String] {
  def compare(o1: String, o2: String): Int = {
    // Shorter strings are always less, regardless of their content
    val lenthDiff = key1.length - key2.length
    if (lenthDiff != 0) lenthDiff
    else {
      @tailrec
      def go(i: Int): Int = {
        if (i == key1.length) 0 // end, they are equal.
        else {
          val char1 = hashChar(key1.charAt(i))
          val char2 = hashChar(key2.charAt(i))
          val diff = char1 - char2
          if (diff == 0) go(i + 1)
          else diff
        }
      }
      go(0)
    }
  }
```
There's a really solid chance this would be faster than what is currently in master.

martijnhoekstra:
Sounds good! I also still want to give the trie approach another go. I also want to understand why the current approach is so much faster in finding an entry than the scala map backed approach is.

bryce-anderson:
I'm general, String.toLowerCase is really expensive both in terms of course
and allocations, so that's probably where you're losing perf.

martijnhoekstra:
That sounds plausible. For the plain tree version I'm afraid to lose a lot of performance on traversing a common prefix several times (e.g. content-*). At least it'll be good to have options to compare.

martijnhoekstra:
Some more benchmark results at https://docs.google.com/spreadsheets/d/1C402zjjqmEhC-Aw5LOf_XwzjYUFcPX1XT0kI-B_7lVQ/edit#gid=578804215 (ran with too few iterations for too short for real results) indicate that the java tree map approach so far is indeed comes pretty close to the implementation master.

finagle-base-http/src/main/scala/com/twitter/finagle/http/headers/JMapBackedHeaderMap.scala
```diff
@@ -0,0 +1,88 @@
...
+
+  final override def keysIterator: Iterator[String] = underlying.synchronized {
+    underlying.values.iterator.asScala.flatMap(header => {
+      def rec(uniq: List[String], todo: List[String]): List[String] = todo match {
+        case Nil => uniq
+        case (head :: tail) =>
+          rec(head :: uniq, todo.filterNot(x => x == head))
+      }
+      rec(Nil, header.names.toList)
+    })
```
bryce-anderson: I don't think this is a new issue, but the iterator may live long enough for the headers to get modified and if it does we face a `ConcurrentModificationException`. Perhaps we should extract the `underlying.values.toArray(new Array[Headers](underlying.size)` operation as a way to make pseudo-safe iterators. I say pseudo-safe since the `Headers` linked list can still mutate, but it should be strictly additive in nature.

finagle-base-http/src/main/scala/com/twitter/finagle/http/headers/MapBackedHeaderMap.scala
```diff
@@ -0,0 +1,108 @@
...
+  final override def keysIterator: Iterator[String] = underlying.synchronized {
+    underlying.valuesIterator.flatMap(header => {
+      def rec(uniq: List[String], todo: List[String]): List[String] = todo match {
+        case Nil => uniq
+        case (head :: tail) =>
+          rec(head :: uniq, todo.filterNot(x => x == head))
+      }
+      rec(Nil, header.names.toList)
+    })
+  }
```
bryce-anderson: I suspect this is costly both in terms of CPU and allocations. In the common case we will have only one entry in in the Headers linked list. Maybe we can make a hybrid implementation to the tune of
```scala
new Iterator[String] {
  private[this] val underlying = safeEntriesIterator() // the extracted array-backed iterator talked about in other comments
  private[this] var currentEntries: List[String] = Nil
  def hasNext = !currentEntries.isEmpty || underlying.hasNext
  def next(): String = {
    if (currentEntries.isEmpty) {
      val hs = underlying.next()
      if (hs.next == null) {
         // Take a shortcut for single value keys which is the common case
         hs.name
      } else {
        currentEntries = gatherEntries???(hs) // extract all unique and recurse one step since we know there will be currentEntries
        next()
      }
    } else {
       ??? // take one from currentEntries
    }
  }
}
```
It might be even better to use a second iterator instead of a `List` for `currentEntries`.

finagle-base-http/src/main/scala/com/twitter/finagle/http/headers/JMapBackedHeaderMap.scala
```diff
@@ -0,0 +1,88 @@
...
+
+
+  def get(key: String): Option[String]
+
+  final def iterator: Iterator[(String, String)] = underlying.synchronized {
+    val underlyingEntries = underlying.entrySet.toArray
+    underlyingEntries.toIterator.flatMap {
+      case es: java.util.Map.Entry[String @unchecked, Header @unchecked] => {
+        es.getValue.iterator.map(nv => (nv.name, nv.value))
+      }
```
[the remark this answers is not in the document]
martijnhoekstra: is there a benefit to pre-allocating the array like that, or is it equivalent to just `toArray`?

bryce-anderson:
@martijnhoekstra, I just wanted to give you a heads up, but I'm working on some optimizations in Headers.scala. It looks like conflicts will be minimal, but just wanted you to know.

martijnhoekstra:
Thanks for letting me know, I'm sure the conflicts won't be a big deal.

martijnhoekstra:
addressed the feedback, and rebased for the Header changes

martijnhoekstra:
@bryce-anderson I think I got them all. Do you have any additional issues?

finagle-base-http/src/main/scala/com/twitter/finagle/http/headers/Header.scala
```diff
@@ -15,6 +16,20 @@ private[http] object Header {
 
   final class Root private[Header] (name: String, value: String) extends Header(name, value) {
 
+    def iterator: Iterator[HeaderMap.NameValue] =
+      if (next == null) Iterator.single(this)
+      else {
+        var cur: Header = this
```
bryce-anderson: If you move `cur` into the definition of `AbstractIterator` you can avoid lifting it to the the heap via `scala.runtime.ObjectRef`.

finagle-base-http/src/main/scala/com/twitter/finagle/http/headers/JMapBackedHeaderMap.scala
```diff
@@ -0,0 +1,99 @@
...
+
+  /**
+   * Underlying headers eagerly copied to an array, without synchronizing
+   * on the underlying collection.
+   */
+  private[this] def copyHeaders: Iterator[Header.Root] = underlying.values.toArray(new Array[Header.Root](underlying.size)).iterator
+
+  final def iterator: Iterator[(String, String)] = underlying.synchronized {
+    copyHeaders.flatMap(_.iterator.map(nv => (nv.name, nv.value)))
+  }
```
bryce-anderson: This looks like `nameValueIterator.map(nv => (nv.name, nv.value))`.

finagle-base-http/src/main/scala/com/twitter/finagle/http/headers/JMapBackedHeaderMap.scala
```diff
@@ -0,0 +1,99 @@
...
+import com.twitter.finagle.http.HeaderMap
+import java.util.function.BiConsumer
+import scala.collection.AbstractIterator
+
+/**
+ * Mutable, thread-safe [[HeaderMap]] implementation, backed by
+ * a mutable [[Map[String, Header]]], where the map key
+ * is forced to lower case
+ */
+private[http] trait JMapBackedHeaderMap extends HeaderMap {
```
bryce-anderson: This feels like it should be an `abstract class`, and probably just vanilla `private`.

finagle-base-http/src/main/scala/com/twitter/finagle/http/headers/JTreeMapBackedHeaderMap.scala
```diff
@@ -0,0 +1,79 @@
...
+  
+  override val underlying: java.util.TreeMap[String, Header.Root] = 
+    new java.util.TreeMap[String, Header.Root](JTreeMapBackedHeaderMap.SharedComparitor)
+
+    def getAll(key: String): Seq[String] = underlying.synchronized {
+      underlying.get(key.toLowerCase) match {
+        case null => Nil
+        case r: Header.Root => r.values
+      }
+    }
```
bryce-anderson: Indentation is off.

finagle-base-http/src/main/scala/com/twitter/finagle/http/headers/JTreeMapBackedHeaderMap.scala
```diff
@@ -0,0 +1,79 @@
...
+
+import com.twitter.finagle.http.HeaderMap
+import scala.annotation.tailrec
+
+/**
+ * Mutable, thread-safe [[HeaderMap]] implementation, backed by
+ * a mutable [[Map[String, Header]]], where the map key
+ * is forced to lower case
+ */
+final private[http] class JTreeMapBackedHeaderMap extends JMapBackedHeaderMap {
```
bryce-anderson: What is the value of separating this from `JMapBackedHeaderMap`? I don't anticipate we'll want to change the underlying map all that often.

finagle-base-http/src/main/scala/com/twitter/finagle/http/headers/JTreeMapBackedHeaderMap.scala
```diff
@@ -0,0 +1,79 @@
...
+    def getAll(key: String): Seq[String] = underlying.synchronized {
+      underlying.get(key.toLowerCase) match {
+        case null => Nil
+        case r: Header.Root => r.values
+      }
+    }
+
+  // Does not validate key and value.
+  def addUnsafe(key: String, value: String): this.type  = underlying.synchronized {
+    def header = Header.root(key, value)
```
bryce-anderson: `header` is only needed if we hit the `null` case. Can we move its instantiation there?

finagle-base-http/src/main/scala/com/twitter/finagle/http/headers/JTreeMapBackedHeaderMap.scala
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  // Does not validate key and value.
+  def setUnsafe(key: String, value: String): this.type = underlying.synchronized {
+    underlying.put(key, Header.root(key, value))
+    this
+  }
+
+  def get(key: String): Option[String] = underlying.synchronized {
+    Option(underlying.get(key)).map(_.value)
```
bryce-anderson: Let's do a `match` to avoid allocating a second `Some` and the lambda used to `map`.

finagle-base-http/src/main/scala/com/twitter/finagle/http/headers/JTreeMapBackedHeaderMap.scala
bryce-anderson: fwiw, I'm not strongly opinionated about this, it's just that having an abstract backing map that is synchronized on in two different traits/classes makes me queasy.

martijnhoekstra: It made sense when the HashMap variety still existed. Not so much anymore.

finagle-base-http/src/main/scala/com/twitter/finagle/http/headers/JTreeMapBackedHeaderMap.scala
martijnhoekstra: It's a def for that reason, but it can go inside the match. The only reason it's outside now is neater indentation, so that's not much of a reason.

bryce-anderson: Oh, I didn't notice it was a def. :) Maybe just inline it where `header` is used and we'll both be happy.

As an aside, do you know if that def allocates a lambda, or does that get lifted out into the object itself?

martijnhoekstra: Yes, I inlined it.

As for the allocation, I assumed it gets lifted to a method all the way to the top, especially since it doesn't capture anything and doesn't escape scope, but I'm not 100% sure either way.

finagle-base-http/src/main/scala/com/twitter/finagle/http/headers/JTreeMapBackedHeaderMap.scala
```diff
@@ -2,29 +2,102 @@ package com.twitter.finagle.http.headers
...
+        else {
+          val h = valuesIterator.next()
+          if (h.next == null) h.name
+          else {
+            currentEntries = h.iterator.map(nv => nv.name).toList.distinct.iterator
+            currentEntries.next()
+          }
+        }
+    }
+  }
```
bryce-anderson: This looks pretty similar in result to `HeadersHash.uniqueNamesIterator`. Could we share that logic?

martijnhoekstra: I factored the (lazier) logic of HeaderHash out, and made both call that.

martijnhoekstra:
CI failures look spurious

finagle-base-http/src/main/scala/com/twitter/finagle/http/headers/Header.scala
```diff
@@ -14,6 +16,21 @@ private[http] sealed class Header private (final val name: String, final val val
 private[http] object Header {
 
   final class Root private[Header] (name: String, value: String) extends Header(name, value) {
+    self => 
+
+    def iterator: Iterator[HeaderMap.NameValue] =
+      if (next == null) Iterator.single(this)
+      else {
+        new AbstractIterator[HeaderMap.NameValue] {
+          var cur: Header = self
```
vkostyukov: private?

finagle-base-http/src/main/scala/com/twitter/finagle/http/headers/JTreeMapBackedHeaderMap.scala
```diff
@@ -0,0 +1,139 @@
...
+final private class JTreeMapBackedHeaderMap extends HeaderMap {
+  import HeaderMap._
+
+  // In general, Map's that are not thread safe are not
+  // durable to concurrent modification and can result in infinite loops
+  // and exceptions.
+  // As such, we synchronize on the underlying collection when performing
+  // accesses to avoid this. In the common case of no concurrent access,
+  // this should be cheap.
+  val underlying: java.util.TreeMap[String, Header.Root] =
```
vkostyukov: private?

finagle-base-http/src/main/scala/com/twitter/finagle/http/headers/JTreeMapBackedHeaderMap.scala
```diff
@@ -0,0 +1,139 @@
...
+  new java.util.TreeMap[String, Header.Root](JTreeMapBackedHeaderMap.SharedComparitor)
+
+  private def foreachConsumer[U](f: ((String, String)) => U):
+    BiConsumer[String, Header.Root] = new BiConsumer[String, Header.Root](){
+      def accept(key: String, header: Header.Root): Unit = header.iterator.foreach(
+        nv => f(nv.name, nv.value)
+      )
+    }
+
+  override def foreach[U](f: ((String, String)) => U): Unit =
```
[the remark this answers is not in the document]
martijnhoekstra: Now I wish I never removed the abstraction to swap out the underlying Map implementation. 

But let's take a step back. How often do people really concurrently iterate over headers and also mutate the header map and don't synchronize that access because they don't care whether or not the modified data should or shouldn't be part of the iteration? Is that a use-case you want to facilitate in the first place?

Thinking about those questions, my conclusion is that throwing a ConcurrentModificationException that makes the user aware of the situation/race condition/potential bug is better than synchronizing and not being deterministic in what happens first. If the user *really* wants that for some unforeseen, unspoken and unspeakable reason, they can always still externally synchronize (maybe with a ReadWriteLock or something).

I think in terms of common usage patterns, that's the fastest and most reasonable thing to do too. Even if a ConcurrentSkipListMap would be pretty cool.

If we do want to get overboard and fancy, it should be validated with benchmarks that reflect real-world usage patterns, including the real-world concurrent patterns, because otherwise we're not really measuring anything close to reality.

martijnhoekstra:
> This looks pretty great @martijnhoekstra! Any chance you can attach benchmark results to the PR description?

Shall I re-run after I do the last round of feedback (and we know what to do with potentially concurrent modification and iteration?)

martijnhoekstra:
quick ( -i 5 -wi 3 -r 1 -w 5)  benchmark run:

```
[info] Benchmark                                          Mode  Cnt     Score     Error  Units
[info] HashBackedHeaderMapBenchmark.create                avgt    5    58.571 ▒  18.232  ns/op
[info] HashBackedHeaderMapBenchmark.createAndAdd          avgt    5   148.290 ▒  28.821  ns/op
[info] HashBackedHeaderMapBenchmark.createAndAdd24        avgt    5  3435.482 ▒ 594.920  ns/op
[info] HashBackedHeaderMapBenchmark.getRandom             avgt    5    28.593 ▒   5.335  ns/op
[info] HashBackedHeaderMapBenchmark.getRealistic          avgt    5    26.842 ▒   2.421  ns/op
[info] HashBackedHeaderMapBenchmark.iterate               avgt    5   265.780 ▒  56.199  ns/op
[info] HashBackedHeaderMapBenchmark.iterateKeys           avgt    5   142.899 ▒  19.002  ns/op
[info] HashBackedHeaderMapBenchmark.iterateRepeatedKeys   avgt    5   109.857 ▒  48.143  ns/op
[info] HashBackedHeaderMapBenchmark.iterateValues         avgt    5   196.082 ▒ 113.488  ns/op
[info] HashBackedHeaderMapBenchmark.keySet                avgt    5     5.904 ▒   1.852  ns/op
[info] JTMapBackedHeaderMapBenchmark.create               avgt    5    15.710 ▒   3.428  ns/op
[info] JTMapBackedHeaderMapBenchmark.createAndAdd         avgt    5    78.716 ▒  18.136  ns/op
[info] JTMapBackedHeaderMapBenchmark.createAndAdd24       avgt    5  2242.704 ▒ 450.955  ns/op
[info] JTMapBackedHeaderMapBenchmark.getRandom            avgt    5    41.988 ▒   6.247  ns/op
[info] JTMapBackedHeaderMapBenchmark.getRealistic         avgt    5    41.815 ▒  24.553  ns/op
[info] JTMapBackedHeaderMapBenchmark.iterate              avgt    5   164.154 ▒  64.958  ns/op
[info] JTMapBackedHeaderMapBenchmark.iterateKeys          avgt    5   141.076 ▒  34.300  ns/op
[info] JTMapBackedHeaderMapBenchmark.iterateRepeatedKeys  avgt    5   110.051 ▒  32.983  ns/op
[info] JTMapBackedHeaderMapBenchmark.iterateValues        avgt    5   353.827 ▒  40.716  ns/op
[info] JTMapBackedHeaderMapBenchmark.keySet               avgt    5     6.344 ▒   2.265  ns/op
```

finagle-base-http/src/main/scala/com/twitter/finagle/http/headers/JTreeMapBackedHeaderMap.scala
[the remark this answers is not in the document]
mosesn: I think the *good* outcome is that they get a ConcurrentModificationException–a bad outcome might be an infinite loop or worse.  See https://ivoanjo.me/blog/2018/07/21/writing-to-a-java-treemap-concurrently-can-lead-to-an-infinite-loop-during-reads/

I think correctness needs to be our #1 concern.  I agree with you that it's extremely unlikely–that's one of the reasons I'm willing to just wrap it in a synchronized block and call it a day.

I'm not sure I understand your concern about not being deterministic about what happens first–that's always going to be the case with concurrent use of a data structure.  My concern is that HeaderMap used to be threadsafe, so people may have been already using it assuming it was threadsafe, and then one day we merge this in and their application breaks in an extremely subtle way.  If this was a new abstraction, I would be open to considering, "HeaderMap is not threadsafe, all concurrent access must be synchronized by the user" but I don't think that's the right way forward given the circumstances.

My bet is the same as yours, that it's extremely unusual to access it concurrently, so I don't think we need to benchmark it.  But I do think that we should protect against it.  Locking is pretty cheap when there's no concurrent access, so I don't think it should be that painful to just lock it.

finagle-base-http/src/main/scala/com/twitter/finagle/http/headers/JTreeMapBackedHeaderMap.scala
```diff
@@ -0,0 +1,139 @@
...
+  }
+
+  // ---- Map/MapLike -----
+
+  def -=(key: String): this.type = removed(key)
+  def +=(kv: (String, String)): this.type = set(kv._1, kv._2)
+
+  /**
+   * Underlying headers eagerly copied to an array, without synchronizing
+   * on the underlying collection.
```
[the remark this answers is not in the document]
martijnhoekstra: I added the note -- I also think that re-entrance is pretty cheap, but I wasn't entirely sure, and it's called only a handful of times, and only `private[this]` I'm happy to change it if you see those trade-offs differently.

finagle-base-http/src/main/scala/com/twitter/finagle/http/headers/JTreeMapBackedHeaderMap.scala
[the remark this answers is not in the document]
martijnhoekstra: Jikes! I didn't realize the concurrent case could lead to non-termination. That makes not synchronizing a non-starter, and the below just an aside for academic curiosity. 

The concern about concurrent iteration and modification (specifically those two) is that the concept of iteration of all headers and concurrently modifying those headers makes no sense -- either you want those in or out, but you always care, or you wouldn't iteratate them. You must synchronise externally for the operation to be meaningful and correct.

That's not always the case for two concurrent modifications that will never modify the same key, or concurrently getting a key, and modifying some keys that you know to be different to the keys you're concurrently getting.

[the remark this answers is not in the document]
bryce-anderson: I've run into this non-termination before... It was a nightmare, especially considering it was happening on a Netty thread.

Normally I'd agree that we'd prefer a `ConcurrentModificationException` and we've tried that in the past but many users are doing the wrong thing in benign places (think reusing a request to do warmup requests or something) and it just wasn't worth the hassle to introduce the behavior change.

I think the solution you've already committed of simply synchronizing the `foreach` call is fine since you're right: this shouldn't be happening very often, if at all. 🤞

mosesn:
@martijnhoekstra the code for this was merged in, but we cut it back over to the old impl for now, so that we can do some performance tests before switching over entirely.  I hope you're partially unblocked for now

mosesn:
@martijnhoekstra we found in someone else's test suite a use case where someone calls something like `headerMap.foreach { case (k, v) => headerMap.add(k, ....) }` which of course triggers a ConcurrentModificationException . . . so we're going to have to copy the iterator anyway.  :/  Sorry about that, we'll be merging it in soon though!

martijnhoekstra:
https://xkcd.com/1172/

enbnt:
This branch was merged https://github.com/twitter/finagle/commit/193c6039db9fb2e2b907a394d1126f951666b4ff#diff-9cea8690cac77728b4a871d2c3eb222c. Just circling back now to close out the PR.

Thank you, @martijnhoekstra!

Issue #757, opened 2019-02-06
Unable to stop Netty 4 Timer


DanielCharczynski:
Hello we are using Finagle as a HttpClient - it works really good. Great job.

Unfortunately Finagle generates resource leak under standard behavior that applications are deployed and undeployed many  times on JBoss AS, Tomcat etc. 
By resource leak I mean that if I 100 times deploy and undeploy an application that uses Finagle I'll get at least 100 threads

The problem is that finagle contains unstopable threads

eg.

```
private[netty4] class Netty4HashedWheelTimer
    extends Netty4Timer(HashedWheelTimer.instance)
    with ServiceLoadedTimer {

  private[this] val log = Logger.get()

  // This timer is "unstoppable".
  override def stop(): Unit =
    log.warning(
      s"Ignoring call to `Timer.stop()` on an unstoppable Netty4Timer.\n" +
        s"Current stack trace: ${Thread.currentThread.getStackTrace.mkString("\n")}"
    )
}
```

kevinoliver:
Ah. I suspect there may be other issues like this as haven't designed for that type of deployment model.

This might require a patch to create a flag that allows you to opt-out of this behavior.

DanielCharczynski:
Thanks for quick reply Kevin.
I was able to close many other threads because API allows me to use my own Timers/Executors/etc
I have problem with last three threads.

In this case private object is used and I have no access to it.

mpnewcomb:
HashedWheelTimer... why are you using this? It is a cpu hog. Start up linkerd and namerd and they just sit their chewing anywhere from 1-3% cpu when nothing is happening...

kevinoliver:
@mpnewcomb do you have any suggestions on how to improve it? if there is no work to do, it should be a thread that sleeps for a few milliseconds, looks for work, then repeats.

drywet:
Hello. This seems to work for me:
Create a class in package `com.twitter.finagle.netty4` to access private properties and stop com.twitter.finagle.util.HashedWheelTimer.nettyHwt
and
HashedWheelTimer.instance
timers

tigerlily-he:
This feature to stop a Netty 4 Timer was added in https://github.com/twitter/finagle/commit/fa4899e1ba1e9b612dbf7428b0f457ae3855b5b8. Closing this issue now that it's resolved.

PR #806, opened 2019-10-04
Add possibility to stop Netty 4 Timer

DanielCharczynski:
Problem

No possibility to stop "Netty 4 Timer".
Detailed description is available here: https://github.com/twitter/finagle/issues/757

Solution

Make HashedWheelTimer object public with additional method to stop "Netty 4 Timer"

Result

Possibility to stop "Netty 4 Timer"

### DIFF
--- a/finagle-netty4/src/main/scala/com/twitter/finagle/netty4/Netty4HashedWheelTimer.scala
+++ b/finagle-netty4/src/main/scala/com/twitter/finagle/netty4/Netty4HashedWheelTimer.scala
@@ -24,7 +24,7 @@ private object timerTickDuration
 /**
  * A Netty timer for use with [[Netty4HashedWheelTimer]].
  */
-private class HashedWheelTimer(
+class HashedWheelTimer(
   statsReceiver: StatsReceiver,
   tickDuration: Duration,
   ticksPerWheel: Int)
@@ -67,7 +67,7 @@ private class HashedWheelTimer(
   self.newTimeout(pendingTasksStat, statsPollInterval.inSeconds, TimeUnit.SECONDS)
 }
 
-private object HashedWheelTimer {
+object HashedWheelTimer {
 
   /**
    * A singleton instance of [[HashedWheelTimer]] that is used for the all service loaded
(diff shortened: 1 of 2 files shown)


finagle-netty4/src/main/scala/com/twitter/finagle/netty4/package.scala
```diff
@@ -15,4 +15,10 @@ package object netty4 {
    */
   private[finagle] val Toggles: ToggleMap =
     StandardToggleMap("com.twitter.finagle.netty4", DefaultStatsReceiver)
+	
+  /** Stops HashedWheelTimer thread */
+  def stopHashedWheelTimer {
```
mosesn: how about we instead make the HashedWheelTimer object public and add a public method `stop` to it?  I think that will be a cleaner API.

finagle-netty4/src/main/scala/com/twitter/finagle/netty4/Netty4HashedWheelTimer.scala
```diff
@@ -24,7 +24,7 @@ private object timerTickDuration
 /**
  * A Netty timer for use with [[Netty4HashedWheelTimer]].
  */
-private class HashedWheelTimer(
+class HashedWheelTimer(
```
mosesn: can we do it while keeping this private?  or does that cause a problem?  if we can't make this private, can we make the constructor private?

finagle-netty4/src/main/scala/com/twitter/finagle/netty4/Netty4HashedWheelTimer.scala
```diff
@@ -67,7 +67,7 @@ private class HashedWheelTimer(
   self.newTimeout(pendingTasksStat, statsPollInterval.inSeconds, TimeUnit.SECONDS)
 }
 
-private object HashedWheelTimer {
+object HashedWheelTimer {
```
mosesn: thinking about it more, I think that we should probably make `instance` private, and just have a stop method on this object.  that way people can't pass around HashedWheelTimer willy-nilly and accidentally get confused and stop it.  you have to know that you're stopping the default one.

DanielCharczynski: Yes, that was one of my ideas. Done

mosesn:
Looks great, thanks!

jyanJing:
Thank you for working on this @DanielCharczynski ! The change is merged in [fa4899e](https://github.com/twitter/finagle/commit/fa4899e1ba1e9b612dbf7428b0f457ae3855b5b8)

Issue #756, opened 2019-02-06
Unable to stop HighResTimer


DanielCharczynski:
Hello we are using Finagle as a HttpClient - it works really good. Great job.

Unfortunately Finagle generates resource leak under standard behavior that applications are deployed and undeployed many  times on JBoss AS, Tomcat etc. 
By resource leak I mean that if I 100 times deploy and undeploy an application that uses Finagle I'll get at least 100 threads

The problem is that finagle contains unstopable threads

Problem is withh **HighResTimer** Thread
There is no possibility to stop it using Finagle API.

```
object HighResTimer {

  /**
   * The default Timer used for configuration.
   *
   * It is a shared resource and as such, `stop` is ignored.
   */
  val Default: com.twitter.util.Timer =
    new JavaTimer(true, Some("HighResTimer")) {
      override def stop(): Unit = ()
    }
```

Unfortunately we can't use our own HighResTimer using params because Default is initialized when Http.client is created and there are two instances

```
 /**
   * Retries failures that are guaranteed to be safe to retry
   * (see [[RetryPolicy.RetryableWriteException]]).
   */
  private[finagle] def moduleRequeueable[Req, Rep]: Stackable[ServiceFactory[Req, Rep]] =
    new Stack.Module[ServiceFactory[Req, Rep]] {
      def role: Stack.Role = Retries.Role

      def description: String =
        "Retries requests, at the service application level, that have been rejected"

      val parameters = Seq(
        implicitly[Stack.Param[Stats]],
        implicitly[Stack.Param[Budget]],
        implicitly[Stack.Param[HighResTimer]]
      )
```

There are two alternative solutions
1. possibility to stop this Thread
2. Do not instantiate HighResTimer when Client is created

kevinoliver:
Ah, yep another instance of this. Just like #757 . 

I think same sort of solution noted there could work here as well.

DanielCharczynski:
Thanks for quick reply Kevin.
I was able to close many other threads because API allows me to use my own Timers/Executors/etc
I have problem with last three threads.

In this case HighResTimer param is injected implicitly when client is instantiated.
It is definitely to early.
In such implementation  StackParam[HighResTimer] makes no sense

PR #808, opened 2019-10-07
make HighResTimer public

DanielCharczynski:
Problem

No possibility to stop HighResTimer.
Detailed description is available here: #756

Solution

Make HighResTimer stoppable

Result

Possibility to stop HighResTimer instance

### DIFF
--- a/finagle-core/src/main/scala/com/twitter/finagle/param/Params.scala
+++ b/finagle-core/src/main/scala/com/twitter/finagle/param/Params.scala
@@ -124,21 +124,28 @@ case class HighResTimer(timer: com.twitter.util.Timer) {
 object HighResTimer {
 
   /**
-   * The default Timer used for configuration.
-   *
-   * It is a shared resource and as such, `stop` is ignored.
+   * The underlying default Timer used for configuration
+   * Default `stop` behaviour is ignored in order to not be accidentally stopped
    */
-  val Default = new JavaTimer(true, Some("HighResTimer")) {
+  private[this] val underlying = new JavaTimer(true, Some("HighResTimer")) {
     override def stop(): Unit = ()
     def stopTimer(): Unit = super.stop()
   }
 
+  /**
+   * The default Timer used for configuration.
+   *
+   * It is a shared resource and as such, `stop` is ignored.
+   */
+  val Default:com.twitter.util.Timer = underlying
 
   /**
    * Stop default HighResTimer
    */
-  def stop() {
-    Default.stopTimer()
+  def stop():Unit = {
+    util.DefaultLogger.warning("Stopping the default Finagle HighResTimer. When timer is stopped, " +
+        "the behaviors of Finagle client and server are undefined.")
+    underlying.stopTimer()
   }
   
   implicit val param: Stack.Param[HighResTimer] =
--- a/finagle-core/src/main/scala/com/twitter/finagle/param/Params.scala
+++ b/finagle-core/src/main/scala/com/twitter/finagle/param/Params.scala
@@ -137,13 +137,13 @@ object HighResTimer {
    *
    * It is a shared resource and as such, `stop` is ignored.
    */
-  val Default:com.twitter.util.Timer = underlying
+  val Default: com.twitter.util.Timer = underlying
 
   /**
    * Stop default HighResTimer
    */
-  def stop():Unit = {
-    util.DefaultLogger.warning("Stopping the default Finagle HighResTimer. When timer is stopped, " +
+  def stop(): Unit = {
+    com.twitter.logging.Logger.get().warning("Stopping the default Finagle HighResTimer. When timer is stopped, " +
         "the behaviors of Finagle client and server are undefined.")
     underlying.stopTimer()
   }
--- a/finagle-core/src/main/scala/com/twitter/finagle/param/Params.scala
+++ b/finagle-core/src/main/scala/com/twitter/finagle/param/Params.scala
@@ -130,7 +130,7 @@ object HighResTimer {
    */
   val Default: com.twitter.util.Timer =
     new JavaTimer(true, Some("HighResTimer")) {
-      override def stop(): Unit = ()
+
     }
 
   implicit val param: Stack.Param[HighResTimer] =
--- a/finagle-core/src/main/scala/com/twitter/finagle/param/Params.scala
+++ b/finagle-core/src/main/scala/com/twitter/finagle/param/Params.scala
@@ -128,14 +128,17 @@ object HighResTimer {
    *
    * It is a shared resource and as such, `stop` is ignored.
    */
-  private[finagle] val Default: com.twitter.util.Timer =
-    new JavaTimer(true, Some("HighResTimer"))
+  val Default = new JavaTimer(true, Some("HighResTimer")) {
+    override def stop(): Unit = ()
+    def stopTimer(): Unit = super.stop()
+  }
+
 
   /**
    * Stop default HighResTimer
    */
   def stop() {
-    Default.stop()
+    Default.stopTimer()
   }
   
   implicit val param: Stack.Param[HighResTimer] =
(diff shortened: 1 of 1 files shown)


finagle-core/src/main/scala/com/twitter/finagle/param/Params.scala
```diff
@@ -130,7 +130,7 @@ object HighResTimer {
    */
   val Default: com.twitter.util.Timer =
     new JavaTimer(true, Some("HighResTimer")) {
```
mosesn: if we're making it stoppable, we should also remove the { } from the construction

DanielCharczynski: Let's think about it one more time.
How can I stop this timer accidentally using dedicated  method `stop`.?
In my opinion people expect such behavior. If I call stop this timer should be stopped.
I thing that previous behavior  was quite unexpected.

finagle-core/src/main/scala/com/twitter/finagle/param/Params.scala
```diff
@@ -128,9 +128,16 @@ object HighResTimer {
    *
    * It is a shared resource and as such, `stop` is ignored.
    */
-  val Default: com.twitter.util.Timer =
+  private[finagle] val Default: com.twitter.util.Timer =
```
mosesn: I don't think we can reduce the visibility on this Timer, people use it all the time.

finagle-core/src/main/scala/com/twitter/finagle/param/Params.scala
```diff
@@ -128,9 +128,16 @@ object HighResTimer {
...
     new JavaTimer(true, Some("HighResTimer"))
```
mosesn: we should keep stop hidden, but we can make another method `private[HighResTimer] def hiddenStop(): Unit = super.stop()`, and then have stop use that method

finagle-core/src/main/scala/com/twitter/finagle/param/Params.scala
```diff
@@ -128,9 +128,16 @@ object HighResTimer {
...
 
+  /**
+   * Stop default HighResTimer
+   */
+  def stop() {
```
mosesn: procedure-style syntax is being deprecated, please make it explicit:
```
def stop(): Unit = {...
```

finagle-core/src/main/scala/com/twitter/finagle/param/Params.scala
DanielCharczynski: Such method will not be visible because it is not defined in com.twitter.util.Timer.
I have to create new class that extends com.twitter.util.Timer.
Do you still want to do it in that way?

finagle-core/src/main/scala/com/twitter/finagle/param/Params.scala
DanielCharczynski: OK

finagle-core/src/main/scala/com/twitter/finagle/param/Params.scala
DanielCharczynski: OK

finagle-core/src/main/scala/com/twitter/finagle/param/Params.scala
```diff
@@ -128,14 +128,17 @@ object HighResTimer {
    *
    * It is a shared resource and as such, `stop` is ignored.
    */
-  private[finagle] val Default: com.twitter.util.Timer =
-    new JavaTimer(true, Some("HighResTimer"))
+  val Default = new JavaTimer(true, Some("HighResTimer")) {
+    override def stop(): Unit = ()
+    def stopTimer(): Unit = super.stop()
```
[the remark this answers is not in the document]
DanielCharczynski: it won't compile
```
[info] Compiling 1 Scala source to C:\Users\GQ28BL\Projects\work\ing\arch\okapi\finagle\finagle\finagle-core\target\scala-2.12\classes ...
[error] C:\Users\GQ28BL\Projects\work\ing\arch\okapi\finagle\finagle\finagle-core\src\main\scala\com\twitter\finagle\param\Params.scala:141:13: value stopTimer is not a member of com.twitter.util.JavaTimer
[error]     Default.stopTimer()
[error]             ^
[error] one error found
[error] (finagle-core / Compile / compileIncremental) Compilation failed
[error] Total time: 22 s, completed 2019-10-10 10:13:17
```

finagle-core/src/main/scala/com/twitter/finagle/param/Params.scala
mosesn: I think this actually hasn't been fixed yet.

finagle-core/src/main/scala/com/twitter/finagle/param/Params.scala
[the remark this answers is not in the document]
mosesn: ok, well this is a bit convoluted, but we can make it work with

```
private[this] val underlying = new JavaTimer... {
  ...
  def stopTimer...
}

val Default: JavaTimer = underlying

def stop(): Unit = underlying.stopTimer
```

and let's add some kind of note that we're doing this because we don't want it to be accidentally stoppable.

finagle-core/src/main/scala/com/twitter/finagle/param/Params.scala
```diff
@@ -128,11 +128,19 @@ object HighResTimer {
...
+  }
+
 
+  /**
+   * Stop default HighResTimer
+   */
+  def stop() {
+    Default.stopTimer()
```
mosesn: can we also add a log message warning users that this might break Finagle?  we did something similar when we pulled in your other patch:

https://github.com/twitter/finagle/commit/fa4899e1ba1e9b612dbf7428b0f457ae3855b5b8

finagle-core/src/main/scala/com/twitter/finagle/param/Params.scala
DanielCharczynski: Done

finagle-core/src/main/scala/com/twitter/finagle/param/Params.scala
DanielCharczynski: Done

finagle-core/src/main/scala/com/twitter/finagle/param/Params.scala
```diff
@@ -123,16 +123,31 @@ case class HighResTimer(timer: com.twitter.util.Timer) {
...
   /**
    * The default Timer used for configuration.
    *
    * It is a shared resource and as such, `stop` is ignored.
    */
-  val Default: com.twitter.util.Timer =
-    new JavaTimer(true, Some("HighResTimer")) {
-      override def stop(): Unit = ()
-    }
+  val Default:com.twitter.util.Timer = underlying
```
mosesn: style: should be `val Default: com.twitter.util.Timer = underlying`

finagle-core/src/main/scala/com/twitter/finagle/param/Params.scala
```diff
@@ -123,16 +123,31 @@ case class HighResTimer(timer: com.twitter.util.Timer) {
...
 
+  /**
+   * Stop default HighResTimer
+   */
+  def stop():Unit = {
```
mosesn: style: should be `def stop(): Unit = {`

finagle-core/src/main/scala/com/twitter/finagle/param/Params.scala
```diff
@@ -123,16 +123,31 @@ case class HighResTimer(timer: com.twitter.util.Timer) {
...
+    util.DefaultLogger.warning("Stopping the default Finagle HighResTimer. When timer is stopped, " +
```
mosesn: instead of util.DefaultLogger, let's use a similar Logger.get() approach to what we did in the other review.

mosesn:
@DanielCharczynski thanks for the ping, we're working on it now!

vkostyukov:
@DanielCharczynski We pulled this in for internal review. Will ping you when it's merged.

mosesn:
This was merged in a few days ago, thanks for the PR!  https://github.com/twitter/finagle/commit/5301aba8ef39dc283b7ee1b511cb27bb27a6dfaa

PR #809, opened 2019-10-11
punctuation error corrected

AANI007:
Problem

Solution

Result

the solution.


jyanJing:
Some process added a comma in README.md, but I think it is OK without it. Not sure if it worth pulling the change, closing it unless someone feels strongly about changing it.

PR #817, opened 2019-12-02
Fix typos

bwignall:
Should be non-semantic.

Uses https://en.wikipedia.org/wiki/Wikipedia:Lists_of_common_misspellings/For_machines to find likely typos.

### DIFF
--- a/doc/src/sphinx/FAQ.rst
+++ b/doc/src/sphinx/FAQ.rst
@@ -63,7 +63,7 @@ a :API:`ClientDiscardedRequestException <com/twitter/finagle/mux/ClientDiscarded
 is the same as above though.
 
 Server operators may prefer to measure their server side success rates excluding these classes of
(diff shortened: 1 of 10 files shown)


mosesn (approved):
Thanks @bwignall we'll pull this in soon!

mosesn:
Actually, do you mind signing our CLA?  @bwignall

bwignall:
@mosesn , do you need anything else from me?

ryanoneill:
@bwignall, we're good. I'll pull this in this morning. 

The way this works is that we take OSS commits and merge them in internally, and then they get pushed back out to GitHub. I'll let you know when the commit has made it's way back to GitHub. Thanks for the submission.

ryanoneill:
This has been merged and internally and pushed back to GitHub as commit https://github.com/twitter/finagle/commit/90efe95ddcc9d3eddc6a1aa9f79154e2d3131e72

Thanks again.

PR #821, opened 2020-01-02
finagle-stats: -> 2.13; finagle-stats-core: -> 2.13

hderms:
Problem

No cross-build for finagle-stats 2.13 and finagle-stats-core 2.13

Solution

include 2.13 in finagle-stats and finagle-stats-core cross versions.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -457,7 +457,8 @@ lazy val finagleStatsCore = Project(
   id = "finagle-stats-core",
   base = file("finagle-stats-core")
 ).settings(
-  sharedSettings
+  sharedSettings,
+  withTwoThirteen
 ).settings(
   name := "finagle-stats-core",
   libraryDependencies ++= Seq(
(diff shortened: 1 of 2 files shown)


yufangong:
I think there are a few places in code that need change to compatible with scala 2.13.
Could you compile this against 2.13 to verify?

hderms:
@yufangong Sorry about that I apparently didn't know how to trigger cross-compilation

hderms:
@yufangong just to make sure I'm doing this properly, if it compiles with `sbt +compile` after having added 2.13 as a cross-compilation target (which this PR already does), then we are good to go from a compilation perspective?

yufangong:
@hderms I think so! Maybe "./sbt +test:compile" just make sure all tests also good. Thanks!

yufangong:
this is merged https://github.com/twitter/finagle/commit/4144d73c9e9f997e219e90888cce108861dc6046. Thanks!!!

PR #825, opened 2020-01-15
finagle-memcached: update bijection-core to 0.9.7

martijnhoekstra:
Problem

bijection-core version is out of date, and there is no 2.13 release for the current version.

Solution

Update bijection-core version to 0.9.7

Result

Bijection-core dependency can be cross-built for 2.13

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -674,7 +674,7 @@ lazy val finagleMemcached = Project(
   libraryDependencies ++= Seq(
     util("hashing"),
     util("zk-test") % "test",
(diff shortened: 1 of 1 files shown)


martijnhoekstra:
This probably needs to be done monorepo wide.

yufangong:
Hey @martijnhoekstra, we rushed in this change here https://github.com/twitter/finagle/commit/de0ec2c6714ae2c4d8bf76f72cccb9e35f3325da, to have it before the release, so I didn't pull in your change, but we did it monorepo wide, and it seems working. Thank you for pointing the direction.

PR #827, opened 2020-01-29
finagle-memcached: -> 213

martijnhoekstra:
Problem

no 2.13 for finagle-memcached

Solution

upgrade. Replace javaClientBase.java with equivalent .scala so that the type aliases of collection compat scala.jdk.CollectionConverters._ can be picked up.

Result

a 2.13 build


martijnhoekstra:
I get the same errors with this branch as I get on develop on Windows. I haven't tested on a different platform if these errors are real, I'm assuming they're not.

martijnhoekstra:
Throw in finagle-benchmark as it's just a build change.

felixbr:
I looked at the travis-ci errors and all failed build jobs seem to fail during download of `sbt-launch.jar`

```
downloading /home/travis/build/twitter/finagle/sbt-launch.jar

The command "./sbt --error ++2.12.8 ci:update finagle-memcached/update" failed 3 times.
```

I cannot see any particular pattern regarding jdk or scala version and the sbt version should always be `1.2.8` if I see this correctly.

martijnhoekstra:
Yeah, the sbt download seems flaky. This is not the first time tests fail on that.

martijnhoekstra:
For the breakout ones, we can go through an Iterator, which should avoid intermediary collection construction.

For the mapValues ones, we can use `transform` instead, but there is still a potential performance hazard there. That should be fixed when https://github.com/scala/scala/pull/8630 lands in 2.12, but still may be slow in 2.11.

yufangong:
This looks good to me except Bryce's comments. I did source integration tests and this change passed.  I think once we have util, scrooge, finagle, finatra, and twitter-server all cross-building 2.13 we will drop 2.11 support, so I won't be too worry about 2.11 perf.

finagle-memcached/src/main/scala/com/twitter/finagle/memcached/Client.scala
```diff
@@ -637,12 +638,12 @@ protected class ConnectedClient(protected val service: Service[Command, Response
   import ClientConstants._
 
   protected def rawGet(command: RetrievalCommand): Future[GetResult] = {
-    val keys: immutable.Set[String] = command.keys.map { case Buf.Utf8(s) => s }(breakOut)
+    val keys: immutable.Set[String] = command.keys.iterator.map { case Buf.Utf8(s) => s }.toSet
 
-    service(command).transform {
+    service(command).transform{ case x => {println(s"transformed $x"); x match {
```
yufangong: assuming this is for debugging, we can remove it when pulling in.

martijnhoekstra: How did I miss that...

Yes, that needs to be removed

finagle-memcached/src/main/scala/com/twitter/finagle/memcached/Client.scala
```diff
@@ -822,7 +823,7 @@ protected class ConnectedClient(protected val service: Service[Command, Response
   def stats(args: Option[String]): Future[Seq[String]] = {
     val statArgs: Seq[Buf] = args match {
       case None => Seq(Buf.Empty)
-      case Some(args) => args.split(" ").map(nonEmptyStringToBuf)(breakOut)
+      case Some(args) => args.split(" ").map(nonEmptyStringToBuf).toSeq
```
bryce-anderson: Could this also benefit from a `.iterator` call?

finagle-memcached/src/main/scala/com/twitter/finagle/memcached/TwemcacheCacheClient.scala
```diff
@@ -59,7 +58,7 @@ trait TwemcacheConnectedClient extends TwemcacheClient { self: ConnectedClient =
   def getvResult(keys: Iterable[String]): Future[GetsResult] = {
     try {
       if (keys == null) throw new IllegalArgumentException("Invalid keys: keys cannot be null")
-      val bufs = keys.map { Buf.Utf8(_) }(breakOut)
+      val bufs = keys.map { Buf.Utf8(_) }.toSeq
```
bryce-anderson: Does this need a `.iterator` call?

finagle-memcached/src/main/scala/com/twitter/finagle/memcached/TwemcacheCacheClient.scala
```diff
@@ -148,7 +147,7 @@ trait TwemcachePartitionedClient extends TwemcacheClient { self: PartitionedClie
   )(f: (TwemcacheClient, Iterable[String]) => Future[A]
   ): Future[Seq[A]] = {
     Future.collect(
-      keys.groupBy(twemcacheClientOf).map(Function.tupled(f))(breakOut)
+      keys.groupBy(twemcacheClientOf).map(Function.tupled(f)).toSeq
```
bryce-anderson: `.iterator` before the `.map`?

finagle-memcached/src/main/scala/com/twitter/finagle/memcached/util/Bufs.scala
```diff
@@ -24,7 +23,7 @@ private[finagle] object Bufs {
     if (strings == null) {
       null
     } else {
-      strings.map(nonEmptyStringToBuf)(breakOut)
+      strings.map(nonEmptyStringToBuf).toSeq
```
bryce-anderson: `.iterator` before the `.map`?

finagle-memcached/src/main/scala/com/twitter/finagle/memcached/util/ParserUtils.scala
```diff
@@ -44,7 +44,7 @@ object ParserUtils {
         segmentStart = segmentEnd + 1
       }
     }
-    split
+    split.toIndexedSeq
```
bryce-anderson: Is this going to spit out a `Vector`? If so, that would be pretty bad.

finagle-memcached/src/main/scala/com/twitter/finagle/memcached/Client.scala
```diff
@@ -1180,7 +1181,7 @@ case class RubyMemCacheClientBuilder(
   def nodes(hostPortWeights: String): RubyMemCacheClientBuilder =
     copy(_nodes = CacheNodeGroup(hostPortWeights).members.map { node: CacheNode =>
       (node.host, node.port, node.weight)
-    }(breakOut))
+    }.toSeq)
```
bryce-anderson: I think we can also do a `.iterator` before the `.map` here.

finagle-memcached/src/main/scala/com/twitter/finagle/memcached/Client.scala
```diff
@@ -1228,7 +1229,7 @@ case class PHPMemCacheClientBuilder(
   def nodes(hostPortWeights: String): PHPMemCacheClientBuilder =
     copy(_nodes = CacheNodeGroup(hostPortWeights).members.map { node: CacheNode =>
       (node.host, node.port, node.weight)
-    }(breakOut))
+    }.toSeq)
```
bryce-anderson: I think we can also do a `.iterator` before the `.map` here.

finagle-memcached/src/main/scala/com/twitter/finagle/memcached/util/ParserUtils.scala
martijnhoekstra: Yes. I pushed an approach in a separate commit that just `toSeq`'s this and changes the return type to Seq (which then propagates forward a bit). I don't see any immediate performance hazards for that.

hamdiallam:
Merged! 7acd423f88312a2f9dd00d38a06ae7ef3e51faf7

PR #840, opened 2020-04-06
finagle-mysql: Add connection init sql stack

krrrr38:
## Problem

Sometimes we want to execute settings query after create connection.
Currently there is no api to add connection-init-sql.

## Solution

Add connection-init-request api into mysql client. this is based on HikariCP `connectionInitSql`.

ref: https://github.com/brettwooldridge/HikariCP/blob/dev/README.md#infrequently-used
image

----

This feature is possible to implement in user applications using stack.
If you don't want to add this feature into finagle-mysql, please let me know and close this PR 🙏

### DIFF
--- a/finagle-mysql/src/main/scala/com/twitter/finagle/mysql/ConnectionInitSql.scala
+++ b/finagle-mysql/src/main/scala/com/twitter/finagle/mysql/ConnectionInitSql.scala
@@ -1,0 +1,33 @@
+package com.twitter.finagle.mysql
+
+import com.twitter.finagle._
+import com.twitter.finagle.mysql.param.ConnectionInitRequest
+import com.twitter.util._
+
+/**
+  * A stack module that executes a request when create a connection.
+  */
+object ConnectionInitSql {
+  val Role: Stack.Role = Stack.Role("ConnectionInitSql")
+
+  private[finagle] def module: Stackable[ServiceFactory[Request, Result]] =
+    new Stack.Module1[ConnectionInitRequest, ServiceFactory[Request, Result]] {
+      val role: Stack.Role = Role
+      val description: String = "Installs a connection init sql module in the stack"
+
+      def make(in: ConnectionInitRequest, next: ServiceFactory[Request, Result]): ServiceFactory[Request, Result] = {
+        in.request match {
+          case Some(req) =>
+            // we assume that this module added below the connection pool
+            next.flatMap { service =>
+              service(req).flatMap {
+                case Error(code, sqlState, message) =>
+                  Future.exception(ServerError(code, sqlState, message))
+                case _ => Future.value(service)
+              }
+            }
+          case None => next
+        }
+      }
+    }
+}
--- a/finagle-mysql/src/test/scala/com/twitter/finagle/mysql/integration/MysqlBuilderTest.scala
+++ b/finagle-mysql/src/test/scala/com/twitter/finagle/mysql/integration/MysqlBuilderTest.scala
@@ -2,6 +2,7 @@ package com.twitter.finagle.mysql.integration
 
 import com.twitter.conversions.DurationOps._
 import com.twitter.finagle.Mysql
+import com.twitter.finagle.mysql.QueryRequest
 import com.twitter.finagle.param
 import com.twitter.finagle.tracing._
 import com.twitter.util.{Await, Awaitable}
@@ -32,6 +33,7 @@ class MysqlBuilderTest extends FunSuite with IntegrationClient {
         .withDatabase("test")
         .withCredentials(username, password)
         .withDatabase(db)
+        .withConnectionInitRequest(QueryRequest("SET SESSION sql_mode='TRADITIONAL,NO_AUTO_VALUE_ON_ZERO,ONLY_FULL_GROUP_BY'"))
         .newRichClient("localhost:3306")
 
       ready(client.query("SELECT 1"))
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,6 +7,11 @@ Note that ``PHAB_ID=#`` and ``RB_ID=#`` correspond to associated messages in com
 Unreleased
 ----------
 
+New Features
+~~~~~~~~~~~~
+
+* finagle-mysql: Add `ConnectionInitSql` stack to set connection init sql.
+
 Runtime Behavior Changes
 ~~~~~~~~~~~~~~~~~~~~~~~~
 
--- a/finagle-mysql/src/main/scala/com/twitter/finagle/Mysql.scala
+++ b/finagle-mysql/src/main/scala/com/twitter/finagle/Mysql.scala
@@ -117,6 +117,7 @@ object Mysql extends com.twitter.finagle.Client[Request, Result] with MysqlRichC
       // Note: there is a stack overflow in insertAfter using CanStackFrom, thus the module.
       .insertAfter(DefaultPool.Role, PoisonConnection.module)
       .prepend(RollbackFactory.module)
+      .insertAfter(StackClient.Role.prepConn, ConnectionInitSql.module)
   }
 
   /**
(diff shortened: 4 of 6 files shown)


finagle-mysql/src/test/scala/com/twitter/finagle/mysql/integration/MysqlBuilderTest.scala
```diff
@@ -32,6 +33,7 @@ class MysqlBuilderTest extends FunSuite with IntegrationClient {
         .withDatabase("test")
         .withCredentials(username, password)
         .withDatabase(db)
+        .withConnectionInitSql(QueryRequest("SET SESSION sql_mode='TRADITIONAL,NO_AUTO_VALUE_ON_ZERO,ONLY_FULL_GROUP_BY'"))
         .newRichClient("localhost:3306")
```
krrrr38, line 37: In this case, general-log become followings.

```
2020-04-06T10:16:40.042051Z         4 Connect   root@172.17.0.1 on test using TCP/IP
2020-04-06T10:16:40.084218Z         4 Query     SET SESSION sql_mode='TRADITIONAL,NO_AUTO_VALUE_ON_ZERO,ONLY_FULL_GROUP_BY'
2020-04-06T10:16:40.108476Z         4 Query     SELECT 1
2020-04-06T10:16:40.138291Z         4 Query     ROLLBACK
2020-04-06T10:16:40.176864Z         4 Prepare   SELECT ?
2020-04-06T10:16:40.191175Z         4 Execute   SELECT 1
2020-04-06T10:16:40.197046Z         4 Query     ROLLBACK
2020-04-06T10:16:40.206384Z         4 Query     ROLLBACK
```

finagle-mysql/src/main/scala/com/twitter/finagle/mysql/ConnectionInitSql.scala
```diff
@@ -0,0 +1,30 @@
...
+        next.flatMap { service =>
+          in.request.fold(Future.value(service)) { req =>
+            service(req).flatMap {
+              case _: OK => Future.value(service)
+              case Error(code, sqlState, message) =>
+                Future.exception(ServerError(code, sqlState, message))
+            }
+          }
+        }
+      }
```
bryce-anderson, line 28: I think we can effectively strip out the ServiceFactory wrapper if we match on `in` before the next.flatMap call.

finagle-mysql/src/main/scala/com/twitter/finagle/mysql/ConnectionInitSql.scala
```diff
@@ -0,0 +1,30 @@
...
+    new Stack.Module1[ConnectionInitRequest, ServiceFactory[Request, Result]] {
+      val role: Stack.Role = Role
+      val description: String = "Installs a connection init sql module in the stack"
+
+      def make(in: ConnectionInitRequest, next: ServiceFactory[Request, Result]): ServiceFactory[Request, Result] = {
+        next.flatMap { service =>
+          in.request.fold(Future.value(service)) { req =>
+            service(req).flatMap {
+              case _: OK => Future.value(service)
+              case Error(code, sqlState, message) =>
```
bryce-anderson, line 23: In the error case I believe we need to make sure to close the `service`.

bryce-anderson:
Could we also add a CHANGELOG entry? Thanks again @krrrr38!

finagle-mysql/src/main/scala/com/twitter/finagle/mysql/ConnectionInitSql.scala
krrrr38: changed to check `in.request` firstly.
https://github.com/twitter/finagle/pull/840/commits/c0bc82e6504441cd5cc3db80ae1d4075fccc24cc

finagle-mysql/src/main/scala/com/twitter/finagle/mysql/ConnectionInitSql.scala
krrrr38: add service.close in error case.
https://github.com/twitter/finagle/pull/840/commits/c0bc82e6504441cd5cc3db80ae1d4075fccc24cc

krrrr38:
@bryce-anderson thank you for your review. I added comment into CHANGELOG.rst in https://github.com/twitter/finagle/pull/840/commits/c0bc82e6504441cd5cc3db80ae1d4075fccc24cc

bryce-anderson (approved):
This looks fine to me but I'm not 100% sure that this shouldn't be part of the handshake operations. This should work fine where it is: I just want to make sure nobody thinks it belongs with the rest of the handshake logic.

ryanoneill:
I like where this is. It's usable yet doesn't further complicate the logic contained in the handshake or dispatcher. I think that's ideal in this scenario.

ryanoneill (approved):

finagle-mysql/src/main/scala/com/twitter/finagle/mysql/ConnectionInitSql.scala
```diff
@@ -0,0 +1,35 @@
...
+        in.request match {
+          case Some(req) =>
+            next.flatMap { service =>
+              service(req).flatMap {
+                case _: OK => Future.value(service)
+                case Error(code, sqlState, message) =>
+                  service.close().flatMap { _ =>
+                    Future.exception(ServerError(code, sqlState, message))
+                  }
+                case r => Future.exception(new IllegalStateException(s"Unsupported response to a modify='$r'"))
```
roanta, line 28: Not sure I would use an IllegalStateException here as
this is more like an unsupported response. Maybe just
use a generic exception instead?

finagle-mysql/src/main/scala/com/twitter/finagle/mysql/ConnectionInitSql.scala
```diff
@@ -0,0 +1,35 @@
...
+
+  private[finagle] def module: Stackable[ServiceFactory[Request, Result]] =
+    new Stack.Module1[ConnectionInitRequest, ServiceFactory[Request, Result]] {
+      val role: Stack.Role = Role
+      val description: String = "Installs a connection init sql module in the stack"
+
+      def make(in: ConnectionInitRequest, next: ServiceFactory[Request, Result]): ServiceFactory[Request, Result] = {
+        in.request match {
+          case Some(req) =>
+            next.flatMap { service =>
```
roanta, line 21: I would add a comment here that this assumes the module is added below the connection pool (otherwise, we would need to return it to the pool on the happy path too).

roanta (approved):
This looks good to me, only a few minor nits.

Thanks @krrrr38!

finagle-mysql/src/main/scala/com/twitter/finagle/mysql/ConnectionInitSql.scala
krrrr38: Currently ConnectionInitRequst accepts all kind of request, so sometimes mysql return not only `OK` but also `ResultSet`.
So I changed response handling which process Error and otherwise in https://github.com/twitter/finagle/pull/840/commits/d6532f8c6498bb8c871774180b586dfcaaa617b0.

finagle-mysql/src/main/scala/com/twitter/finagle/mysql/ConnectionInitSql.scala
krrrr38: add a comment https://github.com/twitter/finagle/pull/840/commits/d6532f8c6498bb8c871774180b586dfcaaa617b0

finagle-mysql/src/main/scala/com/twitter/finagle/mysql/ConnectionInitSql.scala
bryce-anderson: I was wrong: the `flatMap` operation will close the underlying service if we get a `Future.exception` out of it. I'll change it back in the internal process.

bryce-anderson:
Closed via bd4411ddc5e6e0df6b08b78a507280b654e617b9. Thanks!

PR #841, opened 2020-04-07
finagle-redis: Add connection init commands for select and auth

krrrr38:
## Problem

There is no api to set select database and auth password in redis client.
These commands are enough to be executed once in a connection.

## Solution

Add ConnectionInitCommand stack and client api to set database and password.

### DIFF
--- a/finagle-redis/src/main/scala/com/twitter/finagle/redis/exp/ConnectionInitCommand.scala
+++ b/finagle-redis/src/main/scala/com/twitter/finagle/redis/exp/ConnectionInitCommand.scala
@@ -1,0 +1,42 @@
+package com.twitter.finagle.redis.exp
+
+import com.twitter.finagle._
+import com.twitter.finagle.redis.ServerError
+import com.twitter.finagle.redis.param.{Database, Password}
+import com.twitter.finagle.redis.protocol._
+import com.twitter.util.Future
+
+/**
+ * A stack module that executes commands when create a new connection.
+ */
+object ConnectionInitCommand {
+
+  val Role = Stack.Role("RedisInitialCommand")
+
+  def module: Stackable[ServiceFactory[Command, Reply]] =
+    new Stack.Module2[Database, Password, ServiceFactory[Command, Reply]] {
+      val role = Role
+      val description = "Manage redis connections"
+
+      def make(db: Database, password: Password, next: ServiceFactory[Command, Reply]): ServiceFactory[Command, Reply] = {
+        for {
+          service <- next
+          authed <- apply(service, password.code, Auth.apply)
+          selected <- apply(authed, db.index, Select.apply)
+        } yield selected
+      }
+    }
+
+  private def apply[A](service: Service[Command, Reply], param: Option[A], command: A => Command): Future[Service[Command, Reply]] = {
+    param match {
+      case Some(p) =>
+        service(command(p)).flatMap {
+          case StatusReply(_) => Future.value(service)
+          case ErrorReply(msg) =>
+            Future.exception(ServerError(msg))
+          case r => Future.exception(new IllegalStateException(s"Unsupported response to a modify='$r'"))
+        }
+      case None => Future.value(service)
+    }
+  }
+}
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,6 +7,11 @@ Note that ``PHAB_ID=#`` and ``RB_ID=#`` correspond to associated messages in com
 Unreleased
 ----------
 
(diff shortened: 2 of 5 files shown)


finagle-redis/src/main/scala/com/twitter/finagle/redis/exp/ConnectionInitCommand.scala
```diff
@@ -0,0 +1,47 @@
...
+  val Role = Stack.Role("RedisInitialCommand")
+
+  def module: Stackable[ServiceFactory[Command, Reply]] =
+    new Stack.Module2[Database, Password, ServiceFactory[Command, Reply]] {
+      val role = Role
+      val description = "Manage redis connections"
+
+      def make(db: Database, password: Password, next: ServiceFactory[Command, Reply]): ServiceFactory[Command, Reply] = {
+        for {
+          service <- next
```
bryce-anderson, line 23: I think we need to make sure we close the `service` if any of the follow up queries fail.

finagle-redis/src/main/scala/com/twitter/finagle/redis/exp/ConnectionInitCommand.scala
```diff
@@ -0,0 +1,47 @@
...
+          authed <- auth(service, password)
+          selected <- select(authed, db)
+        } yield selected
+      }
+    }
+
+  private def select(service: Service[Command, Reply], db: Database): Future[Service[Command, Reply]] = {
+    db.index.fold(Future.value(service)) { index =>
```
bryce-anderson, line 31: nit: we tend to use `match` instead of a fold. Same comment for below.

finagle-redis/src/main/scala/com/twitter/finagle/redis/exp/ConnectionInitCommand.scala
```diff
@@ -0,0 +1,47 @@
...
+      service(Select(index)).flatMap {
+        case ErrorReply(msg) => Future.exception(ClientError(msg))
+        case _ => Future.value(service)
+      }
+    }
+  }
+
+  private def auth(service: Service[Command, Reply], password: Password): Future[Service[Command, Reply]] = {
```
bryce-anderson, line 39: This feels generalizable: the only difference between `auth` and `select` are the commands.

finagle-redis/src/main/scala/com/twitter/finagle/redis/exp/ConnectionInitCommand.scala
```diff
@@ -0,0 +1,47 @@
...
+    password.code.fold(Future.value(service)) { code =>
+      service(Auth(code)).flatMap {
+        case ErrorReply(msg) => Future.exception(ClientError(msg))
+        case _ => Future.value(service)
```
bryce-anderson, line 43: Should we look for the OK status reply?

bryce-anderson (changes requested):

bryce-anderson:
Could we also add a CHANGELOG entry? And thanks for the @krrrr38!

finagle-redis/src/main/scala/com/twitter/finagle/redis/exp/ConnectionInitCommand.scala
krrrr38: extract same code into private method in https://github.com/twitter/finagle/pull/841/commits/88624eff87dbcef835267ff5f75e01095b553ccf

finagle-redis/src/main/scala/com/twitter/finagle/redis/exp/ConnectionInitCommand.scala
krrrr38: Auth and Select returns `StatusReply`. so handle it in https://github.com/twitter/finagle/pull/841/commits/88624eff87dbcef835267ff5f75e01095b553ccf

ref https://github.com/twitter/finagle/blob/c81a87b0cdf9105f7b2985686331d43d4c56067e/finagle-redis/src/main/scala/com/twitter/finagle/redis/ConnectionCommands.scala#L28-L48

krrrr38:
added CHANGELOG.rst entry in 
https://github.com/twitter/finagle/pull/841/commits/88624eff87dbcef835267ff5f75e01095b553ccf

bryce-anderson (approved):
Thank you!

bryce-anderson:
Closed via 9fe05301e56d2cbab001e48216a1b05f5a6e5e6a. Thanks!

PR #849, opened 2020-05-05
finagle-core: Filter IPv6 address-es when NICs have only local ones

yetanotherion:
Problem

When NICs have IPv6 address-es, `InetAddress.getAllByName` sends both
AAAA and A queries for a given host.

When a DNS server returns both IPv4 and IPv6 address-es, the client can
choose to use the IPv6 address.

When the related NIC has only a local IPv6 address, and that the IPv6
address is chosen, subsequent queries will fail given that
none of the available networks is able to route IPv6 packets.

Netty fixed that bug at
https://github.com/netty/netty/pull/10170

That bug is still present in at least openjdk (1.8.0_191, 11.0.1).

Solution

Filter IPv6 address-es from the result of `InetAddress.getAllByName`
when NICs have only local IPv6 address-es.

### DIFF
--- a/finagle-core/src/main/scala/com/twitter/finagle/util/InetSocketAddressUtil.scala
+++ b/finagle-core/src/main/scala/com/twitter/finagle/util/InetSocketAddressUtil.scala
@@ -1,7 +1,11 @@
 package com.twitter.finagle.util
 
 import java.net.{Inet4Address, Inet6Address, InetAddress, InetSocketAddress,
-  NetworkInterface, SocketAddress, SocketException, UnknownHostException}
+  NetworkInterface, SocketAddress, UnknownHostException}
+
+import scala.collection.JavaConverters._
+import scala.util.control.NonFatal
+
 import com.twitter.logging.Logger
 
 object InetSocketAddressUtil {
@@ -13,31 +17,26 @@ object InetSocketAddressUtil {
   private[finagle] val unconnected =
     new SocketAddress { override def toString = "unconnected" }
 
-  private[this] lazy val anyInterfaceSupportsIpV6: Boolean = {
+  private[this] val anyInterfaceSupportsIpV6: Boolean = {
     try {
-      val interfaces = NetworkInterface.getNetworkInterfaces()
-      while (interfaces.hasMoreElements()) {
-        val iface = interfaces.nextElement()
-        val addresses = iface.getInetAddresses()
-        while (addresses.hasMoreElements()) {
-          val inetAddress = addresses.nextElement()
-          if (inetAddress.isInstanceOf[Inet6Address] && !inetAddress.isAnyLocalAddress() &&
-            !inetAddress.isLoopbackAddress() && !inetAddress.isLinkLocalAddress()) {
-            true
-          }
+      val interfaces = NetworkInterface.getNetworkInterfaces().asScala
+      interfaces.exists { interface =>
+        val addresses = interface.getInetAddresses().asScala
+        addresses.exists { inetAddress =>
+          inetAddress.isInstanceOf[Inet6Address] && !inetAddress.isAnyLocalAddress() &&
+            !inetAddress.isLoopbackAddress() && !inetAddress.isLinkLocalAddress()
         }
       }
-    }
-    catch {
-      case e: SocketException => {
+    } catch {
+      case NonFatal(e) => {
         log.debug(s"Unable to detect if any interface supports IPv6, assuming IPv4-only. Error $e")
+        false
       }
     }
-    false
   }
 
   private[finagle] def getAllByName(host: String): Array[InetAddress] = {
-    def isAddressSupported(a: InetAddress) = {
+    def isAddressSupported(a: InetAddress): Boolean = {
       if (!anyInterfaceSupportsIpV6) a.isInstanceOf[Inet4Address]
       else true
     }
--- a/finagle-core/src/main/scala/com/twitter/finagle/util/InetSocketAddressUtil.scala
+++ b/finagle-core/src/main/scala/com/twitter/finagle/util/InetSocketAddressUtil.scala
@@ -1,13 +1,48 @@
 package com.twitter.finagle.util
 
-import java.net.{InetAddress, InetSocketAddress, SocketAddress, UnknownHostException}
+import java.net.{Inet4Address, Inet6Address, InetAddress, InetSocketAddress,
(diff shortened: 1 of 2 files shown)


finagle-core/src/main/scala/com/twitter/finagle/util/InetSocketAddressUtil.scala
```diff
@@ -1,14 +1,49 @@
...
 object InetSocketAddressUtil {
 
   type HostPort = (String, Int)
 
+  private[this] val log = Logger()
+
   private[finagle] val unconnected =
     new SocketAddress { override def toString = "unconnected" }
 
+  private[this] lazy val anyInterfaceSupportsIpV6: Boolean = {
```
hamdiallam, line 16: is there a reason this has to be lazy?

finagle-core/src/main/scala/com/twitter/finagle/util/InetSocketAddressUtil.scala
```diff
@@ -1,14 +1,49 @@
...
+    try {
+      val interfaces = NetworkInterface.getNetworkInterfaces()
+      while (interfaces.hasMoreElements()) {
+        val iface = interfaces.nextElement()
+        val addresses = iface.getInetAddresses()
+        while (addresses.hasMoreElements()) {
+          val inetAddress = addresses.nextElement()
+          if (inetAddress.isInstanceOf[Inet6Address] && !inetAddress.isAnyLocalAddress() &&
+            !inetAddress.isLoopbackAddress() && !inetAddress.isLinkLocalAddress()) {
+            true
```
hamdiallam, line 26: I see this was taken from the patch in the Netty PR. Missing a `return` here for the desired effect.

This is also java-style code. Typically don't use returns in scala code. This could look something like
```
...
import scala.collection.JavaConverters._
...
try {
    val interfaces = NetworkInterface.getNetworkInterfaces().asScala
    interfaces.exists { interface =>
        val addresses = interface.getInetAddresses().asScala
        addresses.exists { inetAddress =>
            inetAddress.isInstanceOf[Inet6Address] && !inetAddress.isAnyLocalAdress() && !inetAddress.isLoopbackAddress() && !inetAddress.isLinkLocalAddress()
        }
    }
} catch {
    case NonFatal(_) =>
        // log
        false
}

File: finagle-core/src/main/scala/com/twitter/finagle/util/InetSocketAddressUtil.scala [hunk 3]
```diff
@@ -1,14 +1,49 @@
...
+    catch {
+      case e: SocketException => {
+        log.debug(s"Unable to detect if any interface supports IPv6, assuming IPv4-only. Error $e")
+      }
+    }
+    false
+  }
+
+  private[finagle] def getAllByName(host: String): Array[InetAddress] = {
+    def isAddressSupported(a: InetAddress) = {
```
hamdiallam, line 40: return type. `): Boolean`

finagle-core/src/main/scala/com/twitter/finagle/util/InetSocketAddressUtil.scala
yetanotherion: We fixed an internal bug at some point by putting lazy to. a field of a companion object. 
(AFAIR there was a cyclic dependency).
Once fixed, a senior member of the team told us, that we'd rather put lazy by defaults in this case. 
Still given it's a local policy and that I did not find any external reference mentioning that practice, I should not have used it here. Sorry for that, I removed the lazy :)

finagle-core/src/main/scala/com/twitter/finagle/util/InetSocketAddressUtil.scala
yetanotherion: done

yetanotherion:
Even after rebasing, the presubmits
https://travis-ci.com/github/criteo-forks/finagle/jobs/353435359
fail with
``` 
(update) sbt.librarymanagement.ResolveException: 
Error downloading com.twitter:scrooge-sbt-plugin;sbtVersion=1.0;scalaVersion=2.12:20.5.0-SNAPSHOT
```
I get a 404 on https://repo1.maven.org/maven2/com/twitter/scrooge-sbt-plugin_2.12_1.0/20.5.0-SNAPSHOT/scrooge-sbt-plugin-20.5.0-SNAPSHOT.pom too
Is it a known infra issue?

hamdiallam:
I'll look into the 404. We updated our release process recently and have run into a few hiccups. Things seem to be working fine now but i'll double check

yetanotherion:
Thanks @hamdiallam!

hamdiallam:
Merged in as of d90df75ecf91591b088cd071526fc86a8398e288

yetanotherion:
@hamdiallam I'm afraid that one modification done to the PR before merging it in develop, **drops the IPv6 compatibility in finagle.**
PR: https://github.com/twitter/finagle/pull/849/files#diff-58bc638ec46e7503824d40754b6863f9R38
Merged version: https://github.com/twitter/finagle/pull/849/files#diff-58bc638ec46e7503824d40754b6863f9R38

If we take the boolean: **anyInterfaceSupportsIpV6=True**

The **filter** in the two versions behave as:
- **PR**: IPv4 packet: true, IPv6 packet: true
- **Merged version**: IPv4 packet: false, IPv6 packet: **false**

Which is not the wanted behavior, WDYT?

(The two **filter** are coherent when **anyInterfaceSupportsIpV6=False**)

hamdiallam:
Ah yes, you're right. My apologies.

Will push a commit that fixes that

mosesn:
whoops, this was my bad suggestion, sorry @yetanotherion!

yetanotherion:
Thanks for fixing that so fast, could you ping me when the fix will be merged in the develop branch?

hamdiallam:
np & will do. Should be up sometime soon today

hamdiallam:
commit was just pushed! 4fa66b4b1244dad2797e6a8fd94a6f85da28c0bd

yetanotherion:
Great news,
additionally https://github.com/twitter/finagle/commit/4fa66b4b1244dad2797e6a8fd94a6f85da28c0bd
looks much better than
https://github.com/twitter/finagle/pull/849/files#diff-58bc638ec46e7503824d40754b6863f9R38
thanks!

PR #851, opened 2020-05-15
Fix instructions to build via dodo

sideshowcoder:
The `build` script for dodo has been split into multiple so the curl oneliner no
longer works. I think the most straight forward way is to clone dodo and
actually run it.

Dodo depends on bash > 4, which is not the default bash on MacOSX.

Problem

Running the commands in `CONTRIBUTING.md` fails as dodo no longer is a single file bin.

```
bash: line 5: BASH_SOURCE[0]: unbound variable
bash: line 5: /Users/USER/source/finagle/functions: No such file or directory
```

Solution

Tell the user to clone dodo and and use from the clones repository. Also inform the user that running dodo requires bash > 4

Result

Instructions in `CONTRIBUTING.md` result in a successful built.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -9,9 +9,11 @@ means that it may be necessary to publish the `develop` branches of these
 libraries locally in order to work on Finagle's `develop` branch. To do so
 you can use our build tool, [dodo](https://github.com/twitter/dodo).
 
-``` bash
-curl -s https://raw.githubusercontent.com/twitter/dodo/develop/bin/build | bash -s -- --no-test finagle
-```
+1. Clone [dodo](https://github.com/twitter/dodo.git)
+2. `./dodo/bin/build --no-test finagle`
+
+dodo depends on bash version > 4 to run correctly, in case of an error that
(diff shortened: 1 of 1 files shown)


cacoco:
@sideshowcoder interesting. We run this on OS X (the latest default shell is zsh). It surely wasn't the intention to cause this to break. Let me see what we can do. Thanks!

sideshowcoder:
Thanks, I'm not to concerned about the bash > 4 actually was really surprised that that could be the reason also why MacOS sticks with bash 3.2 is a mystery. 

I think the bigger thing is having a local clone of dodo which makes sense to have I guess.

cacoco:
@sideshowcoder yeah, we don't want to force users to clone this repo. We're going to back out of some of the changes to maintain the integrity of the `bin/build` script. Changes should be forthcoming shortly. Thanks.

cacoco:
@sideshowcoder Ok, addressed in https://github.com/twitter/dodo/commit/16ab69dab88b3b58837e97d7164e4603d4dc57cf. Thanks again!

PR #865, opened 2020-07-14
finagle-base-http: Add MultipartDecoder support for chunked requests

jimbo-hf:
Problem

Multipart decoding of chunked requests currently returns `None`.

Solution

Allow chunked requests to be passed to `MultipartDecoder` implementation.
For `Netty4MultipartDecoder`, offer all chunks from request to decoder before attempting to `getBodyHttpDatas`.

Result

`MultipartDecoder` now returns a `Future` and supports decoding of chunked requests.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -10,6 +10,9 @@ Unreleased
 Breaking API Changes
 ~~~~~~~~~~~~~~~~~~~~
 
+* finagle-base-http: Change return type of `MultipartDecoder.decodeFull` to return a `Future` and add
+  support for chunked requests.
+
 * finagle-core: Correct the spelling of `Tracing.recordClientSendFrargmet()` to
   `Tracing.recordClientSendFragment()` ``PHAB_ID=D505617``
 
--- a/finagle-base-http/src/main/scala/com/twitter/finagle/http/exp/MultipartDecoder.scala
+++ b/finagle-base-http/src/main/scala/com/twitter/finagle/http/exp/MultipartDecoder.scala
@@ -2,7 +2,7 @@ package com.twitter.finagle.http.exp
 
 import com.twitter.finagle.http.{MediaType, Method, Request}
 import com.twitter.finagle.util.LoadService
-import com.twitter.util.StorageUnit
+import com.twitter.util.{Future, StorageUnit}
 
 /**
  * A utility that represents a decoder for a [[Multipart]] data.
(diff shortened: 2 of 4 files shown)


mosesn:
This is really cool!  Thanks so much for the PR @jimbo-hf :). Before we dive into the code review, let's chat a little bit about this design.  My understanding of the proposed implementation is to update `decodeFull` to return a `Future`, reading the entire stream into memory, and then using the same `Multipart` abstractions that users are used to.  This provides a simple abstraction for users, and doesn't require adding new types, which is a great design property!

However, the downsides are that it's a breaking change, and it loses some of the flexibility of streaming data.  For example, you can't process the data incrementally.

What would you think about keeping the current API, but introducing a new implementation of `Multipart.FileUpload` which exposes a `Reader` instead of a `Buf`?  The main downside is that it may fail at runtime instead of compile-time when people update and suddenly start multipart-decoding streamed messages, but we could get around that by potentially marking Multipart as `sealed`?  wdyt?

jimbo-hf:
@mosesn so we've been discussing internally and we think there are three 'levels' of interface we could provide:

a) the current Future-based one we've got: simple, and we know it works, but doesn't provide any opportunity to take advantage of streaming (I don't think this is quite as bad as it sounds though because netty will stream large files to disk for us anyway)

b) one that provides a Reader[Part], where each Part is either an attribute or a file in a format similar to the current one: this is not actually all that great an API imo because you don't know what you're going to get for each part until you get it - I think many common use cases will end up streaming the whole thing up-front anyway; I also don't think this is possible using the netty-provided decoder, which expects to have seen all the data before you can start reading it

c) a modification of the above where the individual files are always provided as streams: quite nice because it avoids forcing the in-memory/on-disk decision onto the user, but note that you'd have to finish streaming each file before continuing reading from the main stream

mosesn:
c) sounds ideal to me.  Thinking more about a), I feel like the benefit of streaming is that you can parse parts of it before fully buffering the message.  But in a) all you get is being able to parse the headers before the rest of the message, which doesn't seem as valuable.  For a), I wonder if a good work-around would be to disable streaming?  That way I think netty would rewrite the chunked message under the hood to be fully buffered.  If it's not, that might be a good bugfix?

jimbo-hf:
@mosesn thanks for the feedback
Think we'll disable streaming for now on our side. Might come back and take a look at option (c) in the future if we get a chance.
Will close the PR.

PR #866, opened 2020-07-17
finagle-redis: Add the RedisTracingFilter and RedisLoggingFilter back into the default Redis Stack

rpless:
Problem

I was experimenting with the Tracing module and instrumented a Redis Client. I noticed that Redis rpc calls did not get named unlike the Http calls that were being traced. After some discussion on gitter, @mosesn pointed out that a commit where the `RedisLoggingFilter` and `RedisTracingFilter` had been ported during the move from Netty 3 -> Netty 4, but had not been added back to the stack.

Solution

I added the `RedisLoggingFilter` and the `RedisTracingFilter` to the default client stack.

Result

This will add stats and tracing back to the Redis Client by default.

I followed the mysql client for adding the `role` and `module` definitions. If there as a better way to do this let me know and I can change it.

### DIFF
--- a/finagle-redis/src/main/scala/com/twitter/finagle/Redis.scala
+++ b/finagle-redis/src/main/scala/com/twitter/finagle/Redis.scala
@@ -74,7 +74,7 @@ object Redis extends Client[Command, Reply] with RedisRichClient {
       .insertBefore(DefaultPool.Role, RedisPool.module)
       .insertAfter(StackClient.Role.prepConn, ConnectionInitCommand.module)
       .replace(StackClient.Role.protoTracing, RedisTracingFilter.module)
-      .insertAfter(RedisLoggingFilter.role, RedisLoggingFilter.module)
+      .insertBefore(StackClient.Role.protoTracing, RedisLoggingFilter.module)
 
     private[finagle] val hashRingStack: Stack[ServiceFactory[Command, Reply]] =
       stack.insertAfter(BindingFactory.role, RedisPartitioningService.module)
--- a/finagle-redis/src/main/scala/com/twitter/finagle/Redis.scala
+++ b/finagle-redis/src/main/scala/com/twitter/finagle/Redis.scala
@@ -14,6 +14,7 @@ import com.twitter.finagle.param.{
 }
 import com.twitter.finagle.redis.RedisPartitioningService
 import com.twitter.finagle.redis.exp.{ConnectionInitCommand, RedisPool}
+import com.twitter.finagle.redis.filter.{RedisLoggingFilter, RedisTracingFilter}
 import com.twitter.finagle.redis.param.{Database, Password}
 import com.twitter.finagle.redis.protocol.{Command, Reply, StageTransport}
 import com.twitter.finagle.service.{ResponseClassifier, RetryBudget}
@@ -72,6 +73,8 @@ object Redis extends Client[Command, Reply] with RedisRichClient {
     private val stack: Stack[ServiceFactory[Command, Reply]] = StackClient.newStack
       .insertBefore(DefaultPool.Role, RedisPool.module)
       .insertAfter(StackClient.Role.prepConn, ConnectionInitCommand.module)
+      .replace(StackClient.Role.protoTracing, RedisTracingFilter.module)
+      .insertAfter(RedisLoggingFilter.role, RedisLoggingFilter.module)
 
     private[finagle] val hashRingStack: Stack[ServiceFactory[Command, Reply]] =
       stack.insertAfter(BindingFactory.role, RedisPartitioningService.module)
--- a/finagle-redis/src/main/scala/com/twitter/finagle/redis/filter/RedisLoggingFilter.scala
+++ b/finagle-redis/src/main/scala/com/twitter/finagle/redis/filter/RedisLoggingFilter.scala
@@ -1,12 +1,30 @@
 package com.twitter.finagle.redis.filter
 
-import com.twitter.finagle.{Service, SimpleFilter}
+import com.twitter.finagle._
 import com.twitter.finagle.redis.protocol._
 import com.twitter.finagle.redis.util.BufToString
 import com.twitter.finagle.stats.StatsReceiver
 import com.twitter.util.{Future, Return}
 
-private[redis] class RedisLoggingFilter(stats: StatsReceiver) extends SimpleFilter[Command, Reply] {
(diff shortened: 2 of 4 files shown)


rpless:
Realized I pushed with an old work account which is why the CLA wasn't showing up as signed. Fixed by pushing the same commit over the original one but associated with the correct email.

finagle-redis/src/main/scala/com/twitter/finagle/redis/filter/RedisLoggingFilter.scala
```diff
@@ -1,12 +1,30 @@
...
+  def module: Stackable[ServiceFactory[Command, Reply]] =
+    new Stack.Module1[param.Stats, ServiceFactory[Command, Reply]] {
+      val role: Stack.Role = RedisLoggingFilter.role
+      val description: String = RedisLoggingFilter.description
+
+      def make(statsParam: param.Stats, next: ServiceFactory[Command, Reply]): ServiceFactory[Command, Reply] = {
+        if (statsParam.statsReceiver.isNull)
+          next
+        else
+          new RedisLoggingFilter(statsParam.statsReceiver.scope("http")).andThen(next)
```
bryce-anderson: I suspect the scope isn't really supposed to be `"http"`.

bryce-anderson (changes requested):

finagle-redis/src/main/scala/com/twitter/finagle/redis/filter/RedisLoggingFilter.scala
rpless: Ah yes this looks copy-paste on my part, the original commit does not seem to add a scope. I can remove it. Alternatively some modules seem to scope the receiver to the protocol name like `http` so I could change it to that if it makes sense.

bryce-anderson: I suspect we want to follow what was originally there. It would normally be the job of whoever passes the stats receiver to us to define the scope, and if we need to go deeper this module would be in charge of naming that scope and passing it on.

finagle-redis/src/main/scala/com/twitter/finagle/Redis.scala
```diff
@@ -72,6 +73,8 @@ object Redis extends Client[Command, Reply] with RedisRichClient {
     private val stack: Stack[ServiceFactory[Command, Reply]] = StackClient.newStack
       .insertBefore(DefaultPool.Role, RedisPool.module)
       .insertAfter(StackClient.Role.prepConn, ConnectionInitCommand.module)
+      .replace(StackClient.Role.protoTracing, RedisTracingFilter.module)
+      .insertAfter(RedisLoggingFilter.role, RedisLoggingFilter.module)
```
hamdiallam: Does the `RedisLoggingFilter.role` exist in the client stack already? I believe using `insertAfter` here will not put it into the stack, since the role doesn't exists. Maybe before the tracing filter?

rpless: Ah yes, the role does not currently exist in the stack. Docs from `insertAfter` say: `If no elements match the role, then an unmodified stack is returned.` So looks like this is currently won't add the logging filter. I'll change it to insert before the tracing filter.

mosesn (approved):

bryce-anderson (approved):

jyanJing:
Hi @rpless ,
@enbnt and I have been having troubles pulling in this request, we both see the following error:

```
error: while searching for:
import com.twitter.util.{Duration, Monitor}
import java.net.SocketAddress

import com.twitter.finagle.redis.param.{Database, Password}

trait RedisRichClient { self: Client[Command, Reply] =>

error: patch failed: finagle/finagle-redis/src/main/scala/com/twitter/finagle/Redis.scala:23
error: finagle/finagle-redis/src/main/scala/com/twitter/finagle/Redis.scala: patch does not apply
Checking patch finagle/finagle-redis/src/main/scala/com/twitter/finagle/redis/filter/RedisLoggingFilter.scala...
Checking patch finagle/finagle-redis/src/main/scala/com/twitter/finagle/redis/filter/RedisTracingFilter.scala...
Checking patch finagle/finagle-redis/src/main/scala/com/twitter/finagle/redis/filter/RedisLoggingFilter.scala...
Checking patch finagle/finagle-redis/src/main/scala/com/twitter/finagle/Redis.scala...
error: while searching for:
      .insertBefore(DefaultPool.Role, RedisPool.module)
      .insertAfter(StackClient.Role.prepConn, ConnectionInitCommand.module)
      .replace(StackClient.Role.protoTracing, RedisTracingFilter.module)
      .insertAfter(RedisLoggingFilter.role, RedisLoggingFilter.module)

    private[finagle] val hashRingStack: Stack[ServiceFactory[Command, Reply]] =
      stack.insertAfter(BindingFactory.role, RedisPartitioningService.module)

error: patch failed: finagle/finagle-redis/src/main/scala/com/twitter/finagle/Redis.scala:74
error: finagle/finagle-redis/src/main/scala/com/twitter/finagle/Redis.scala: patch does not apply
```

@enbnt has tried to do a merge on Github but the error persists. 

Would you please do a rebase then I can try to pull it again? Thank you!

rpless:
Hi @jyanJing I believe that happens because one of the commits in this PR conflicts with [this commit in develop](https://github.com/twitter/finagle/commit/f8ab37c34cc5ec7b364cd18b7d49699d680724ac). I have rebased this Pull Request against `develop`. Please let me know if there are any issues.

jyanJing:
Hi @rpless, 

Thank you for rebasing the branch! I am able to pull it in and get the team's review. Since this change is made to the stack module with the tracing and logging filter, we are very careful with changes in that area because it might bring up latencies for services that use finagle-redis. So I am working with 2 other teams to canary the change now, to make sure we capture any behavior changes. It is not about your change, it is a critical area to be precautious about.

I wanted to let you know that it will take longer to merge this change, but I am actively working on it! Really appreciate for making this change!

Best regards,
Jing

jyanJing:
Hi @rpless 

canaries looks good, the change is [merged](https://github.com/twitter/finagle/commit/8755363c6ef9c87e8dd49007e8a44992d609ab03),

Thank you for the excellent work and your patience!

rpless:
Thanks for setting up the canary to test the change @jyanJing!

PR #876, opened 2020-11-11
Update build.sbt to get aarch64 binaries and fix behavior of NonReentrantReadWriteLock on Arm

geoffreyblake:
Problem

There are 2 small problems addressed in this PR:
1) build.sbt does not properly pull down netty-epoll-native for
Arm64,  meaning by default Arm with use NIO instead of
native-epoll. 
2) Current behavior of NonReentrantReadWriteLock in the 
finagle-stats-core package is sub-optimal on Arm and will enter into
slow-path synchronization via kernel futexes more often than
expected (measuring 100x more often) in some cases.

Solution

1) Modify build.sbt to get the aarch64 packages for netty-epoll-native
in addition to x86
2) Try the fast path acquire up to 5 times before failing over to the
AbstractQueuedSynchronizer slow path in NonReentrantReadWriteLock
for Arm64, all other systems try the fast path once.

### DIFF
--- a/finagle-stats-core/src/main/scala/com/twitter/finagle/stats/NonReentrantReadWriteLock.scala
+++ b/finagle-stats-core/src/main/scala/com/twitter/finagle/stats/NonReentrantReadWriteLock.scala
@@ -3,6 +3,17 @@ package com.twitter.finagle.stats
 import java.util.concurrent.locks.AbstractQueuedSynchronizer
 import scala.annotation.tailrec
 
+import com.twitter.app.GlobalFlag
+
+object rwLockSpinWait
+    // Defaults to -1 which allows the NonReentrantReadWriteLock class to 
+    // set a system specific default.
+    extends GlobalFlag[Int](-1, 
+      """Experimental flag.  Control how many times NonReentrantReadWriteLock 
+      | retries an acquire before executing its slow path acquire.  Default
+      | value of -1 lets the system decide how long to spin""".stripMargin
+    )
+
 /**
  * This is a cheaper read-write lock than the ReentrantReadWriteLock.
  *
@@ -23,10 +34,30 @@ private[stats] final class NonReentrantReadWriteLock extends AbstractQueuedSynch
   // we haven't provided an implementation for isHeldExclusively because the
   // docs for AQS advise implementors to provide implementations "as applicable"
   // and our use case doesn't require knowing when the lock is held exclusively.
+  private val SpinMax: Int = if (rwLockSpinWait() == -1) {
+    // Default number of spins for aarch64 machines set to 5 by default. 
+    // See below comment on reason.
+    if (System.getProperty("os.arch") == "aarch_64") 5
+    else 1
+  } 
+  else rwLockSpinWait()
 
   override def tryAcquireShared(num: Int): Int = {
-    val cur = getState()
-    if (cur >= 0 && compareAndSetState(cur, cur + 1)) 1 else -1
+    tryAcquireShared(getState(), 1)
+  }
+
+  // Try to acquire the lock up to SpinMax times before failing and 
+  // allowing AbstractQueuedSynchronizer to execute the slow-path lock.
+  // This is important on Arm systems where compareAndSetState may fail
+  // more often due to its weaker memory model.
+  @tailrec
+  private[this] def tryAcquireShared(cur: Int, tries: Int): Int = {
+    if (cur >= 0 && compareAndSetState(cur, cur + 1)) 1
+    // When JDK8 support is deprecated, Thread.onSpinWait() should be
+    // used in the else if clause to avoid flooding the memory 
+    // system with compareAndSetState in a tight loop.
+    else if (tries < SpinMax) tryAcquireShared(getState(), tries + 1)
+    else -1
   }
 
   @tailrec
--- a/build.sbt
+++ b/build.sbt
@@ -44,6 +44,7 @@ val netty4Libs = Seq(
   "io.netty" % "netty-handler" % netty4Version,
   "io.netty" % "netty-transport" % netty4Version,
   "io.netty" % "netty-transport-native-epoll" % netty4Version classifier "linux-x86_64",
+  "io.netty" % "netty-transport-native-epoll" % netty4Version classifier "linux-aarch64",
   // this package is a dep of native-epoll above, explicitly add this for coursier plugin
   "io.netty" % "netty-transport-native-unix-common" % netty4Version,
   "io.netty" % "netty-handler-proxy" % netty4Version
@@ -52,6 +53,7 @@ val netty4LibsTest = Seq(
   "io.netty" % "netty-handler" % netty4Version % "test",
   "io.netty" % "netty-transport" % netty4Version % "test",
   "io.netty" % "netty-transport-native-epoll" % netty4Version % "test" classifier "linux-x86_64",
+  "io.netty" % "netty-transport-native-epoll" % netty4Version classifier "linux-aarch64",
   // this package is a dep of native-epoll above, explicitly add this for coursier plugin
   "io.netty" % "netty-transport-native-unix-common" % netty4Version % "test",
   "io.netty" % "netty-handler-proxy" % netty4Version % "test",


geoffreyblake:
While I'm waiting on getting approval for the CLA, here is reproducer of the problem we've seen with the NonReentrantReadWriteLock.  On Arm we've seen a 50% performance improvement with this PR when enabling stats:

```scala
import net.blakgeof.benchmark.thrift._

import com.twitter.app.App
import com.twitter.conversions.DurationOps._
import com.twitter.finagle._
import com.twitter.finagle.thrift.Protocols
import com.twitter.util.{Await, Duration, Future, FuturePool}
import com.twitter.finagle.ThriftMux
import com.twitter.finagle.stats.{MetricsStatsReceiver, LoadedStatsReceiver, StatsReceiver, Stat}
import com.twitter.finagle.util.DefaultTimer.Implicit


object Main extends App {
  val batch = flag[Int]("batch", 500000, "how many total requests to make")
  val depth = flag[Int]("depth", 1000, "how many concurrent requests to make")
  val ips = flag[String]("ips", "127.0.0.1:8888", "list of IPs to contact")
  val nclient = flag[Int]("nclient", 32, "how many concurrent clients to run")
  val test = flag[String]("test", "shared", "Test to perform: shared")
  val stats = flag[Boolean]("stats", false, "Toggle stats on/off")
  val garbage = "*"

  // best practice to turn this on.
  // The app won't start if parameters are wrong or missing
  override def failfastOnFlagsNotParsed: Boolean = true

  // Sends requests, but does not log any stats
  def processClientUntimed(c: BenchmarkService.MethodPerEndpoint): Unit = {
    var ctr: Int = batch()
    while (ctr > 0) {
      val futures = for (_ <- 1 to depth()) yield {
        val datum = BenchmarkDatum(0, garbage)
        c.reflect(datum)
      }
      futures.map(Await.ready(_, 1.day))
      ctr = ctr - depth()
    }
  }

  // Uses a stat to time the Future, though, not sure how to do an open loop tester here...
  def processClientTimed(c: BenchmarkService.MethodPerEndpoint, l: Stat): Unit = {
    var ctr: Int = batch()
    while (ctr > 0) {
      val futures = for (_ <- 1 to depth()) yield {
        val datum = BenchmarkDatum(0, garbage)
        Stat.timeFuture(l) { c.reflect(datum) }
      }
      futures.map(Await.ready(_, 1.day))
      ctr = ctr - depth()
    }
  }

  // Create a batch of Futures to spawn asynchronously and wait, goes for maximum throughput.
  def sendLoop(latency: Stat, clients: Seq[BenchmarkService.MethodPerEndpoint]): Unit = {
    val clientFutures = for (client <- clients) yield {
      FuturePool.unboundedPool { 
        stats() match {
          case true => processClientTimed(client, latency)
          case false => processClientUntimed(client)
        }
      }
    }
    clientFutures.map(Await.ready(_, 1.day))
  }

  // Use a shared client for all threads
  def testSharedClient(s: StatsReceiver): Unit = {
    var latency: Stat = null
    val clients = stats() match {
      case true => 
        val client = ThriftMux.client
          .withStatsReceiver(s)
          .withPerEndpointStats
          .build[BenchmarkService.MethodPerEndpoint](s"${ips()}")
        latency = s.stat("latency_ms")
        for (_ <- 1 to nclient()) yield { client }
      case false => 
        val client = ThriftMux.client
          .build[BenchmarkService.MethodPerEndpoint](s"${ips()}")
        for (_ <- 1 to nclient()) yield { client }
    }
    println(s"Sending work through shared client, each with ${ips().split(",").size} backends")
    sendLoop(latency, clients)
  }

  def main() {
    println(s"Hello, Client!")

    // Create a default stats reciever
    var metricsStatsReceiver: StatsReceiver = null
    if (stats()) {
      metricsStatsReceiver = new MetricsStatsReceiver()
      LoadedStatsReceiver.self = metricsStatsReceiver
    }

    test() match {
      case "shared" => testSharedClient(metricsStatsReceiver)
      case _ => println(s"ERROR: ${test()} not supported")
    }
  }
}
```

mosesn:
This is great work, thanks @geoffreyblake!  How would you feel about making this optimization based on the architecture?  Eg check if we're using ARM, and only use the spin-waiting approach then?  Also, have you benchmarked this approach vs the current approach on x86?  I wonder if this is an improvement on x86 too, or if it's something peculiar to ARM that makes it faster.

One other thing I wonder is whether this approach is still faster on ARM than the other approaches we looked at (spinning on atomicint, phaser, reentrantrwlock) which were slower on x86.  Would you be down to take a stab at re-benchmarking this for us on ARM?  I can try to go back and find the research we did here to settle on the AQS impl if you have the bandwidth.

geoffreyblake:
@mosesn , I'll run the above test case on my x86 server test box with this fix to make sure there are no ill effects.  My Scala knowledge is still rather rudimentary, so is there a preferred way to do architecture detection if there is a difference in behavior on x86?

mosesn:
@geoffreyblake thanks! I think the "os.arch" [system property](https://docs.oracle.com/javase/tutorial/essential/environment/sysprop.html) is probably our best bet.  We can access it the same way as you do from Java–there's some Scala sugar for access system properties, but I don't think we have to use it.

geoffreyblake:
@mosesn , sorry for the wait, but I tested on x86 with my reproducer attached to this PR and see no harm done with the code as is.  I did update it to be tunable through the flag `com.twitter.finagle.stats.rwLockSpinWait`, and set the default to `5` after some runs on my Arm system.    

As for why Arm is affected by this more, Arm does have a weak memory model compared to x86 so it is plausible the Arm system can see the old value of the flag more easily and fail its CompareAndSet when trying to acquire.  A small amount of spinning helps get the most recent value on Arm to the core attempting the CompareAndSet.  

I do not think this needs to be made more complex as the AbstractQueuedSynchronizer does a full-fledged MCS style lock once the fast-path fails.

finagle-stats-core/src/main/scala/com/twitter/finagle/stats/NonReentrantReadWriteLock.scala
```diff
@@ -25,8 +26,24 @@ private[stats] final class NonReentrantReadWriteLock extends AbstractQueuedSynch
...
   override def tryAcquireShared(num: Int): Int = {
-    val cur = getState()
-    if (cur >= 0 && compareAndSetState(cur, cur + 1)) 1 else -1
+    val SpinMax = 20
+    var ret: Int = -1
+    var cur: Int = 0
+    var tries: Int = 0
+
+    // Try to acquire the lock up to SpinMax times before failing and 
+    // allowing AbstractQueuedSynchronizer to execute the slow-path lock.
```
mosesn: can we add a note this is for ARM reasons?

finagle-stats-core/src/main/scala/com/twitter/finagle/stats/NonReentrantReadWriteLock.scala
```diff
@@ -25,8 +26,24 @@ private[stats] final class NonReentrantReadWriteLock extends AbstractQueuedSynch
   // and our use case doesn't require knowing when the lock is held exclusively.
 
   override def tryAcquireShared(num: Int): Int = {
-    val cur = getState()
-    if (cur >= 0 && compareAndSetState(cur, cur + 1)) 1 else -1
+    val SpinMax = 20
+    var ret: Int = -1
```
mosesn, line 30: this might be easier to read as a boolean?

var acquired = false

finagle-stats-core/src/main/scala/com/twitter/finagle/stats/NonReentrantReadWriteLock.scala
```diff
@@ -25,8 +26,24 @@ private[stats] final class NonReentrantReadWriteLock extends AbstractQueuedSynch
...
+    // Try to acquire the lock up to SpinMax times before failing and 
+    // allowing AbstractQueuedSynchronizer to execute the slow-path lock.
+    do {
+      cur = getState()
+      ret = if (cur >= 0 && compareAndSetState(cur, cur + 1)) {
+        1
+      } else {
+        Thread.onSpinWait()
+        -1
+      }
```
mosesn: I think this reads more simply as:

```
if (cur >= 0 && compareAndSetState(cur, cur + 1)) {
  ret = 1
} else {
  Thread.onSpinWait()
}
```

finagle-stats-core/src/main/scala/com/twitter/finagle/stats/NonReentrantReadWriteLock.scala
```diff
@@ -25,8 +26,24 @@ private[stats] final class NonReentrantReadWriteLock extends AbstractQueuedSynch
...
+    var tries: Int = 0
+
+    // Try to acquire the lock up to SpinMax times before failing and 
+    // allowing AbstractQueuedSynchronizer to execute the slow-path lock.
+    do {
+      cur = getState()
+      ret = if (cur >= 0 && compareAndSetState(cur, cur + 1)) {
+        1
+      } else {
+        Thread.onSpinWait()
```
mosesn: hmm, looking at this more deeply, it seems like Thread.onSpinWait only shows up in JDK9, but we haven't deprecated JDK8 yet.  How important is this to the ARM implementation?

mosesn (changes requested):
Thanks!  Can you post the jmh benchmark results?  And which benchmark did you run, was it https://github.com/twitter/finagle/blob/develop/finagle-benchmark/src/main/scala/com/twitter/finagle/stats/StatsReceiverBenchmark.scala ?  Or did you only run the macro benchmark you posted here?

finagle-stats-core/src/main/scala/com/twitter/finagle/stats/NonReentrantReadWriteLock.scala
geoffreyblake: Please let me know if what I did is not going to be accepted on style reasons, but I simply commented it out with an explanation of what to do when JDK8 is deprecated.  `onSpinWait()` can be omitted, it has a small negative effect though because the system is not getting throttled when it should be, so instead `compareAndSetState` gets spammed in a tight loop, making it pause a little in between is good for both x86 and Arm.

finagle-stats-core/src/main/scala/com/twitter/finagle/stats/NonReentrantReadWriteLock.scala
geoffreyblake: see below

geoffreyblake:
As for running JMH benchmarks @mosesn, please paste the sbt command line I need to run, be happy to report the results.  So far I've just been running my above reproducer for now.

mosesn:
I think `./sbt 'project finagle-benchmark' 'jmh:run .*StatsReceiver.*'` will work.  Trying it out locally to confirm that it runs in multi-threaded mode by default . . .

finagle-stats-core/src/main/scala/com/twitter/finagle/stats/NonReentrantReadWriteLock.scala
```diff
@@ -23,10 +32,31 @@ private[stats] final class NonReentrantReadWriteLock extends AbstractQueuedSynch
   // we haven't provided an implementation for isHeldExclusively because the
   // docs for AQS advise implementors to provide implementations "as applicable"
   // and our use case doesn't require knowing when the lock is held exclusively.
+  val SpinMax: Int = rwLockSpinWait.get.get
```
mosesn: s/rwLockSpinWait.get.get/rwLockSpinWait()

mosesn:
OK, I tried it locally, here's what you actually want: `./sbt 'project finagle-benchmark' 'jmh:run -t 3 .*MetricsStatsReceiver.*'`

finagle-stats-core/src/main/scala/com/twitter/finagle/stats/NonReentrantReadWriteLock.scala
mosesn: This looks fine to me, since you left an explanatory comment.  Got it, that makes sense!

geoffreyblake:
Here are the results I got from the benchmarks:
The scores have moved around some, especially for the `MetricsStatsReceiverBenchmark.add` workload, it improved on Arm, though it did get worse on x86.  For a macro-workload such as the attached reproducer, this increase in latency did no harm.  If we want to special case this for Arm, I can and rerun the micros and the macro workload to see if anything changes.

```
Arm64
Benchmark                                      Mode  Cnt    Score    Error  Units
MetricsStatsReceiverBenchmark.add              avgt    5  755.577 ± 67.337  ns/op
                                                          583.332 ± 44.895  ns/op
MetricsStatsReceiverBenchmark.counter          avgt    5  290.489 ± 20.528  ns/op
                                                          269.848 ± 15.975  ns/op
MetricsStatsReceiverBenchmark.counterAndIncr   avgt    5  406.187 ± 17.464  ns/op
                                                          354.792 ± 10.801  ns/op
MetricsStatsReceiverBenchmark.incr             avgt    5   10.531 ±  0.083  ns/op
                                                           10.485 ±  0.193  ns/op
MetricsStatsReceiverBenchmark.scope            avgt    5   98.358 ± 10.333  ns/op
                                                          121.884 ±  0.483  ns/op
MetricsStatsReceiverBenchmark.scopeAndCounter  avgt    5  354.088 ± 11.787  ns/op
                                                          422.523 ± 13.759  ns/op
MetricsStatsReceiverBenchmark.scopeAndStat     avgt    5  433.229 ±  7.257  ns/op
                                                          414.470 ± 24.180  ns/op
MetricsStatsReceiverBenchmark.scopedCounter    avgt    5  332.714 ±  6.423  ns/op
                                                          324.314 ±  5.970  ns/op
MetricsStatsReceiverBenchmark.scopedStat       avgt    5  307.627 ±  8.480  ns/op
                                                          339.571 ± 24.469  ns/op
MetricsStatsReceiverBenchmark.stat             avgt    5  202.166 ± 11.054  ns/op
                                                          251.819 ± 21.994  ns/op
MetricsStatsReceiverBenchmark.statAndAdd       avgt    5  613.260 ± 22.566  ns/op
                                                          703.450 ± 31.770  ns/op
```

```
x86
MetricsStatsReceiverBenchmark.add              avgt    5  186.206 ± 3.173  ns/op
                                                          327.102 ± 1.283  ns/op
MetricsStatsReceiverBenchmark.counter          avgt    5  128.572 ± 2.518  ns/op
                                                          133.856 ± 3.094  ns/op
MetricsStatsReceiverBenchmark.counterAndIncr   avgt    5  155.888 ± 1.544  ns/op
                                                          118.834 ± 2.569  ns/op
MetricsStatsReceiverBenchmark.incr             avgt    5   12.639 ± 0.013  ns/op
                                                           12.653 ± 0.082  ns/op
MetricsStatsReceiverBenchmark.scopeAndCounter  avgt    5  218.245 ± 2.924  ns/op
                                                          209.605 ± 6.039  ns/op
MetricsStatsReceiverBenchmark.scopeAndStat     avgt    5  210.151 ± 5.322  ns/op
                                                          213.807 ± 4.469  ns/op  
MetricsStatsReceiverBenchmark.scopedCounter    avgt    5  133.982 ± 6.524  ns/op
                                                          135.596 ± 3.535  ns/op
MetricsStatsReceiverBenchmark.scopedStat       avgt    5  130.239 ± 1.935  ns/op
                                                          101.527 ± 2.505  ns/op
MetricsStatsReceiverBenchmark.stat             avgt    5  123.955 ± 0.693  ns/op
                                                          126.539 ± 2.976  ns/op
MetricsStatsReceiverBenchmark.statAndAdd       avgt    5  255.342 ± 1.162  ns/op
                                                          257.764 ± 0.752  ns/op
```

finagle-stats-core/src/main/scala/com/twitter/finagle/stats/NonReentrantReadWriteLock.scala
geoffreyblake: Fixed

mosesn:
@geoffreyblake would you mind running the `add` benchmark again to confirm the current results (without changes) and then run it after adding the architecture check for x86?  These microbenchmarks are pretty sensitive to the other stuff running on your computer, so I normally run them a couple times if I have a big swing between two runs.

If the difference here between the `add` metrics is correct, then we would definitely want to make this architecture-specific.  Although it seems like a corner case, the reason we made this investment was because our customers told us their histograms were contended on `add`.

geoffreyblake:
@mosesn, I specialized the code as the x86 results are definitely better on `add` with only 1 spin.  The `add` operation was where we originally saw contention on arm, which is the reason for this PR.   Results below with arm defaulting to 5 spins and x86 now set to 1. 

```
Arm64 results with specialized code, default to 5 spins:
[info] Benchmark                                      Mode  Cnt    Score    Error  Units
[info] MetricsStatsReceiverBenchmark.add              avgt    5  365.852 ± 54.946  ns/op
```

```
x86 results with specialized code
[info] Benchmark                                      Mode  Cnt    Score   Error  Units
[info] MetricsStatsReceiverBenchmark.add              avgt    5  181.759 ± 3.400  ns/op
```

mosesn (approved):
thanks for bearing with me, I think this patch looks great!  I'm out for Thanksgiving, but one of my coworkers should pull this in shortly.

finagle-stats-core/src/main/scala/com/twitter/finagle/stats/NonReentrantReadWriteLock.scala
```diff
@@ -23,10 +35,41 @@ private[stats] final class NonReentrantReadWriteLock extends AbstractQueuedSynch
   // we haven't provided an implementation for isHeldExclusively because the
   // docs for AQS advise implementors to provide implementations "as applicable"
   // and our use case doesn't require knowing when the lock is held exclusively.
+  val SpinMax: Int = if (rwLockSpinWait() == -1) {
```
bryce-anderson: Can this be private?

finagle-stats-core/src/main/scala/com/twitter/finagle/stats/NonReentrantReadWriteLock.scala
```diff
@@ -23,10 +35,41 @@ private[stats] final class NonReentrantReadWriteLock extends AbstractQueuedSynch
...
+      if (cur >= 0 && compareAndSetState(cur, cur + 1)) {
+        ret = 1
+      } else {
+        // XXX: When JDK8 support is deprecated, onSpinWait() should be
+        //      used to avoid flooding the memory system with compareAndSetState in
+        //      a tight loop.
+        // Thread.onSpinWait()
+        ret = -1
+      }
+      tries = tries + 1
```
bryce-anderson: ffti: `tries += 1`

finagle-stats-core/src/main/scala/com/twitter/finagle/stats/NonReentrantReadWriteLock.scala
```diff
@@ -23,10 +35,41 @@ private[stats] final class NonReentrantReadWriteLock extends AbstractQueuedSynch
...
+    do {
+      cur = getState()
+      if (cur >= 0 && compareAndSetState(cur, cur + 1)) {
+        ret = 1
+      } else {
+        // XXX: When JDK8 support is deprecated, onSpinWait() should be
+        //      used to avoid flooding the memory system with compareAndSetState in
+        //      a tight loop.
+        // Thread.onSpinWait()
+        ret = -1
```
bryce-anderson: Is this assignment necessary? If you want you could put the `tries` increment in this `else` branch.

bryce-anderson (approved):
Can you update the PR description to have the right number of spins?

finagle-stats-core/src/main/scala/com/twitter/finagle/stats/NonReentrantReadWriteLock.scala
geoffreyblake: Good point.  Done.

huangshiyou:
> Here are the results I got from the benchmarks:
> The scores have moved around some, especially for the `MetricsStatsReceiverBenchmark.add` workload, it improved on Arm, though it did get worse on x86. For a macro-workload such as the attached reproducer, this increase in latency did no harm. If we want to special case this for Arm, I can and rerun the micros and the macro workload to see if anything changes.
> 
> ```
> Arm64
> Benchmark                                      Mode  Cnt    Score    Error  Units
> MetricsStatsReceiverBenchmark.add              avgt    5  755.577 ± 67.337  ns/op
>                                                           583.332 ± 44.895  ns/op
> MetricsStatsReceiverBenchmark.counter          avgt    5  290.489 ± 20.528  ns/op
>                                                           269.848 ± 15.975  ns/op
> MetricsStatsReceiverBenchmark.counterAndIncr   avgt    5  406.187 ± 17.464  ns/op
>                                                           354.792 ± 10.801  ns/op
> MetricsStatsReceiverBenchmark.incr             avgt    5   10.531 ±  0.083  ns/op
>                                                            10.485 ±  0.193  ns/op
> MetricsStatsReceiverBenchmark.scope            avgt    5   98.358 ± 10.333  ns/op
>                                                           121.884 ±  0.483  ns/op
> MetricsStatsReceiverBenchmark.scopeAndCounter  avgt    5  354.088 ± 11.787  ns/op
>                                                           422.523 ± 13.759  ns/op
> MetricsStatsReceiverBenchmark.scopeAndStat     avgt    5  433.229 ±  7.257  ns/op
>                                                           414.470 ± 24.180  ns/op
> MetricsStatsReceiverBenchmark.scopedCounter    avgt    5  332.714 ±  6.423  ns/op
>                                                           324.314 ±  5.970  ns/op
> MetricsStatsReceiverBenchmark.scopedStat       avgt    5  307.627 ±  8.480  ns/op
>                                                           339.571 ± 24.469  ns/op
> MetricsStatsReceiverBenchmark.stat             avgt    5  202.166 ± 11.054  ns/op
>                                                           251.819 ± 21.994  ns/op
> MetricsStatsReceiverBenchmark.statAndAdd       avgt    5  613.260 ± 22.566  ns/op
>                                                           703.450 ± 31.770  ns/op
> ```
> 
> ```
> x86
> MetricsStatsReceiverBenchmark.add              avgt    5  186.206 ± 3.173  ns/op
>                                                           327.102 ± 1.283  ns/op
> MetricsStatsReceiverBenchmark.counter          avgt    5  128.572 ± 2.518  ns/op
>                                                           133.856 ± 3.094  ns/op
> MetricsStatsReceiverBenchmark.counterAndIncr   avgt    5  155.888 ± 1.544  ns/op
>                                                           118.834 ± 2.569  ns/op
> MetricsStatsReceiverBenchmark.incr             avgt    5   12.639 ± 0.013  ns/op
>                                                            12.653 ± 0.082  ns/op
> MetricsStatsReceiverBenchmark.scopeAndCounter  avgt    5  218.245 ± 2.924  ns/op
>                                                           209.605 ± 6.039  ns/op
> MetricsStatsReceiverBenchmark.scopeAndStat     avgt    5  210.151 ± 5.322  ns/op
>                                                           213.807 ± 4.469  ns/op  
> MetricsStatsReceiverBenchmark.scopedCounter    avgt    5  133.982 ± 6.524  ns/op
>                                                           135.596 ± 3.535  ns/op
> MetricsStatsReceiverBenchmark.scopedStat       avgt    5  130.239 ± 1.935  ns/op
>                                                           101.527 ± 2.505  ns/op
> MetricsStatsReceiverBenchmark.stat             avgt    5  123.955 ± 0.693  ns/op
>                                                           126.539 ± 2.976  ns/op
> MetricsStatsReceiverBenchmark.statAndAdd       avgt    5  255.342 ± 1.162  ns/op
>                                                           257.764 ± 0.752  ns/op
> ```

@geoffreyblake Hi, may I know the JMH results above have the changes applied or not? Thanks.

finagle-stats-core/src/main/scala/com/twitter/finagle/stats/NonReentrantReadWriteLock.scala
```diff
@@ -23,10 +36,40 @@ private[stats] final class NonReentrantReadWriteLock extends AbstractQueuedSynch
...
+    // Try to acquire the lock up to SpinMax times before failing and 
+    // allowing AbstractQueuedSynchronizer to execute the slow-path lock.
+    // This is important on Arm systems where compareAndSetState may fail
+    // more often due to its weaker memory model.
+    do {
+      cur = getState()
+      if (cur >= 0 && compareAndSetState(cur, cur + 1)) {
+        ret = 1
+      } else {
+        // XXX: When JDK8 support is deprecated, onSpinWait() should be
```
jyanJing: Mostly curious, does `XXX` have a special meaning here? If not, I would vote for removing it since we don't use this annotation elsewhere and it might confuse others when they read the comment.

finagle-stats-core/src/main/scala/com/twitter/finagle/stats/NonReentrantReadWriteLock.scala
```diff
@@ -23,10 +36,40 @@ private[stats] final class NonReentrantReadWriteLock extends AbstractQueuedSynch
...
+        //      used to avoid flooding the memory system with compareAndSetState in
+        //      a tight loop.
+        // Thread.onSpinWait()
+        tries += 1
+      }
+    } while (cur >= 0 && ret == -1 && tries < SpinMax)
```
jyanJing: We can remove the `cur >= 0` check here since it is always true, right?

finagle-stats-core/src/main/scala/com/twitter/finagle/stats/NonReentrantReadWriteLock.scala
```diff
@@ -23,10 +36,40 @@ private[stats] final class NonReentrantReadWriteLock extends AbstractQueuedSynch
   // we haven't provided an implementation for isHeldExclusively because the
   // docs for AQS advise implementors to provide implementations "as applicable"
   // and our use case doesn't require knowing when the lock is held exclusively.
+  private val SpinMax: Int = if (rwLockSpinWait() == -1) {
+    // Default number of spins for aarch64 machines set to 5 by default. 
+    // See below comment on reason.
+    if (System.getProperty("os.arch") == "aarch_64") {
```
jyanJing: ffti: we can omit the `{ }` for single line statement, see: https://docs.scala-lang.org/style/control-structures.html

finagle-stats-core/src/main/scala/com/twitter/finagle/stats/NonReentrantReadWriteLock.scala
jyanJing: On another note, since this is on a hot path, wdyt optimizing the method as in the following, with `@tailrec`, also avoid usages of mutable vars:

```
override def tryAcquireShared(num: Int): Int = {
    // Try to acquire the lock up to SpinMax times before failing and
    // allowing AbstractQueuedSynchronizer to execute the slow-path lock.
    // This is important on Arm systems where compareAndSetState may fail
    // more often due to its weaker memory model.
    tryAcquireShared(getState(), -1, 1)
  }

  @tailrec
  private[this] def tryAcquireShared(cur: Int, ret: Int, tries: Int): Int = {
    if (cur >= 0 && compareAndSetState(cur, cur + 1)) 1
    else if (tries < SpinMax) tryAcquireShared(getState(), -1, tries + 1)
    else ret
  }
```

finagle-stats-core/src/main/scala/com/twitter/finagle/stats/NonReentrantReadWriteLock.scala
```diff
@@ -1,8 +1,21 @@
 package com.twitter.finagle.stats
 
 import java.util.concurrent.locks.AbstractQueuedSynchronizer
+import java.util.Properties
+import java.lang.Thread
```
jyanJing: Would you please remove these 2 imports since they are not used?

jyanJing (commented):
Thanks @geoffreyblake for working on this! I left some comments on potential optimizations of the method. It is on a hot path, so might worth iterating on it :-)

finagle-stats-core/src/main/scala/com/twitter/finagle/stats/NonReentrantReadWriteLock.scala
geoffreyblake: No, it is not always true.  If a writer slips in, it will set cur to -1.  Therefore this check is required.

finagle-stats-core/src/main/scala/com/twitter/finagle/stats/NonReentrantReadWriteLock.scala
geoffreyblake: I was under the impression java.util.Properties was needed to access System.getProperty("os.arch"), but I'll double check. 

We can remove java.lang.Thread, but its needed for access SpinWait() when Finagle moves on from Java8.

geoffreyblake:
@huangshiyou, no I have not updated JMH, and also not clear on how to do it.  If you can point me in the correct direction, I can help here.

finagle-stats-core/src/main/scala/com/twitter/finagle/stats/NonReentrantReadWriteLock.scala
jyanJing: That's true, thanks!

finagle-stats-core/src/main/scala/com/twitter/finagle/stats/NonReentrantReadWriteLock.scala
geoffreyblake: To be honest, the entire purpose of this loop is not to be super-fast but consume some time to allow a previous lock holder to release without requiring an acquirer to synchronize via the kernel.  I'd like to to keep it as is if that is ok.

jyanJing: I see, but we are iterating the path for up to 5 times, do you think that would help with release the pervious loc holder? With the recursive solution, it also helps shoving the allocations. This part of code will be called every time we add data to the stat, so it would be great to optimize it if possible.

finagle-stats-core/src/main/scala/com/twitter/finagle/stats/NonReentrantReadWriteLock.scala
jyanJing: As an FYI, we have `fatal_warnings` turned on in pants, so unused imports will fail the build. So either you or us would need to remove it or comment it in order to merge the change.

huangshiyou:
@geoffreyblake Thanks for the reply. Any instructions on reproducing the performance issue using the example above given by you? I am testing how much performance improvement can be on my side. Apology if this is an obvious question. I am new to scala.

finagle-stats-core/src/main/scala/com/twitter/finagle/stats/NonReentrantReadWriteLock.scala
geoffreyblake: For the code in Stats, the critical sections are pretty short, and empirically on Arm, retrying allowed the system to successfully obtain the lock more often and not suffer the slow-path execution in AbstractQueuedSynchronizer.  

Testing with the micro-benchmarks and setting SpinMax to 1 had no impact on the runtime for x86.  

I'll test the suggested more elegant looking solution, though since I'm new to Scala the idea of trying to avoid explicit loops is a bit alien to me.

jyanJing: Thanks for considering it, I did a bench marks for both execution time and allocations, with 1 iteration, there is no difference in execution time, but with 5 iterations, the non-recursive solution takes twice as much as time compared to the recursive solution.
But what worries me more is the allocation, with the non-recursive solution, every time user calls `add()` on a `Stat`, it will allocation 3 integers, which can be avoided with the recursive solution.

geoffreyblake: @jyanJing , thanks for the explanation, little details like that help me understand better the concerns you had with the non-recursive solution.  I cleaned up your suggested fix and tested that it does no harm for my macro test case and the micro-benchmarks.  I think the PR is ready to be merged at this point.

jyanJing (approved):
Looks great! Thank you for working on this!

jyanJing:
Thank you for working on this @geoffreyblake , really appreciate your patience on iterating on this and the thorough benchmark testing! [The change is landed](https://github.com/twitter/finagle/commit/d45dfb022355a8105bd76d37dc77628d4e0cd330).

PR #883, opened 2020-12-07
Add fixedinet scheme resolution to global Namer

vip4f:
Problem

Name paths with fixed addresses don't need background DNS updates. But "fixedinet" scheme was not recognized by global namer and path like "/$/fixedinet/localhost/1234" was not resolved.

Solution

Add additional checks for "fixedinet" scheme in InetPath object inside global Namer.

Result

Global namer recognizes both "inet" and "fixedinet" schemes and uses corresponding Resolver.

### DIFF
--- a/finagle-core/src/main/scala/com/twitter/finagle/Namer.scala
+++ b/finagle-core/src/main/scala/com/twitter/finagle/Namer.scala
@@ -72,8 +72,8 @@ object Namer {
 
     private[this] object InetPath {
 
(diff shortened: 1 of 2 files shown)


jyanJing (approved):

jyanJing:
Thanks @vi-p4f, I successfully pulled the change and getting internal reviews.

jyanJing:
Thank you @vi-p4f for contributing, the change has [landed](https://github.com/twitter/finagle/commit/8798b92a6d693063a54d2500757c6c8d8b1c07f0)

PR #884, opened 2020-12-22
Upgrade to netty 4.1.56.Final and netty-tcnative 2.0.35.Final

peter-janssen:
Problem

There are new netty releases and the classifier for AArch64 was changed from linux-aarch64 to linux-aarch_64. This results in dependency problems with other projects.

Solution

Upgrade to netty 4.1.56.Final and netty-tcnative 2.0.35.Final for security fixes and AArch64 dependency rename.

https://netty.io/news/2020/12/17/4-1-55-Final.html
https://netty.io/news/2020/12/17/4-1-56-Final.html

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -6,8 +6,8 @@ val releaseVersion = "21.1.0-SNAPSHOT"
 
 val libthriftVersion = "0.10.0"
 
-val defaultNetty4Version = "4.1.51.Final"
-val defaultNetty4StaticSslVersion = "2.0.34.Final"
+val defaultNetty4Version = "4.1.56.Final"
+val defaultNetty4StaticSslVersion = "2.0.35.Final"
 
 val useNettySnapshot: Boolean = sys.env.get("FINAGLE_USE_NETTY_4_SNAPSHOT") match {
   case Some(useSnapshot) => useSnapshot.toBoolean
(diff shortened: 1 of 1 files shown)


build.sbt
```diff
@@ -53,7 +53,7 @@ val netty4LibsTest = Seq(
   "io.netty" % "netty-handler" % netty4Version % "test",
   "io.netty" % "netty-transport" % netty4Version % "test",
   "io.netty" % "netty-transport-native-epoll" % netty4Version % "test" classifier "linux-x86_64",
-  "io.netty" % "netty-transport-native-epoll" % netty4Version classifier "linux-aarch64",
+  "io.netty" % "netty-transport-native-epoll" % netty4Version classifier "linux-aarch_64",
```
peter-janssen: Not sure why this is not test scoped

bryce-anderson: Looks like a mistake. Do you want to do it (either in this PR or another with another being preferred) or should we just take care of it?

bryce-anderson (approved):
Thanks for the PR! To add a time expectation, we'll need to run some internal tests for this upgrade and that probably won't happen until after the holidays.

build.sbt
peter-janssen: I took care of it.

bryce-anderson:
This is getting hung up on some changes in TLS behavior. See [here](https://github.com/netty/netty/issues/10911) for more details.

bryce-anderson:
@peter-janssen, the patch to fix the ticket mentioned above should be in so we'll need to wait for another release before we can upgrade. Thanks for the PR, and once a release happens please feel free to submit another PR if we haven't already gotten around to it.

Issue #885, opened 2021-01-05
Use of Jackson 2.12.0


SerCeMan:
Hi, finagle maintainers!

Jackson has recently version [released 2.12](https://github.com/FasterXML/jackson/wiki/Jackson-Release-2.12) and finagle currently depends on 2.11.2
https://github.com/twitter/finagle/blob/c4dc4fdc5beedec031a5af25a512e5319843f02b/build.sbt#L64-L70

I was wondering if there are any know issues that could prevent finagle from using 2.12.0 or using finagle with a transitive jackson dependency on 2.12.0

yufangong:
Hi Sergey,

That's cool! Can you open an issue on GitHub propose a version bump and we
can start testing with the new version!
Thank you!

Yufan

yufangong:
Thanks, @SerCeMan. We will schedule some testing and see if we can bump the Jackson version!

SerCeMan:
Hey, @yufangong! Thanks, I created a pull request here: https://github.com/twitter/finagle/pull/886.

SerCeMan:
Just a note, a patch version, `2.12.1`, [was recently released](https://github.com/FasterXML/jackson/wiki/Jackson-Release-2.12.1).

tigerlily-he:
Hi @SerCeMan, The January 2022 release of Finagle has the Jackson upgrade to 2.13.1. https://github.com/twitter/finagle/commit/831b25120addc828b7de83ae20dfbc7c2606125a.

PR #886, opened 2021-01-06
Upgrade jackson libraries to 2.12.1

SerCeMan:
Reference: https://github.com/twitter/finagle/issues/885

FasterXML has recently version released jackson [version 2.12.0](https://github.com/FasterXML/jackson/wiki/Jackson-Release-2.12) libraries, the dependency can be upgraded.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -62,7 +62,7 @@ val netty4LibsTest = Seq(
 val netty4Http = "io.netty" % "netty-codec-http" % netty4Version
 val netty4Http2 = "io.netty" % "netty-codec-http2" % netty4Version
 val opencensusVersion = "0.24.0"
-val jacksonVersion = "2.12.0"
+val jacksonVersion = "2.12.1"
 val jacksonLibs = Seq(
   "com.fasterxml.jackson.core" % "jackson-core" % jacksonVersion,
   "com.fasterxml.jackson.core" % "jackson-databind" % jacksonVersion,
--- a/build.sbt
+++ b/build.sbt
@@ -62,7 +62,7 @@ val netty4LibsTest = Seq(
 val netty4Http = "io.netty" % "netty-codec-http" % netty4Version
 val netty4Http2 = "io.netty" % "netty-codec-http2" % netty4Version
 val opencensusVersion = "0.24.0"
-val jacksonVersion = "2.11.2"
+val jacksonVersion = "2.12.0"
 val jacksonLibs = Seq(
   "com.fasterxml.jackson.core" % "jackson-core" % jacksonVersion,
   "com.fasterxml.jackson.core" % "jackson-databind" % jacksonVersion,


yufangong:
Thank you for the update! 
We would like to keep all of our libs on the same version, those are:
util/build.sbt
finagle/build.sbt
twitter-server/build.sbt
finatra/build.sbt

~~Can you make this PR cover all of those? thanks!~~ 

Edit: sorry, I think the ask was impractical, to making a PR across repos. We have an internal ticket scheduled to upgrade the version all across, ty!

yufangong:
Unfortunately, We have run into a high-risk issue in the internal source: https://github.com/FasterXML/jackson-module-scala/issues/495.
I'm afraid we need to wait until a fix before trying to bump up, the deprecation of ScalaObjectMapper in the patched release is also concerning, so the version update won't be a quick and easy one this time.

SerCeMan:
Thanks for the update!

SerCeMan:
Thanks for the update! Just to clarify, would the above issue cause issues if finagle gets Jackson 2.12 as a transitive dependency? Or is the issue with one of the downstream consumers of Finagle and not Finagle itself?

yufangong:
Thank you for bringing this up to let it falls on our radar, @SerCeMan. I believe the concern is more for the customers, especially Finatra customers as it provides default JSON [serde](https://twitter.github.io/finatra/user-guide/json/index.html). Not only we want to keep the ecosystem on the same versions but also the issue is hard to catch without thoroughly detecting this kind of use case: serialize fields in traits and objects vs serialize fields in classes.

SerCeMan:
Hey, @yufangong! I'm just checking if the upgrade was unblocked. If I understand correctly, the referenced issue was closed. :)

yufangong:
@SerCeMan thanks so much for the reminder! We will take give it another try shortly.

SerCeMan:
Hey, @yufangong! I'm wondering if there were any new issues with the attempt, I noticed that Jackson is moving to developing the new 2.13 branch now, https://github.com/FasterXML/jackson#active-developed-jackson-2x-branches :)

tigerlily-he:
Hi @SerCeMan, I work with Yufan at Twitter. Thanks for checking in! We're planning to upgrade Jackson in September and we'll look into Jackson 2.12 and 2.13.

mosesn:
@SerCeMan update on this: we upgraded to 2.13.0 https://github.com/twitter/finagle/commit/b79b1d7e374f275e114b58d67d6e72273b97137c before discovering a [regression](https://github.com/FasterXML/jackson-databind/issues/3308) that doubled the cost of serialization in jackson-databind.  We're going to roll back for now, and will roll forward once they release 2.13.1.

SerCeMan:
Hey, @mosesn! Thank you for the update!

tigerlily-he:
Hi @SerCeMan, The January 2022 release of Finagle has the Jackson upgrade to 2.13.1. https://github.com/twitter/finagle/commit/831b25120addc828b7de83ae20dfbc7c2606125a. 

Thank you for following this journey on the Jackson upgrade! I will close this PR now.

SerCeMan:
@tigerlily-he That's awesome, thanks 🎉

PR #887, opened 2021-01-07
Bumping zookeeper version due to security issues with Netty

hllrsr:
## Problem

The most recent release from finagle is using `org.apache.zookeeper/zookeeper` on the version `3.5.0-alpha`.
This zookeeper version is using `io.netty/netty` on the version `3.7.0` which introduces several vulnerable scenarios on the code.

## Solution

The solution for this issue was bumping `zkVersion` on the project, to an compatible version of zookeeper that is using netty in a newer version (most precisely, zookeeper was bumped to the version `3.5.6`).

## Result

The result of this pull-request will be finagle be free from this vulnerable scenarios from this outdated dependency.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -33,7 +33,7 @@ val netty4StaticSslVersion: String =
 val nettyVersionInfo = settingKey[String]("A setting reference for printing the netty version info")
 
 // zkVersion should be kept in sync with the 'util-zk' dependency version
-val zkVersion = "3.5.0-alpha"
+val zkVersion = "3.5.6"
 
 val scalaCollectionCompat = "org.scala-lang.modules" %% "scala-collection-compat" % "2.1.2"
 val caffeineLib = "com.github.ben-manes.caffeine" % "caffeine" % "2.8.5"


mosesn (approved):
looks good to me, thanks @hllrsr!  We'll have to do some internal testing of this too, but I think it should be good to go.

yufangong (approved):

hllrsr:
@mosesn no worries! I'm ok with merging this change on the develop branch.

build.sbt
```diff
@@ -33,7 +33,7 @@ val netty4StaticSslVersion: String =
 val nettyVersionInfo = settingKey[String]("A setting reference for printing the netty version info")
 
 // zkVersion should be kept in sync with the 'util-zk' dependency version
-val zkVersion = "3.5.0-alpha"
+val zkVersion = "3.5.6"
```
yufangong, line 36: Could you also bump the util-zk version in the dependency? https://github.com/twitter/util
thanks!

yufangong: Hey, don't worry about this, we update it along with this PR. We plan to merge this next week asap, thank you!

hllrsr: Thanks!

PR #888, opened 2021-02-03
finagle-serversets: Handle KeeperState.Closed ZooKeeper session state

hmhagberg:
**Problem**

Upon closing ZooKeeper connection, ClientCnxn dispatches KeeperState.Closed
event to watchers. Because there is no mapping from KeeperState.Closed to
SessionState, closing ZooKeeper connection results in a NoSuchElementException
in ClientCnxn event thread. The exception is caught and logged.

Existing tests in ZkSessionEndToEndTest reproduce the issue. An example how the issue looks like
```
DEBUG - Disconnecting client for session: 0x1000180f0f40000
DEBUG - An exception was thrown while closing send thread for session 0x1000180f0f40000 : Unable to read additional data from server sessionid 0x1000180f0f40000, likely server has closed socket
ERROR - Error while calling watcher 
java.util.NoSuchElementException: key not found: Closed
	at scala.collection.MapLike.default(MapLike.scala:235)
	at scala.collection.MapLike.default$(MapLike.scala:234)
	at scala.collection.AbstractMap.default(Map.scala:65)
	at scala.collection.MapLike.apply(MapLike.scala:144)
	at scala.collection.MapLike.apply$(MapLike.scala:143)
	at scala.collection.AbstractMap.apply(Map.scala:65)
	at com.twitter.finagle.serverset2.client.apache.ApacheSessionState$.apply(types.scala:44)
	at com.twitter.finagle.serverset2.client.apache.ApacheWatcher.process(ApacheWatcher.scala:18)
	at org.apache.zookeeper.ClientCnxn$EventThread.processEvent(ClientCnxn.java:535)
	at org.apache.zookeeper.ClientCnxn$EventThread.run(ClientCnxn.java:510)
INFO - Session: 0x1000180f0f40000 closed
INFO - EventThread shut down for session: 0x1000180f0f40000
```

**Solution**

Add mapping from KeeperState.Closed to SessionState.Closed and handle the new
state in relevant places.

### DIFF
--- a/doc/src/sphinx/metrics/ServiceDiscovery.rst
+++ b/doc/src/sphinx/metrics/ServiceDiscovery.rst
@@ -171,3 +171,6 @@ Under the \`zkclient\` scope
 
 **session_expired**
   A counter of the number of session expirations.
+
+**session_closed**
+  A counter of the number of closed sessions.
--- a/finagle-serversets/src/main/scala/com/twitter/finagle/serverset2/ServiceDiscoverer.scala
+++ b/finagle-serversets/src/main/scala/com/twitter/finagle/serverset2/ServiceDiscoverer.scala
@@ -38,7 +38,7 @@ private[serverset2] object ServiceDiscoverer {
         case SessionState.Unknown =>
           Unknown
         case SessionState.Expired | SessionState.NoSyncConnected | SessionState.AuthFailed |
-            SessionState.Disconnected =>
+            SessionState.Disconnected | SessionState.Closed =>
           Unhealthy
         case SessionState.ConnectedReadOnly | SessionState.SaslAuthenticated |
             SessionState.SyncConnected =>
--- a/finagle-serversets/src/main/scala/com/twitter/finagle/serverset2/ZkSession.scala
+++ b/finagle-serversets/src/main/scala/com/twitter/finagle/serverset2/ZkSession.scala
@@ -168,6 +168,8 @@ private[serverset2] class ZkSession(
                 u() = Activity.Failed(new Exception("" + sessionState))
               // Do NOT keep retrying, wait to be reconnected automatically by the underlying session
 
+              case WatchState.SessionState(SessionState.Closed) =>
+
               case WatchState.SessionState(sessionState) =>
                 logger.error(s"Unexpected session state $sessionState. Session: $sessionIdAsHex")
                 u() = Activity.Failed(new Exception("" + sessionState))


jyanJing:
Thanks @hmhagberg for making the change, your approach looks good to me, one thing I am not sure is if we should add a new session state as `Closed` or reuse the existing `Disconnected`, one question I have: is the `KeeperState.Closed` something new? I am curious about why this problem hasn't occurred before, it looks like we missed a very common case and we should see the exceptions a lot. And if this state is new, then yes, we should introduce a new mapping session state as `Closed`.

jyanJing:
Hi @hmhagberg , it turns out that @enbnt already fixed this in [9f6bc4bca9adb83b7d1db6894243470ba0c3d7eb](https://github.com/twitter/finagle/commit/9f6bc4bca9adb83b7d1db6894243470ba0c3d7eb), if you rebase master, the issue should be fixed! Thanks @enbnt for working on it!

hmhagberg:
Good timing! I didn't notice it had been just fixed before submitting this PR.

However, I believe the existing fix has a small issue. `ZkSession#watchedOperation` does not handle the newly added `SessionState.Closed`. This means that when `ZkSession` is closed, any ongoing watch operation activity (`globOf`/`existsOf`) ends up in failed state and "Unexpected session state" is logged. I modified `ZkSession.close` test to demonstrate this (it requires using `globOf`/`existsOf` before closing the session).

hmhagberg:
> Thanks @hmhagberg for making the change, your approach looks good to me, one thing I am not sure is if we should add a new session state as `Closed` or reuse the existing `Disconnected`, one question I have: is the `KeeperState.Closed` something new? I am curious about why this problem hasn't occurred before, it looks like we missed a very common case and we should see the exceptions a lot. And if this state is new, then yes, we should introduce a new mapping session state as `Closed`.

Also to answer this, `Closed` state was introduced in ZooKeeper 3.5.5 (commit https://github.com/apache/zookeeper/commit/6748a0e3f58f2a398dec4c6988bc70ea4363b807). It is generated client-side, when explicitly closing ZooKeeper session, so it is definitely separate from `Disconnected` state. I can't say why it was missed (we see the exceptions occasionally) but as far as I can tell, it is mostly safe to ignore.

jyanJing:
> Good timing! I didn't notice it had been just fixed before submitting this PR.
> 
> However, I believe the existing fix has a small issue. `ZkSession#watchedOperation` does not handle the newly added `SessionState.Closed`. This means that when `ZkSession` is closed, any ongoing watch operation activity (`globOf`/`existsOf`) ends up in failed state and "Unexpected session state" is logged. I modified `ZkSession.close` test to demonstrate this (it requires using `globOf`/`existsOf` before closing the session).

Thanks for pointing this out, so one problem is that our internal Zookeeper version is not aligned with the oss one, let me pull your change and run some test!

finagle-serversets/src/test/scala/com/twitter/finagle/serverset2/ZkSessionEndToEndTest.scala
```diff
@@ -27,7 +27,7 @@ class ZkSessionEndToEndTest extends FunSuite with BeforeAndAfter {
   def toSpan(d: Duration): Span = Span(d.inNanoseconds, Nanoseconds)
 
   implicit val patienceConfig =
-    PatienceConfig(timeout = toSpan(1.second), interval = toSpan(zkTimeout))
+    PatienceConfig(timeout = toSpan(5.second), interval = toSpan(zkTimeout))
```
jyanJing: Any reasons to increase the timeout?

doc/src/sphinx/metrics/ServiceDiscovery.rst
```diff
@@ -171,3 +171,6 @@ Under the \`zkclient\` scope
 
 **session_expired**
   A counter of the number of session expirations.
+
+**session_closed**
```
jyanJing, line 175: Thank you for adding this!

jyanJing (commented):
Hi @hmhagberg ,

The change looks good to me overall. 

It turns out that our internal ZK version is behind so it won't be able to pick up the new `KeeperState.closed`, the change landed earlier has a [workaround](https://github.com/twitter/finagle/blob/9f6bc4bca9adb83b7d1db6894243470ba0c3d7eb/finagle-serversets/src/main/scala/com/twitter/finagle/serverset2/client/apache/types.scala#L39)  which is to map any unrecognized `KeeperState` to an empty state. So the test you added won't pass internally due to we are unable to return a `SessionState.closed`. But it wouldn't harm to add this state mapping to the main classes. So I would vote for keeping the changes in main but only keeping the test change on your end for now.

Thanks again for working on this :-)

hmhagberg:
Makes sense, not including the test is fine for me.

I reverted the test changes and also modified the commit message to better reflect what is actually being done.

finagle-serversets/src/test/scala/com/twitter/finagle/serverset2/ZkSessionEndToEndTest.scala
hmhagberg: The tests (also existing ones) were flaky for me and increasing timeout helped. I ran them now again a couple of times without any issues so I'm not sure what the failures were about. In any case I reverted all test changes.

jyanJing:
Thanks @hmhagberg, I will pull your change and get reviews today!

jyanJing:
The change has [landed](https://github.com/twitter/finagle/commit/4ae0cf3cc6bb150f5a842187d4a3e2755a10fbe0), thanks again for working on this @hmhagberg !

hmhagberg:
Cheers! Thanks for helping with this

Issue #909, opened 2021-07-19
`CookieMap#getAll` does not return all values for a given cookie name


guilgaly:
**Describe the bug**
The Scaladoc for `com.twitter.finagle.http.CookieMap#getAll` says that it "Fetches all cookies with the given name from this map.". However it appears to always return only the first value for a cookie name when handling a request and accessing the request cookies.

**To Reproduce**
1. Send an HTTP request with a cookie header containing different values for the same name (from a web browser, this will happen if the same cookie name was set with different paths, some more specific than others); e.g. `Cookie: FOO=value1; FOO=value2; FOO=value3`.
2. In the server code, retrieve the cookie values with `request.cookies.getAll("FOO")`; it will return `Seq("value1")` instead of `Seq("value1", "value2", "value3")`.

**Expected behavior**
I think `CookieMap#getAll` would be expected to return all values for a given name when several values are given in the  request cookies.

**Environment**
For the cookie header format: tested with Chrome 91 on macOS 10.15.7. 
For Finagle: tested with Finagle 20.3.0 (comes with Finch 0.32.1). Looking at the code for `com.twitter.finagle.http.CookieMap` and `com.twitter.finagle.http.netty4.Netty4CookieCodec` in the `develop` branch, it should still be the same.

**Additional context**
I see two causes for this:
1. `com.twitter.finagle.http.netty4.Netty4CookieCodec#decodeServer` calls `io.netty.handler.codec.http.cookie.ServerCookieDecoder#decode` rather than `io.netty.handler.codec.http.cookie.ServerCookieDecoder#decodeAll`. According to the Netty Javadoc, `decodeAll` "includes all cookie values present, even if they have the same name", unlike `decode`.
2. If we change 1., then `com.twitter.finagle.http.CookieMap#addNoRewrite` only keeps the last value added for a given name (that's because of how equality works for `com.twitter.finagle.http.Cookie`; the cookies would only differ by their paths, and we of course cannot know the paths for the request cookies on the server side).

I'd be happy to try and submit a pull request, if this is in fact not the expected behavior.

yufangong:
Hi @guilgaly, I agree with you this smells like a bug. Pull requests are welcomed! thanks!

In terms of 2, can you explain a bit more of what would be the behavior change and how it would affect the use of `com.twitter.finagle.http.CookieMap#add`? Thanks!

guilgaly:
[...]

PR #911, opened 2021-08-13
finagle-netty4-http: allow setting multiple values for a request cookie

guilgaly:
See issue #909.

Problem:
In a Response, `CookieMap` allows setting multiple cookie values for a given name as long as they differ by their domain or path. However, in a Request, if the client sent several values for a given cookie name, only the last one is currently available.

Solution:
Fix a bug in `Netty4CookieCodec` when decoding request cookies, and change `CookieMap` behavior to allow inserting multiple values for a request cookie. Also preserve insertion order for the different values for a given cookie name, to preserve the order in which the client sent the values in a request.

### DIFF
--- a/finagle-base-http/src/main/scala/com/twitter/finagle/http/CookieMap.scala
+++ b/finagle-base-http/src/main/scala/com/twitter/finagle/http/CookieMap.scala
@@ -83,7 +83,7 @@ class CookieMap private[finagle] (message: Message, cookieCodec: CookieCodec)
     def hasNext: Boolean = outer.hasNext || !inner.isEmpty
     def next(): (String, Cookie) = {
       if (inner.isEmpty) {
-        inner = outer.next()._2
+        inner = outer.next()._2.reverse
       }
 
       val result = inner.head
@@ -103,13 +103,13 @@ class CookieMap private[finagle] (message: Message, cookieCodec: CookieCodec)
   def getValue(name: String): Option[String] = get(name).map(_.value)
 
   /**
-   * Fetches all cookies with the given `name` from this map.
+   * Fetches all cookies with the given `name` from this map, in order of insertion.
    */
-  def getAll(name: String): Seq[Cookie] = underlying(name)
+  def getAll(name: String): Seq[Cookie] = underlying(name).reverse
 
   /**
    * Adds the given `cookie` (which is a tuple of cookie `name`
-   * and Cookie` itself) into this map. If there are already cookies
+   * and `Cookie` itself) into this map. If there are already cookies
    * with the given `name` in the map, they will be removed.
    */
   protected def addCookie(cookie: (String, Cookie)): this.type = {
@@ -127,6 +127,11 @@ class CookieMap private[finagle] (message: Message, cookieCodec: CookieCodec)
     this += ((cookie.name, cookie))
   }
 
+  /**
+   * Adds the given `cookies` (which are tuples of cookie `name`
+   * and `Cookie` itself) into this map. If there are already cookies
+   * with the given `name` in the map, they will be removed.
+   */
   protected def addCookies(
     cookies: scala.collection.TraversableOnce[(String, Cookie)]
   ): this.type = {
@@ -157,17 +162,21 @@ class CookieMap private[finagle] (message: Message, cookieCodec: CookieCodec)
   private[this] def addNoRewrite(name: String, cookie: Cookie): Unit = {
     val prev = underlying(name)
     val next =
-      if (prev.contains(cookie)) cookie :: prev.filter(c => c != cookie)
+      if (message.isResponse && prev.contains(cookie)) cookie :: prev.filter(c => c != cookie)
       else cookie :: prev
 
     underlying(name) = next
   }
 
   /**
-   * Adds the given `cookie` with `name` into this map. Existing cookies
-   * with this name but different domain/path will be kept. If there is already
-   * an identical cookie (different value but name/path/domain is the same) in the
-   * map, it will be replaced within a new version.
+   * Adds the given `cookie` with `name` into this map.
+   *
+   * On a Response: existing cookies with this name but different domain/path
+   * will be kept. If there is already an identical cookie (different value but
+   * name/path/domain is the same) in the map, it will be replaced within a new
+   * version.
+   *
+   * On a Request: existing cookies with this name will be kept.
    *
    * @see [[addAll]] for adding cookies in bulk
    */
@@ -177,10 +186,14 @@ class CookieMap private[finagle] (message: Message, cookieCodec: CookieCodec)
   }
 
   /**
-   * Adds the given `cookie` into this map. Existing cookies with this name
-   * but different domain/path will be kept. If there is already an identical
-   * cookie (different value but name/path/domain is the same) in the map,
-   * it will be replaced within a new version.
+   * Adds the given `cookie` into this map.
+   *
+   * On a Response: existing cookies with this name but different domain/path
+   * will be kept. If there is already an identical cookie (different value but
+   * name/path/domain is the same) in the map, it will be replaced within a new
+   * version.
+   *
+   * On a Request: existing cookies with this name will be kept.
    *
    * @see [[addAll]] for adding cookies in bulk
    */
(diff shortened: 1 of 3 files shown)


yufangong:
@guilgaly Thanks for the PR! It looks like there are two tests failing, can you check it out? Thanks!

guilgaly:
@yufangong it should be fixed now!

The order of iteration on the CookieMap can be different between Scala 2.12 and 2.13 because of changes in the underlying collection, and I broke these tests by no longer accepting either order - so they only passed on Scala 2.13. (My changes should guarantee that we iterate in order of insertion for cookies with the same name, but there is still no guarantee for cookies with different names.)

finagle-base-http/src/main/scala/com/twitter/finagle/http/CookieMap.scala
```diff
@@ -103,13 +103,13 @@ class CookieMap private[finagle] (message: Message, cookieCodec: CookieCodec)
   def getValue(name: String): Option[String] = get(name).map(_.value)
 
   /**
-   * Fetches all cookies with the given `name` from this map.
+   * Fetches all cookies with the given `name` from this map, in order of insertion.
```
mosesn: suggestion:
from the RFC:
https://datatracker.ietf.org/doc/html/rfc6265#section-4.2.2
> servers SHOULD NOT rely upon the order in which these cookies appear in the header

we shouldn't make the ordering part of the contract, imo

finagle-base-http/src/main/scala/com/twitter/finagle/http/CookieMap.scala
```diff
@@ -157,17 +162,21 @@ class CookieMap private[finagle] (message: Message, cookieCodec: CookieCodec)
...
   }
 
   /**
-   * Adds the given `cookie` with `name` into this map. Existing cookies
-   * with this name but different domain/path will be kept. If there is already
-   * an identical cookie (different value but name/path/domain is the same) in the
-   * map, it will be replaced within a new version.
+   * Adds the given `cookie` with `name` into this map.
+   *
+   * On a Response: existing cookies with this name but different domain/path
```
mosesn: issue: this response vs request behavior is strange to me.  what's the motivation behind it?  I don't see anything in the RFC that would encourage it.

finagle-base-http/src/main/scala/com/twitter/finagle/http/CookieMap.scala
```diff
@@ -157,17 +162,21 @@ class CookieMap private[finagle] (message: Message, cookieCodec: CookieCodec)
   private[this] def addNoRewrite(name: String, cookie: Cookie): Unit = {
     val prev = underlying(name)
     val next =
-      if (prev.contains(cookie)) cookie :: prev.filter(c => c != cookie)
+      if (message.isResponse && prev.contains(cookie)) cookie :: prev.filter(c => c != cookie)
```
mosesn, line 165: question: we still want to prevent illegal duplicative cookies (eg ones with the same path, domain, and name), right?  I'm a bit worried that we might be overreaching here.  can we ensure that we continue to filter those out?

mosesn (commented):
Thanks for the PR!  I think there are some details in the semantics that we should work out before merging this in.

tigerlily-he:
Hi @guilgaly, Friendly ping. How is this PR going? Moses left some questions and comments in a review.

finagle-base-http/src/main/scala/com/twitter/finagle/http/CookieMap.scala
guilgaly: In a request, the cookie header does not provide any information on the domain/path for the cookie, only the cookie name and value; yet the cookie header can contain multiple values for the same name (presumably because they were set with different domains/paths that can both match for this request, but we don't have this information when receiving the request on the server side).

finagle-base-http/src/main/scala/com/twitter/finagle/http/CookieMap.scala
guilgaly: The way I see it:
- The RFC does also recommend that the client use a specific order in https://datatracker.ietf.org/doc/html/rfc6265#section-5.4, although that's only a "should", not a "must" (hence why it shouldn't be relied upon). Major browsers do respect that AFAIK.
- The recommendation not to rely on the order is, again only a "should not", not a "must not"; up to you to see if the HTTP library should be opinionated on that, I guess?
- I think it would be a bit surprising to have a `Seq` (rather than e.g. a `Set`) but with values in a potentially "random" order, different from the order we see when looking at the actual HTTP request.
- There may also be a question about the behavior of `get()`/`getValue()` (which use `headOption`) when several values are present. Always taking the first value in the list as sent by the client seems reasonable to me?

guilgaly:
@tigerlily-he thanks for the ping! I must have missed the notification at the time, and after that I was mostly offline for a while and forgot to check...

finagle-base-http/src/main/scala/com/twitter/finagle/http/CookieMap.scala
mosesn: > The recommendation not to rely on the order is, again only a "should not", not a "must not"; up to you to see if the HTTP library should be opinionated on that, I guess?

I don't think we need to be opinionated about it, and for that reason, I think it's nice that the implementation preserves order, but also that we don't put into the contract that it does.  Servers can still be well-behaved and ignore the order, and it's better for people to program as if that's the case.  If people want to program in a way that's explicitly discouraged by the spec they can for now, but support for them may be dropped in the future.

> There may also be a question about the behavior of get()/getValue() (which use headOption) when several values are present. Always taking the first value in the list as sent by the client seems reasonable to me?

I agree it's reasonable, but I don't think that we should make it part of the contract.

With that said, I don't feel super strongly about this.  It's just my opinion.  I'm OK with shipping it even if you decide not to change it!

finagle-base-http/src/main/scala/com/twitter/finagle/http/CookieMap.scala
mosesn: OK, that makes sense to me, thanks!  It might be good to explain that in the scaladoc.

guilgaly: I'm going to try to clarify the Scaladoc then 👍

finagle-base-http/src/main/scala/com/twitter/finagle/http/CookieMap.scala
guilgaly: I can drop it from the stated contract in the Scaladoc, no problem with that for me.

guilgaly:
@mosesn I modified the Scaladoc per your comments, let me know what you think. 🙂

finagle-base-http/src/main/scala/com/twitter/finagle/http/CookieMap.scala
hamdiallam: +1

hamdiallam:
Hey! @guilgaly The PR looks good to me.

Wanted to surface Moses' question. Is there benefit to adding `message.isResponse` conditional check when adding a cookie? While it's right the on the request path there is no domain/path, is there any benefit to having multiple cookies with the same name on the request? It seems to be that it'd be beneficial to avoid duplicates on both the request and response. On the server side it's ensured that this duplication wouldn't occur, so it makes sense to simply be universal client-side as well

mosesn (approved):
Thanks!!

finagle-base-http/src/main/scala/com/twitter/finagle/http/CookieMap.scala
joybestourous: @guilgaly hey, just bumping this question (and @hamdiallam's related question from a few days ago)

dotordogh: @guilgaly bumping this once more :)

mosesn:
@guilgaly have you had a chance to look at @hamdiallam's question?  We should be good to merge it in after you have a chance to address it!

tigerlily-he:
Hi @guilgaly, We'd like to merge this PR after you address this question. 

> Hey! @guilgaly The PR looks good to me.
> 
> Wanted to surface Moses' question. Is there benefit to adding `message.isResponse` conditional check when adding a cookie? While it's right the on the request path there is no domain/path, is there any benefit to having multiple cookies with the same name on the request? It seems to be that it'd be beneficial to avoid duplicates on both the request and response. On the server side it's ensured that this duplication wouldn't occur, so it makes sense to simply be universal client-side as well

Issue #918, opened 2021-10-05
Use canonical host name for per host stats


fuCtor:
We use DNS balancing in our project. For investigation of behaviour (request, latency, etc) per host we configure this feature. But in stats we see only one metric with given hostname. As result result all stats are collapsed in single branch.

As solution I suggest use canonical host name. Canonical hostname are calculated as reverse lookup by IP address. As result in stats we receive name of definite endpoint. For behaviour switching are using flag.

As workaround we use resolver, which transform result from InetResolver.

PR #919, opened 2021-10-05
finagle-core: canonical name for per host stats

fuCtor:
Issue #918

Problem:
With DNS balancing per host stats are collapsing in single branch.

Solution:
Use canonical host name in stats receiver scope.

### DIFF
--- a/finagle-core/src/main/scala/com/twitter/finagle/loadbalancer/flags.scala
+++ b/finagle-core/src/main/scala/com/twitter/finagle/loadbalancer/flags.scala
@@ -41,14 +41,6 @@ object perHostStats
         "\tor the NullStatsReceiver if none given."
     )
 
-object useCanonicalHostname
-  extends GlobalFlag[Boolean](
-    false,
-    "enable/default using canonical host name in stats.\n" +
-      "\tWhen enabled, canonical host name for endpoint will be used.\n" +
-      "\tWhen disabled, given hostname will be used."
-  )
-
 package exp {
 
   import com.twitter.finagle.loadbalancer.aperture.EagerConnectionsType
--- a/finagle-core/src/main/scala/com/twitter/finagle/loadbalancer/flags.scala
+++ b/finagle-core/src/main/scala/com/twitter/finagle/loadbalancer/flags.scala
@@ -41,6 +41,14 @@ object perHostStats
         "\tor the NullStatsReceiver if none given."
     )
 
+object useCanonicalHostname
+  extends GlobalFlag[Boolean](
+    false,
+    "enable/default using canonical host name in stats.\n" +
+      "\tWhen enabled, canonical host name for endpoint will be used.\n" +
+      "\tWhen disabled, given hostname will be used."
+  )
+
 package exp {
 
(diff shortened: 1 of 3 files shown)


finagle-core/src/test/scala/com/twitter/finagle/loadbalancer/LoadBalancerFactoryTest.scala
```diff
@@ -63,17 +63,75 @@ class LoadBalancerFactoryTest extends AnyFunSuite with Eventually with Integrati
...
+    }
+  }
+
+  test("reports with given name when flag is false") {
+    new PerHostFlagCtx {
+      val sr = new InMemoryHostStatsReceiver
+      val sr1 = new InMemoryStatsReceiver
+
+      val nonCanonicalPort: String = "github.com:443"
+      val canonicalPerHostStatKey = Seq(label, nonCanonicalPort, "available")
```
yufangong: suggestion(nit): Let's rename it to nonCanonical

finagle-core/src/test/scala/com/twitter/finagle/loadbalancer/LoadBalancerFactoryTest.scala
```diff
@@ -40,14 +40,14 @@ class LoadBalancerFactoryTest extends AnyFunSuite with Eventually with Integrati
       perHostStats.let(true) {
         client
           .configured(LoadBalancerFactory.HostStats(sr))
-          .newService(port)
+          .newService(port)("test")
```
yufangong, line 43: question: If we keep the perHostFlag as App, we should not need to initiate requests here and below, why do you prefer this way?

yufangong (commented):
Thanks for the PR. While waiting for the tests to complete, I have some nits and questions.

finagle-core/src/test/scala/com/twitter/finagle/loadbalancer/LoadBalancerFactoryTest.scala
fuCtor: Maybe I launch test wrong, but on exactly fail test result always successful. When I made this changes, test has showing correct result.

finagle-core/src/main/scala/com/twitter/finagle/loadbalancer/flags.scala
```diff
@@ -41,6 +41,14 @@ object perHostStats
         "\tor the NullStatsReceiver if none given."
     )
 
+object useCanonicalHostname
```
hamdiallam: question: any reason we want to use a flag instead of stack param. perhaps an `AddressFormatter` stack param that can be invoked on an address?

fuCtor: It's a good idea. Maybe some body want to use IP+Port instead of any host name and would define custom formatter.

yufangong (approved):

yufangong:
Merged in https://github.com/twitter/finagle/commit/cde98b358a92d8dd722767fd08c6e38ba00bd37b
Thanks for the PR!

PR #923, opened 2022-01-21
Add MinSendBackupAfterMs to config in BackupRequestFilter

### LINKED ISSUE #905: Make MinSendBackupAfterMs of the BackupRequestFilter configurable
The MinSendBackupAfterMs is now set to 1 ms. Unfortunately this causes backup requests to be send when there is low load. It would be useful if we could set the minimum a little bit higher for certain endpoints.

emilhotkowski:
Problem

The MinSendBackupAfterMs is now set to 1 ms. Unfortunately this causes backup requests to be send when there is low load.

Solution

Added **MinSendBackupAfterMs** to the stack param Configured in BackupRequestFilter and propagated changes to `MethodBuilder` by adding new versions of `idempotent` function.

Result

Developers will be able to set their own value of **MinSendBackupAfterMs** if necessary.

Closes twitter/finagle#905

### DIFF
--- a/finagle-core/src/main/scala/com/twitter/finagle/client/BackupRequestFilter.scala
+++ b/finagle-core/src/main/scala/com/twitter/finagle/client/BackupRequestFilter.scala
@@ -41,16 +41,6 @@ object BackupRequestFilter {
   private val LostAnnotation =
     Annotation.Message("Client Backup Request Lost")
 
-  /**
-   * Use a minimum non-zero delay to prevent sending unnecessary backup requests
-   * immediately for services where the latency at the percentile where a backup will be sent is
-   * ~0ms. This is preferable to not sending any backups in the aforementioned case; by sending
-   * a backup after 1ms we can still reduce the higher latencies at greater latency percentiles.
-   * For example, if p99 latency is 0 and we are configured to send backups at the p99 latency,
-   * we can a reduce p999 latency of 10 ms to close to 1ms.
-   */
-  private val MinSendBackupAfterMs: Int = 1
-
   private[finagle] val SupersededRequestFailure =
     Failure.ignorable("Request was superseded by another in BackupRequestFilter")
 
@@ -86,7 +76,7 @@ object BackupRequestFilter {
 
   object Param {
 
-    private[client] case class Configured(maxExtraLoad: Tunable[Double], sendInterrupts: Boolean)
+    private[client] case class Configured(maxExtraLoad: Tunable[Double], sendInterrupts: Boolean, minSendBackupAfterMs: Int = 1)
         extends Param
     case object Disabled extends Param
     implicit val param: Stack.Param[BackupRequestFilter.Param] = Stack.Param(Disabled)
@@ -127,7 +117,23 @@ object BackupRequestFilter {
    *                       is returned and the result of the outstanding request is superseded. For
    *                       protocols without a control plane, where the connection is cut on
    *                       interrupts, this should be "false" to avoid connection churn.
+   *
+   * @param minSendBackupAfterMs Use a minimum non-zero delay to prevent sending unnecessary backup requests
+   *                             immediately for services where the latency at the percentile where a
+   *                             backup will be sent is ~0ms.
    */
+  def Configured(maxExtraLoad: Double, sendInterrupts: Boolean, minSendBackupAfterMs: Int): Param = {
+    require(
+      maxExtraLoad >= 0 && maxExtraLoad < 1.0,
+      s"maxExtraLoad must be between 0.0 and 1.0, was $maxExtraLoad"
+    )
+    require(
+      minSendBackupAfterMs >= 1,
+      s"minSendBackupAfterMs must be greater or equal to 1ms, was $minSendBackupAfterMs"
+    )
+    Param.Configured(Tunable.const(role.name, maxExtraLoad), sendInterrupts, minSendBackupAfterMs)
+  }
+
   def Configured(maxExtraLoad: Double, sendInterrupts: Boolean): Param = {
     require(
       maxExtraLoad >= 0 && maxExtraLoad < 1.0,
@@ -136,17 +142,22 @@ object BackupRequestFilter {
     Param.Configured(Tunable.const(role.name, maxExtraLoad), sendInterrupts)
   }
 
+  def Configured(maxExtraLoad: Tunable[Double], sendInterrupts: Boolean, minSendBackupAfterMs: Int): Param =
+    Param.Configured(maxExtraLoad, sendInterrupts, minSendBackupAfterMs)
+
   def Configured(maxExtraLoad: Tunable[Double], sendInterrupts: Boolean): Param =
     Param.Configured(maxExtraLoad, sendInterrupts)
 
   private[this] def mkFilterFromParams[Req, Rep](
     maxExtraLoad: Tunable[Double],
     sendInterrupts: Boolean,
+    minSendBackupAfterMs: Int,
     params: Stack.Params
   ): BackupRequestFilter[Req, Rep] =
     new BackupRequestFilter[Req, Rep](
       maxExtraLoad,
       sendInterrupts,
+      minSendBackupAfterMs,
       params[param.ResponseClassifier].responseClassifier,
       params[Retries.Budget].retryBudget,
       params[Histogram].lowestDiscernibleMsValue,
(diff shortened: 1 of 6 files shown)


tigerlily-he:
First, can you add a PR description that follows this format? https://github.com/twitter/finagle/blob/develop/.github/PULL_REQUEST_TEMPLATE/pull_request_template.md

This will become the commit message when it is merged.

emilhotkowski:
@tigerlily-he done!

finagle-core/src/main/scala/com/twitter/finagle/client/BackupRequestFilter.scala
```diff
@@ -41,16 +41,6 @@ object BackupRequestFilter {
   private val LostAnnotation =
     Annotation.Message("Client Backup Request Lost")
 
-  /**
-   * Use a minimum non-zero delay to prevent sending unnecessary backup requests
-   * immediately for services where the latency at the percentile where a backup will be sent is
-   * ~0ms. This is preferable to not sending any backups in the aforementioned case; by sending
-   * a backup after 1ms we can still reduce the higher latencies at greater latency percentiles.
-   * For example, if p99 latency is 0 and we are configured to send backups at the p99 latency,
-   * we can a reduce p999 latency of 10 ms to close to 1ms.
```
jyanJing, line 50: nit: s/we can a/we can/

finagle-core/src/main/scala/com/twitter/finagle/client/BackupRequestFilter.scala
```diff
@@ -41,16 +41,6 @@ object BackupRequestFilter {
...
-   */
-  private val MinSendBackupAfterMs: Int = 1
```
jyanJing, line 52: Can you specify the unit of this parameter in the comment? It sounds like this should be milliseconds in the comment and inferred from the name, but it is good to mention that explicitly.

finagle-core/src/main/scala/com/twitter/finagle/client/BackupRequestFilter.scala
```diff
@@ -127,7 +117,23 @@ object BackupRequestFilter {
    *                       is returned and the result of the outstanding request is superseded. For
    *                       protocols without a control plane, where the connection is cut on
    *                       interrupts, this should be "false" to avoid connection churn.
+   *
+   * @param minSendBackupAfterMs Use a minimum non-zero delay to prevent sending unnecessary backup requests
```
jyanJing, line 121: dido: it would be great to mention the unit of the parameter unit explicitly, especially this becomes a public parameter now :-)

finagle-core/src/main/scala/com/twitter/finagle/client/BackupRequestFilter.scala
```diff
@@ -271,6 +282,13 @@ private[client] class BackupRequestFactory[Req, Rep](
  *                       protocols without a control plane, where the connection is cut on
  *                       interrupts, this should be "false" to avoid connection churn.
  *
+ * @param minSendBackupAfterMs Use a minimum non-zero delay to prevent sending unnecessary backup requests
```
jyanJing, line 285: dido: specify unit explicitly.

finagle-core/src/main/scala/com/twitter/finagle/client/MethodBuilder.scala
```diff
@@ -229,6 +229,9 @@ final class MethodBuilder[Req, Rep] private[finagle] (
    *                       is returned and the result of the outstanding request is superseded. For
    *                       protocols without a control plane, where the connection is cut on
    *                       interrupts, this should be "false" to avoid connection churn.
+   * @param minSendBackupAfterMs Use a minimum non-zero delay to prevent sending unnecessary backup requests
```
jyanJing, line 232: dido

finagle-core/src/main/scala/com/twitter/finagle/client/MethodBuilder.scala
```diff
@@ -237,6 +240,18 @@ final class MethodBuilder[Req, Rep] private[finagle] (
...
+    minSendBackupAfterMs: Int,
+    classifier: service.ResponseClassifier
+  ): MethodBuilder[Req, Rep] = {
+    val brfParam =
+      if (maxExtraLoad == 0) BackupRequestFilter.Disabled
+      else BackupRequestFilter.Configured(maxExtraLoad, sendInterrupts, minSendBackupAfterMs)
+    addBackupRequestFilterParamAndClassifier(brfParam, classifier)
+  }
+
   def idempotent(
```
jyanJing, line 255: since we are already here, would you mind adding the comments to the existing `idempotent` method, too? Thank you!
Also we could have one method call another here, too, to avoid duplicate code.

finagle-core/src/main/scala/com/twitter/finagle/client/BackupRequestFilter.scala
```diff
@@ -127,7 +117,23 @@ object BackupRequestFilter {
...
+      s"maxExtraLoad must be between 0.0 and 1.0, was $maxExtraLoad"
+    )
+    require(
+      minSendBackupAfterMs >= 1,
+      s"minSendBackupAfterMs must be greater or equal to 1ms, was $minSendBackupAfterMs"
+    )
+    Param.Configured(Tunable.const(role.name, maxExtraLoad), sendInterrupts, minSendBackupAfterMs)
+  }
+
   def Configured(maxExtraLoad: Double, sendInterrupts: Boolean): Param = {
```
jyanJing, line 137: hmmm, it looks like there are some duplicate code block, wdyt having one method call the other one? Maybe something like: 
`def Configured(maxExtraLoad: Double, sendInterrupts: Boolean): Param = Configured(maxExtraLoad, sendInterrupts, 1)`

finagle-http/src/main/scala/com/twitter/finagle/http/MethodBuilder.scala
```diff
@@ -241,6 +241,16 @@ class MethodBuilder private (mb: client.MethodBuilder[Request, Response])
       )
     )
 
+  def idempotent(maxExtraLoad: Double, minSendBackupAfterMs: Int): MethodBuilder =
```
jyanJing, line 244: dido, we can have 1 method call the other one here, too.

finagle-http/src/main/scala/com/twitter/finagle/http/MethodBuilder.scala
```diff
@@ -255,6 +265,16 @@ class MethodBuilder private (mb: client.MethodBuilder[Request, Response])
       )
     )
 
+  def idempotent(maxExtraLoad: Tunable[Double], minSendBackupAfterMs: Int): MethodBuilder =
```
jyanJing, line 268: dido

finagle-thriftmux/src/main/scala/com/twitter/finagle/thriftmux/MethodBuilder.scala
```diff
@@ -328,6 +328,12 @@ class MethodBuilder(
       mb.idempotent(maxExtraLoad, sendInterrupts = true, ResponseClassifier.RetryOnThrows)
     )
 
+  def idempotent(maxExtraLoad: Double, minSendBackupAfterMs: Int): MethodBuilder =
```
jyanJing, line 331: dido

finagle-thriftmux/src/main/scala/com/twitter/finagle/thriftmux/MethodBuilder.scala
```diff
@@ -339,6 +345,12 @@ class MethodBuilder(
       mb.idempotent(maxExtraLoad, sendInterrupts = true, ResponseClassifier.RetryOnThrows)
     )
 
+  def idempotent(maxExtraLoad: Tunable[Double], minSendBackupAfterMs: Int): MethodBuilder =
```
jyanJing, line 348: dido

finagle-core/src/main/scala/com/twitter/finagle/client/MethodBuilder.scala
emilhotkowski: The only problem I had with reusing these methods is that I would need to specify default value for **minSendBackupAfterMs** in idempotent method. For example :
```
def idempotent(
    maxExtraLoad: Double,
    sendInterrupts: Boolean,
    minSendBackupAfterMs: Int,
    classifier: service.ResponseClassifier
  ): MethodBuilder[Req, Rep] = {
    val brfParam =
      if (maxExtraLoad == 0) BackupRequestFilter.Disabled
      else BackupRequestFilter.Configured(maxExtraLoad, sendInterrupts, minSendBackupAfterMs)
    addBackupRequestFilterParamAndClassifier(brfParam, classifier)
  }

  def idempotent(
    maxExtraLoad: Double,
    sendInterrupts: Boolean,
    classifier: service.ResponseClassifier
  ): MethodBuilder[Req, Rep] =
    idempotent(maxExtraLoad, sendInterrupts, 1, classifier)
```
We would end up having "default" value for this parameter in two places. When calling idempotent method or when creating Configured object. We can always define something like and reuse it.
`private[finagle] val DefaultMinSendBackupAfterMs: Int = 1`
WDYT?

jyanJing:
It looks like the pr is already pulled and merging in progress internally, please ignore my comments, sorry about it!

PR #949, opened 2023-05-18
Add Scala 3 support to build and CI matrix

felixbr:
Hi, this is early work on supporting a native Scala 3 build for Finagle.

The changes I've made are close to the current setup in twitter/util (including [my PR](https://github.com/twitter/util/pull/308) there for Scala 3 CI).

Right now this only adds Scala 3 to `finagle-init` as a PoC because other modules are still blocked on twitter/util as far as I know.

I'm aware that many Finagle modules are already usable with Scala 3 projects using the cross-compilation (= `.cross(CrossVersion.for3Use2_13)`) feature. This works decently well for applications but is not a suitable solution for libraries which build on Finagle or more complex cross-building setups.

Note: Currently this branches from `release` not from `develop` because the latter seems to be unstable. I will rebase onto `develop` before converting the draft into a real PR ready to be merged.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -21,7 +21,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        scala: [2.12.12, 2.13.6]
+        scala: [2.12.12, 2.13.6, 3.2.2]
         java: ['1.8', '1.11']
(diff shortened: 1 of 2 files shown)


xin301x:
这是来自QQ邮箱的假期自动回复邮件。
 
您好，我最近正在休假中，无法亲自回复您的邮件。我将在假期结束后，尽快给您回复。

PR #954, opened 2023-08-07
`finagle-http2`: Add a per session request limit option

beatrichartz:
### Problem

When running finagle in a setup using HTTP/2 via application load balancers (ALB), the session is very likely subject to a limit on the number of requests. E.g. when running [via NGINX, the request limit is 1K requests per connection by default](http://nginx.org/en/docs/http/ngx_http_core_module.html#keepalive_requests). On AWS, ALB have a 10k request limit. This means that in setups that use finagle and ALBs for their L7 features, the session is subject to races happening on connection close regularly. Here's an illustration of the race happening after a limit of 10K requests has been reached:

Illustration of race condition leading to Connection Closed Errors

Netty fires inactive channel for any request that is in flight for the closed session, leading Finagle to propagate a `ChannelClosedException`. There is no possible remediation for this race in Netty since Finagle manages sessions/connections.

This condition can be easily reproduced by running a Finagle server & client with an ALB like NGINX in between, sending concurrent requests up until the limit. 

### Solution

Allow configuration of a `MaxRequestsPerSession` option that aligns a session with underlying request limits. The number of`MaxConcurrentStreams` is deducted to account for any requests in flight. When this number of requests is reached, mark the session as closed so it is shut down and a new session is created.

This option is off by default and therefore opt-in.

### Result

When the configured number of requests is reached, the session is terminated and a new session is opened:

Illustration of session closing and reopening behavior using the per session request limit

### DIFF
--- a/finagle-http2/src/main/scala/com/twitter/finagle/http2/param/Params.scala
+++ b/finagle-http2/src/main/scala/com/twitter/finagle/http2/param/Params.scala
@@ -97,6 +97,20 @@ object MaxFrameSize {
   implicit val param = Stack.Param(MaxFrameSize(None))
 }
 
+/*
+ * A class for configuring the maximum number of requests per connection
+ * MaxConcurrentStreams will be deducted from this number to allow requests
+ * in flight up to this number.
+ */
+case class MaxRequestsPerSession(maxRequestsPerSession: Option[Long]) {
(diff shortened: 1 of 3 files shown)


mattdickinson5:
hmm, for some reason this landed attributing the wrong person - apologies!

beatrichartz:
No worries, glad it landed! 👍

PR #957, opened 2023-09-20
finagle-core: Deposit budget once in MethodBuilder

### LINKED ISSUE #956: Methodbuilder retry budget gets double deposit
**Describe the bug**
The methodbuilder introduces a RetryFilter that gets a RetryBudget. This budget is passed to the stack and picked up by the RequeueFilter. 
On a successful response the Retry and RequeueFilter deposit both for the same request, effectively doubling the budget. 

**To Reproduce**
With such a budget 
```scala
val budget = Retries.Budget(RetryBudget(ttl = 10.seconds, 10, 0.1))
```
do 5 calls using the method builder interface and see that the `budget.retryBudget.balance == 101` 

expected is to reach 101 after 10 calls. (10% gets deposited). 

```scala
  test("retries and requeues do not double deposit RetryBudget") {
    val stats = new InMemoryStatsReceiver()
    val budget = Retries.Budget(RetryBudget(ttl = 10.seconds, 10, 0.1))

    val svc = Service.mk[Int, Int] { _ =>
        Future.value(1)
    }

    val stack = Retries
      .moduleRequeueable[Int, Int]
      .toStack(Stack.leaf(Stack.Role("test"), ServiceFactory.const(svc)))
    val ps =
      Stack.Params.empty +
        param.Label(clientName) +
        param.Stats(stats) +
        budget
    val stackClient = TestStackClient(stack, ps)

    val methodBuilder = MethodBuilder.from("retry_it", stackClient)
    val client = methodBuilder
      .newService("client")

    assert(budget.retryBudget.balance == 100)
    (1 to 10).foreach(_ => Await.result(client(1), 5.seconds))
    assert(budget.retryBudget.balance == 101) // currently it's 102. 
  }
```

**Expected behavior**
Deposit is only called once for a succesful request.

**Additional context**
The RequeueFilter has a WithdrawOnlyBudget for this exact case, but that only works for the Client (`Retries.moduleWithRetryPolicy`) , methodbuilder is not aware of this configuration.

DieBauer:
Problem

The methodbuilder interface applies the same retrybudget for RetryFilter and RequeueFilter in the same stack. Resulting in double deposit of retrybudget.

Solution

Re-use the WithdrawOnlyRetryBudget to prevent depositing successes in the RequeueFilter and only deposit in the RetryFilter. Same solution as the `Retries.moduleWithRetryPolicy` takes.

Result

Retrybudget only gets deposited once for a request.

Closes #956

### DIFF
--- a/finagle-core/src/main/scala/com/twitter/finagle/client/MethodBuilder.scala
+++ b/finagle-core/src/main/scala/com/twitter/finagle/client/MethodBuilder.scala
@@ -2,10 +2,7 @@ package com.twitter.finagle.client
 
 import com.twitter.finagle.Filter.TypeAgnostic
 import com.twitter.finagle.client.MethodBuilderTimeout.TunableDuration
-import com.twitter.finagle.service.Filterable
-import com.twitter.finagle.service.ResponseClass
-import com.twitter.finagle.service.ResponseClassifier
-import com.twitter.finagle.service.TimeoutFilter
+import com.twitter.finagle.service.{Filterable, ResponseClass, ResponseClassifier, Retries, RetryBudget, TimeoutFilter}
 import com.twitter.finagle.stats.LazyStatsReceiver
 import com.twitter.finagle.stats.StatsReceiver
 import com.twitter.finagle.tracing.TraceInitializerFilter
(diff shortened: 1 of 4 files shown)


mattdickinson5:
@DieBauer  - this code isn't compiling, and when i fix it the `MethodBuilderRetryTest` test does not pass.

DieBauer:
I've rebased to develop (SNAPSHOTS were shifted). And an assert was indeed wrong. In the last testcase .retry.disabled, you do not expect any 'retries' stat to be present! 

Let me know what you think!

PR #959, opened 2023-10-29
Portable shebang

andreoss:
Problem

- /bin/bash is not available on non-FHS distros, such as NixOS

Solution

- replace /bin/bash with /usr/bin/env bash 	

Result

See also

https://github.com/twitter/dodo/pull/15
https://github.com/twitter/finatra/pull/589
https://github.com/twitter/scrooge/pull/366
https://github.com/twitter/util/pull/314
https://github.com/twitter/twitter-server/pull/77

### DIFF
--- a/doc/src/sphinx/code/client-server-anatomy/sbt
+++ b/doc/src/sphinx/code/client-server-anatomy/sbt
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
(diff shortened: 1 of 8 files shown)


xin301x:
这是来自QQ邮箱的假期自动回复邮件。
 
您好，我最近正在休假中，无法亲自回复您的邮件。我将在假期结束后，尽快给您回复。

